PR #6, opened 2016-02-29
Fix SWARM-310

heiko-braun:
DO NOT MERGE. This is for review purposes.
- Prevents scanning the whole filesystem. 
- Blow up if module references are not jar:file: or file:

### DIFF
--- a/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/NestedJarResourceLoader.java
+++ b/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/NestedJarResourceLoader.java
@@ -41,10 +41,11 @@ public class NestedJarResourceLoader {
 
         //System.err.println( "** " + base + ", " + rootPath + ", " + loaderPath + ", " + loaderName );
 
-        if (base.toExternalForm().startsWith("jar:file:")) {
-            int endLoc = base.toExternalForm().indexOf(".jar!");
+        String urlString = base.toExternalForm();
+        if (urlString.startsWith("jar:file:")) {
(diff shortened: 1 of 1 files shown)


bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/NestedJarResourceLoader.java
```diff
@@ -76,8 +77,14 @@ public static ResourceLoader loaderFor(URL base, String rootPath, String loaderP
...
             }
         }
+        else if (urlString.startsWith("file:")) {
+            return ResourceLoaders.createFileResourceLoader(
+                    loaderPath,
+                    new File(urlString.substring(5, urlString.length()))
+            );
+        }
 
-        return ResourceLoaders.createFileResourceLoader(loaderPath, new File(rootPath));
```
heiko-braun, line 80: The `new File(rootPath)` was causing the full FS scan. 'rootPath' is a parameter we can ignore, since we don't deal with module directories before the fact.

bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/NestedJarResourceLoader.java
```diff
@@ -76,8 +77,14 @@ public static ResourceLoader loaderFor(URL base, String rootPath, String loaderP
...
+        throw new IllegalArgumentException("Illegal module loader base: "+ base);
```
heiko-braun, line 87: I don't see why should support anything else then `jar:file:` and `file:`

PR #8, opened 2016-03-03
[SWARM-320] CNFE for org.jboss.arquillian.test.spi.TestResult in org.…

tdiesler:
…wildfly.swarm.arquillian.daemon:api


kenfinnigan:
@tdiesler can you modify the commit to only changed code?

Looks like there's changes that break the code style we're using

PR #9, opened 2016-03-04
[SWARM-323] NPE in ServiceActivatorArchiveImpl with no archive suffix

tdiesler:


kenfinnigan:
It's hard to know what changes are relevant to the fix as there seem to be some changes that revert recent code style updates and changes such as `this.asset` to `asset`.

PR #10, opened 2016-03-07
Swarm 324: Bootstrap performance issues

heiko-braun:
This is for review purposes only.

Moving to persistent jandex indexes

### DIFF
--- a/container/runtime/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java
+++ b/container/runtime/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java
@@ -18,6 +18,7 @@ package org.wildfly.swarm.container.runtime;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.lang.reflect.Modifier;
 import java.net.URL;
@@ -48,8 +49,10 @@ import org.jboss.as.server.Services;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ValueExpression;
 import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.CompositeIndex;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.Index;
+import org.jboss.jandex.IndexReader;
 import org.jboss.jandex.Indexer;
 import org.jboss.modules.Module;
 import org.jboss.modules.ModuleIdentifier;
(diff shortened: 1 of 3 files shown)


heiko-braun:
@kenfinnigan Here's a working example for wfs-core. Maybe we can use it as a baseline for discussing the changes needed to enable persistent indexes?

heiko-braun:
@kenfinnigan Please not that it requires these changes to be merged into the jandex plugin. Hence the 1.0.4-SNAPSHOT dependency which causes the CI build to fail.

https://github.com/wildfly/jandex-maven-plugin/pull/9

kenfinnigan:
FYI, I had added the jandex plugin to parent pom so it can be shared across more than core.

Is there any benefit to customizing the index name? We're not doing anything WildFly Swarm specific in building the jandex right?

heiko-braun:
The custom index name is necessary, because otherwise it conflicts with indexes that WF expects at that location.

kenfinnigan:
Is it building differently than what WF expects? Otherwise I would not expect it to be an issue

heiko-braun:
yes, the indexes seem to be different. i didn't have the time to fully track down all the inner workings of deployer chains in WF and how they leverage jandex, but I can tell that our integration tests fail with indexes provided by us that carry the default name

heiko-braun:
The latest update pushes the jandex plugin into the fractions. If this passes the tests, then we would be ready to apply the changes to the other fraction outside of core

heiko-braun:
The latest commit removes the jandex file from main/module.xml

heiko-braun:
Cleaned up and squashed the commits to make it more comprehensible

heiko-braun:
With this build setup (jandex in the core pom) all our own fractions that don't reside within core inherit everything they need to make it work. 

Only fractions that don't inherit from core would need to add the jandex plugin. But that's just the way it is. I don't think we do much about that.

heiko-braun:
With regard to the choices of 

a) jandex in the fraction plugin and
b) jandex in it's own plugin (like it is now)

the first solution (a) sounds nice, but it would require adding the fraction plugin to any runtime module. currently it's only attached to the main module, but indexes generated at that level don't help. 

the second option (b) seems to be most versatile, doesn't require additional changes to any downstream module and creates an index with every jar. 

**Long Story Short** 

It's complicated, don't ask any further, this is the best approach I could find to balance ease of use with imposed constraints.

heiko-braun:
From my point of view this is ready to be merged

bobmcwhirter:
I'd say merge, and then I'll rebase my :spi stuff and get that in today, hopefully.

PR #12, opened 2016-03-07
DO NOT MERGE -- SPI testing

bobmcwhirter:
Remove SPI-like things from this repository to the wildfly-swarm-spi.git repository (and org.wildfly.swarm:spi, :spi-api, and :spi-runtime GAVs).

Refactor this codebase for the different GAV for these previously-mentioned
SPI-like things, along with the packages that changed.

Primarily interfaces:
- Fraction
- ServerConfiguration
- DefaultDeploymentFactory

Additionally, ArtifactManager is now exposed (and slightly hackily)
as ArtifactLookup in the SPI.

Various common bits of ShrinkWrap (DependenciesContainer) have also
moved upstream to the SPI.

Additionally, normalize some of the maven projects within this repository,
so that:
- foo/pom.xml
  foo-parent aggregator
- foo/foo/pom.xml
  user-facing org.wildfly.swarm:foo dependency, which more cleanly gathers
  together foo-api, foo-runtime and foo-modules with appropriate Maven scopes
  to help with intra-repository build/test ordering, and also provides
  simple a better handle to the entirety of a fraction.
  
  It is this POM where inter-fraction runtime dependencies are specified.
  For instance, jaxrs requires undertow, right? So
  
  jaxrs/jaxrs/pom.xml -> org.wildfly.swarm:undertow
- foo/api/pom.xml
  Previously was :foo, and now is :foo-api
  Has removed reference to :container, since it should only be using stuff
  from :spi-api.
  
  In the cases where one API builds upon another, :foo-api may have a dependency
  upon :bar-api (not :bar).
  
  If jaxrs-api needed some classes from the Undertow API, it would be
  
  jaxrs/api/pom.xml -> org.wildfly.swarm:undertow-api
- foo/runtime/pom.xml
  Still foo-runtime, unchanged, aside from depending on :foo-api now
  instead of simply :foo
- foo/modules/**
  unchanged, other than modules/.../api/module.xml now references
  the foo-api.jar, as you might hope and expect from a module with
  a slot="api"

Ultimately, this change brings about a 33% increase (or so) of
the number of pom.xml's in the project.  You're welcome, Toby.

It also radically reduces the size of each pom.xml. I believe we
had been accumulating dependencies through copy/pasting without
actually realizing if they were truly needed or not.  Those have
been removed where noticed.


undertow/api/src/test/java/org/wildfly/swarm/undertow/JbossWebContainerTest.java
```diff
@@ -42,6 +45,7 @@ public void testSettingContextRoot() throws Exception {
     }
 
     @Test
+    @Ignore
```
kenfinnigan: Any reason the test is ignored?

container/api/src/test/java/org/wildfly/swarm/container/JARArchiveTest.java
```diff
@@ -1,108 +0,0 @@
-/**
```
kenfinnigan, line 1: This and subsequent tests for previous container api classes have been removed but not added to :spi repository. Oversight?

kenfinnigan:
We don't need this anymore right?

bobmcwhirter:
Kill it!

## 

Sent from my iPhone

On Mar 11, 2016, at 10:41 PM, Ken Finnigan notifications@github.com wrote:

We don't need this anymore right?

kenfinnigan:
This has been dealt with

PR #15, opened 2016-03-18
Support for @Health annotations

heiko-braun:
Includes support for security realms

NOTE: Requires https://github.com/wildfly-swarm/wildfly-swarm-spi/pull/1 to be merged first


container/runtime/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
```diff
@@ -69,13 +71,31 @@ public void debug(boolean debug) {
...
+            throw new RuntimeException(e);
+        }
+
+        Index index = indexer.complete();
+
+
+        // 2. give the factions a chance to handle the deployment
         for (ServerConfiguration each : this.configurations) {
             each.prepareArchive(deployment);
+            each.processMetaData(deployment, index);
```
kenfinnigan: We may want to do the metadata processing in a loop after this one, to account for one Fraction doing something in prepareArchive() that impacts another Fractions processMetaData().

Thinking some more, we may even want to create the meta data index after looping through prepareArchive()

kenfinnigan:
Does the spi module.xml in /container need to be updated to have a dependency on jandex module? Now that the spi has a provided dependency for it?

heiko-braun:
The last commit adds the security realm support

kenfinnigan:
With the latest changes, is the MapIdentityManager needed still?

heiko-braun:
Yes, the Map IDM is not needed anymore

heiko-braun:
@kenfinnigan please keep this on hold. I still need to tweak the authentication bits a little bit

container/runtime/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
```diff
@@ -69,13 +71,34 @@ public void debug(boolean debug) {
     @Override
     public void deploy(Archive<?> deployment) throws DeploymentException {
 
+        // 1. give the factions a chance to handle the deployment
```
kenfinnigan: Apologies Heiko, you're going to be annoyed by this! After saying what I did on Friday, I think, I realized that maybe some calls to prepareArchive() need to rely on things in the index? I'm not entirely sure, but it was something that popped into my head.

For instance, would we want prepareArchive() to be able to search for annotations, etc to then add/remove code from the Archive depending on what they find?

If so, then we do need to index first, and now you hate me ;-)

So, if we took that approach, do we need to re-index the index after prepareArchive()?

kenfinnigan:
No problem, wasn't planning to merge just yet

container/runtime/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
heiko-braun: I'd suggest we'll leave it as is until we hit the cases you mention.

heiko-braun:
Cleaned up and squashed everything into one commit. From my point of view this is good to go. I'd suggest to move any further into dedicated branches (i.e. @kenfinnigan thoughts on the RuntimeDeployer)

heiko-braun:
My vote on merging this one: +1 

@kenfinnigan ?

container/runtime/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
gastaldi: s/factions/fractions

heiko-braun:
The latest commit fixes a typo and squashes everything into one again

gastaldi:
Looks good :+1:

kenfinnigan:
Just a quickie on RuntimeDeployer. Would we envisage that something like https://github.com/wildfly-swarm/wildfly-swarm-swagger/blob/master/adapter/runtime/src/main/java/org/wildfly/swarm/swagger/runtime/SwaggerConfiguration.java#L63 is replaced with using the index or not?

heiko-braun:
Looks like I broke the security with one of my latest commits. I'll need to enhance the test coverage for this and fix it, before we can move on.

heiko-braun:
@kenfinnigan re: Swagger. Yes, that's one of the intended use cases.

kenfinnigan:
Given that, we will need to prepare the index before calling prepareArchive() but also pass the index into that method, which is an SPI change.

Is there an alternative I haven't thought of?

heiko-braun:
@kenfinnigan we can still shuffle things around later, but I would suggest not to make all possible use cases (i.e. swagger) a pre-requisite for merging what we have so far.

heiko-braun:
Squashed everything again and fixed the JAXRSArquillianTest

PR #18, opened 2016-03-24
Introduce Health check API

heiko-braun:
```
    @GET
    @Health
    @Path("/health")
    public HealthStatus checkHealth() {
        return HealthStatus.up().withAttribute("status", "UP");
    }

```


monitor/modules/src/main/resources/modules/org/wildfly/swarm/monitor/main/module.xml
```diff
@@ -2,7 +2,7 @@
 
   <dependencies>
     <!-- For when run with bonafide IDE classpath -->
-    <system export="true">
+    <system export="false">    <!-- this needs to be suppressed, otherwise we classloading problems in the jaxrs tests-->
```
kenfinnigan: Wouldn't this prevent 'monitor' from working when run from an IDE?

jaxrs/runtime/src/main/java/org/wildfly/swarm/jaxrs/runtime/JAXRSConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public JAXRSConfiguration() {
+        super(JAXRSFraction.class, EXTENSION_MODULE);
+    }
+
+    @Override
+    public Optional<Map<QName, XMLElementReader<List<ModelNode>>>> getSubsystemParsers() throws Exception {
+        return AbstractParserFactory.mapParserNamespaces(new JAXRSParserFactory());
+    }
+
```
kenfinnigan: What's the benefit of changing 'jaxrs' from using @Configuration in api to having 'JAXRSConfiguration' in runtime?

jaxrs/modules/src/main/resources/modules/org/wildfly/swarm/jaxrs/runtime/module.xml
```diff
@@ -9,6 +9,7 @@
     <module name="org.wildfly.swarm.container" slot="runtime"/>
     <module name="org.wildfly.swarm.configuration"/>
 
+    <module name="org.jboss.shrinkwrap"/>
```
kenfinnigan: Was there a problem that this was needed to fix?

heiko-braun: That's a leftover. It can be removed

jaxrs/runtime/src/main/java/org/wildfly/swarm/jaxrs/runtime/JAXRSConfiguration.java
heiko-braun: I had an alternative impl. that did require this. Until i know whether the alternative needs to be revisited, I'd like to keep the config. If, at the end, we decide against it, i'll remove it.

monitor/modules/src/main/resources/modules/org/wildfly/swarm/monitor/main/module.xml
heiko-braun: yes, but it was the only way I could get it working. i'll catch you on irc to discuss the dependencies and how they can be aligned ti improve the situation.

heiko-braun:
The update now contains the following changes:
- /health simply returns a list of subresources, /health/app/foo & /health/app/bar
- the 'inheritSecurity' attribute determines whether a 403 for direct access of any subresource will be returned as means to protect the security
- HealthStatus.withAttribute("key", "val") allows to append arbitrary values to a health check response

monitor/runtime/src/main/java/org/wildfly/swarm/monitor/runtime/SecureHttpContexts.java
```diff
@@ -119,10 +121,37 @@ private HttpHandler secureHandler(final HttpHandler toWrap, SecurityRealm securi
         handler = new PredicateHandler(exchange -> {
             return Queries.isSecuredHealthEndpoint(monitor, exchange.getRelativePath())
                     || HttpContexts.getDefaultContextNames().contains(exchange.getRelativePath());
-        }, handler, next);
+        }, handler, toWrap);
 
+
```
heiko-braun, line 124: can be removed

PR #20, opened 2016-03-30
[SWARM-390] resolve artifacts in parallel

hwellmann:


tools/src/main/java/org/wildfly/swarm/tools/DependencyManager.java
```diff
@@ -247,9 +249,7 @@ protected void analyzeDependencies(boolean resolveTransitive) throws Exception {
             this.dependencies.clear();
             this.dependencies.addAll(newDeps);
         } else {
-            for (ArtifactSpec each : this.dependencies) {
-                resolveArtifact(each);
-            }
+            this.resolver.resolveAll(this.dependencies);
```
gastaldi: Should this line call the newly created `resolveAllArtifacts` method?

hwellmann: Yes, makes sense. Also in L248. I'll update the PR.

PR #22, opened 2016-04-04
[SWARM-121] Enable messaging clustering

jmesnil:
JIRA: https://issues.jboss.org/browse/SWARM-121

### DIFF
--- a/messaging/messaging/pom.xml
+++ b/messaging/messaging/pom.xml
@@ -60,6 +60,14 @@
       <groupId>org.wildfly.swarm</groupId>
       <artifactId>security</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.wildfly.swarm</groupId>
+      <artifactId>jgroups</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.wildfly.swarm</groupId>
+      <artifactId>undertow</artifactId>
+    </dependency>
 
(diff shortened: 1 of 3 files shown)


jmesnil:
https://github.com/wildfly-swarm/wildfly-swarm-examples/pull/60 is an example illustrating how to create a messaging cluster of 2 Swarm servers.

messaging/messaging/pom.xml
```diff
@@ -60,6 +60,14 @@
       <groupId>org.wildfly.swarm</groupId>
       <artifactId>security</artifactId>
     </dependency>
+    <dependency>
```
jmesnil, line 63: This JGruoups dependency is only required if users want to use clustering.
Maybe it should not be added to the messaging POM and the user should add it explicitly to its application POM instead...

messaging/messaging/pom.xml
```diff
@@ -60,6 +60,14 @@
...
+      <groupId>org.wildfly.swarm</groupId>
+      <artifactId>jgroups</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.wildfly.swarm</groupId>
```
jmesnil, line 68: Same remark than above (the Messaging subsystem depends on Undertow to multiplex its messaging protocol on top of HTTP)

PR #23, opened 2016-04-08
Add utility methods for enabling AJP

emag:
## Motivation

For enabling AJP, some boilerplate codes(adding ajp socket-binding, listener) is needed.
## Modifications

Added utility methods enabling AJP.
## Result

More easily enabling AJP.

### DIFF
--- a/undertow/api/src/main/java/org/wildfly/swarm/undertow/UndertowFraction.java
+++ b/undertow/api/src/main/java/org/wildfly/swarm/undertow/UndertowFraction.java
@@ -89,6 +89,17 @@ public class UndertowFraction extends Undertow<UndertowFraction> implements Frac
     }
 
     /**
+     * Create the default HTTP and AJP fraction.
+     *
+     * @return The configured fraction.
+     * @see #enableAJP()
+     */
+    public static UndertowFraction createDefaultAndEnableAJPFraction() {
+        return createDefaultFraction()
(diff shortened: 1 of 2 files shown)


gastaldi:
LGTM

undertow/api/src/main/java/org/wildfly/swarm/undertow/UndertowFraction.java
```diff
@@ -134,6 +172,9 @@ public void initialize(Fraction.InitContext initContext) {
         initContext.socketBinding(
                 new SocketBinding("https")
                         .port(SwarmProperties.propertyVar(UndertowProperties.HTTPS_PORT, "8443")));
+        initContext.socketBinding(
+                new SocketBinding("ajp")
+                        .port(SwarmProperties.propertyVar(UndertowProperties.AJP_PORT, "8009")));
```
kenfinnigan: Should this call to create a SocketBinding be wrapped in a check for this.enableAJP?

Otherwise we're creating a SocketBinding for no reason, right?

emag: @kenfinnigan Thank you for checking. I agree with your comment of the former. For users who don't use AJP, it may be a little bit overhead. I'll re-sent the p-r.

PR #25, opened 2016-04-13
Add methods manipulating security-domain in jboss-web.xml

emag:


kenfinnigan:
It's also worth exposing the new methods on https://github.com/wildfly-swarm/wildfly-swarm-core/blob/master/undertow/api/src/main/java/org/wildfly/swarm/undertow/descriptors/JBossWebContainer.java as then you can do `archive.setSecurityDomain()`

PR #26, opened 2016-04-15
SWARM-311 introduce MySQL JPA fraction

tremes:


tremes:
Hmm couldn't transfer `operalaunchers-1.1.jar`. It doesn't seem to be related to my changes.

PR #32, opened 2016-04-28
Swarm 391

rbattenfeld:
Hi Wildfly-Swarm, this a pull request for supporting resource adapter deployments in wildfly-swarm. I was able to write a test for the wildfly-swarm-examples repo. I will push this change later this week.

This version should support external ironjacamar.xml files. The declarative approach is not finalized yet. First, I have to understand, why the resource-adapter class in config api is not covering for example xa stuff.

Please let me know what I should make better for this push

Thanks,
Ralf


rbattenfeld:
I have corrected the wrong pom versions.

heiko-braun:
Thanks @rbattenfeld . I'll have a look this week.

heiko-braun:
@kenfinnigan @bobmcwhirter feel free to grab this one. I won't be able to look into before next week

resource-adapters/api/src/main/java/org/wildfly/swarm/resourceadapters/IronJacamarXmlAssetImpl.java
```diff
@@ -0,0 +1,204 @@
...
+import org.wildfly.swarm.config.resource.adapters.ResourceAdapter;
+import org.wildfly.swarm.config.resource.adapters.resource_adapter.AdminObjects;
+import org.wildfly.swarm.config.resource.adapters.resource_adapter.ConfigProperties;
+import org.wildfly.swarm.config.resource.adapters.resource_adapter.ConnectionDefinitions;
+import org.wildfly.swarm.container.util.XmlWriter;
+
+/**
+ * @author Ralf Battenfeld
+ */
+enum IronJacamarXmlAssetImpl {
```
bobmcwhirter, line 30: Instead of my crappy XmlWriter, if there's a shrinkwrap-descriptor for this file, that might be best.  If there isn't, we can maybe ask @tobias to help us get one.

rbattenfeld, line 30: Yes, there is one. I was actually doing this for the descriptor project. But there is no one for the resource adapter since this is hosted in the ironjacamar project. I see your point. I'd like to enhance this when there is a ironjacamar.xml support in the config api as Heiko mentioned.

kenfinnigan:
Merged

kenfinnigan:
Thanks for this @ralfbattenfeld!

Do you happen to have an example showcasing this PR that we could integrate into our examples repo?

If you do, a PR would be great

Thanks again!

PR #34, opened 2016-05-10
SWARM-447 StageConfig CDI integration rework

mkouba:
- make ConfigValue a qualifier
- replace ConfigExtension with producers
- remove Configured
- add ConfigValueProducer test

Unlike the original solution:
- let the CDI container perform the injection (`@ConfigValue` qualifier must be used together with `@Inject`)
- allow to use constructor injection and initializers
- allow to inject `Resolver<String>` if custom conversion is needed
- see also `ConfigValueProducerTest` for examples


heiko-braun:
Cool, looks good to me. Have you tested it with the config examples?

cdi/test/src/test/java/org/wildfly/swarm/cdi/ConfigAwareBean.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Inject
+    ConfigAwareBean(@ConfigValue("logger.level") String logLevel,
+            @ConfigValue("swarm.port.offset") Resolver<String> resolver) {
+        this.logLevel = logLevel;
+        this.portOffsetResolver = resolver;
+    }
+
+    @Inject
+    void init(@ConfigValue("logger.level") String logLevel) {
```
heiko-braun, line 46: I now see what you mean. This is definitely better then what we had before.

mkouba:
> Have you tested it with the config examples? 

No, I have no idea how to do it ;-)

heiko-braun:
All you need to do is to build core with your changes and then run `mvn clean verify` on the examples: https://github.com/wildfly-swarm/wildfly-swarm-examples

heiko-braun:
You can also shortcut the examples by doing:

1) `cd base` `mvn clean install`
2) `cd ../config-options` `mvn clean verify`

heiko-braun:
These are the tests that should pass: https://github.com/wildfly-swarm/wildfly-swarm-examples/tree/master/config-options/cdi-injection

heiko-braun:
The reason for running the examples is that they verify specific class loading combinations

mkouba:
> These are the tests that should pass...

These will have to be fixed - `@Configured` is removed and injection points require `@Inject` now.

PR #43, opened 2016-06-09
[SWARM-507] Autodiscover external configuration files

tdiesler:


container/runtime/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java
```diff
@@ -23,7 +23,14 @@
 import java.lang.reflect.Method;
```
kenfinnigan, line 23: It doesn't appear that any of the changes in RuntimeServer are relevant to this fix, is that correct?

tdiesler, line 23: Yes, I may have removed some slack, while going along

container/api/src/main/java/org/wildfly/swarm/container/Container.java
```diff
@@ -445,21 +487,59 @@ public Archive createDefaultDeployment() {
...
+
+    /**
+     * @return The URL to the default deployment or null
+     */
+    private URL getDefaultDeploymentURL() throws IOException {
+        if (defaultDeploymentURL == null) {
+            Archive<?> archive = getDefaultDeployment();
+            if (archive != null) {
+                File tmpdir = TempFileManager.INSTANCE.newTempDirectory("deployment", ".d");
+                new ExplodedExporterImpl(archive).exportExploded(tmpdir);
```
gastaldi, line 522: Use archive.as(ExplodedExporter.class).exportExploded instead (Stick to the API rather than the impl)

tdiesler, line 522: This approach would fail in module [arquillian/test](https://github.com/wildfly-swarm/wildfly-swarm-core/tree/master/arquillian/test)

kenfinnigan:
Any chance of a test in /container to verify these changes?

Would be nice to ensure that we don't break the behavior going forward

tdiesler:
It is verified [here](https://github.com/wildfly-swarm/wildfly-swarm-examples/tree/master/camel/camel-jpa/src/main/resources/META-INF) - is that not good enough?

kenfinnigan:
That's fine

PR #46, opened 2016-06-17
Add gradle support to ArtifactManager

mhofstetter:
Creating a org.wildfly.swarm.undertow.WARArchive with ShrinkWrap and trying to add all dependencies with "addAllDependencies" results in multiple problems if you are using gradle as your buildtool, having multiple projects (gradle multip-project-build) and starting the Main-Class from the IDE.
- The exclusion for all org.wildfly.swarm artifacts doesn't work: The filter for the local maven repository and the local gradle-cache are using a different approach for storing the artifacts in the filesystem. Maven creates a hierarchical folder-structure for the group of the maven artifact whereas gradle creates one folder for the group.
- Creating an on-the-fly-JavaArchive for directories in the classpath creates multiple JavaArchives for one "javaproject": Gradle separates the outputfolder for resources and classes. But they have to be packed into the same JavaArchive - otherwise CDI doesn't work, when the beans.xml is placed in a different JAR then the classes.
- The filter for not including classpath-elements which are placed under "user.dir" included to many elements when there when one projectname is the "prefix" of another projectname.

### DIFF
--- a/container/api/src/main/java/org/wildfly/swarm/internal/ArtifactManager.java
+++ b/container/api/src/main/java/org/wildfly/swarm/internal/ArtifactManager.java
@@ -18,11 +18,14 @@ package org.wildfly.swarm.internal;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


kenfinnigan:
I've merged this in.

Thanks a bunch for submitting this.

PR #49, opened 2016-06-23
SWARM-528 first part of the issue - allow to pass port offset to test.

tremes:
Test would be nice but when I overwrite default value in arquillian.xml then it will affect other tests as well (was thinking about next test right to the `ArquillianTest`). So maybe we can introduce some additional profile or something like that.


heiko-braun:
That's great. Thanks @tremes

PR #51, opened 2016-06-29
Simplified management-console

gastaldi:
This PR depends on https://github.com/wildfly/wildfly-core/pull/1645 to be available
as the existing code throws a NPE in WildFly Core (See WFCORE-1629). 

At last but not least, -Dswarm.http.eager=true should be true to enable it

### DIFF
--- a/management-console/src/main/java/org/wildfly/swarm/management/console/runtime/ManagementConsoleConfiguration.java
+++ b/management-console/src/main/java/org/wildfly/swarm/management/console/runtime/ManagementConsoleConfiguration.java
@@ -15,15 +15,10 @@
  */
 package org.wildfly.swarm.management.console.runtime;
 
-import java.net.URL;
 import java.util.Collections;
-import java.util.Enumeration;
 import java.util.List;
 
-import org.jboss.modules.Module;
-import org.jboss.modules.ModuleLoader;
 import org.jboss.shrinkwrap.api.Archive;
-import org.jboss.shrinkwrap.api.ShrinkWrap;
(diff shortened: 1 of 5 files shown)


gastaldi:
Also this change introduces a dependency on the Management fraction.

kenfinnigan:
is this already merged?

gastaldi:
This should only be merged after we update WildFly, otherwise it throws a NPE inside WF Core classes

management-console/src/main/java/org/wildfly/swarm/management/console/runtime/ManagementConsoleConfiguration.java
```diff
@@ -40,11 +35,15 @@ public ManagementConsoleConfiguration() {
 
     @Override
     public List<Archive> getImplicitDeployments(ManagementConsoleFraction fraction) throws Exception {
-        WARArchive war = ArtifactLookup.get()
-                .artifact("org.jboss.as:jboss-as-console:jar:2.8.25.Final:resources", "management-console-ui.war")
```
heiko-braun, line 44: Ideally this would not be hardcode. Take a look at the swagger web app fraction. It Does retrieve the actual version the maven dependencies through some level of indirection (generated properties file)

gastaldi, line 44: @heiko-braun, how would that be different from putting in the java class? I could also make it filterable when copying resources in maven.

bobmcwhirter:
Simpler via CDI changes and removed hard coded version.

PR #52, opened 2016-07-07
The fixed the pattern for the default color formatter

ericparton:
The pattern for the default color formatter in the logging fraction didn't include colors, so I updated the pattern


jamezp:
This looks good to me.

gastaldi:
Thank you

PR #57, opened 2016-07-18
SWARM-565: Encapsulate dependency resolution

heiko-braun:
This relates to https://issues.jboss.org/browse/SWARM-565

I've been trying to improve the resolution of dependencies that happens when `deployment.addAllDependencies();` is called from **within an IDE**, i.e. in `org.wildfly.swarm.examples.jaxrs.shrinkwrap.Main()`

This first commit is merely for reviewing purposes, to discuss the approach in general. Some notable changes are:
- the algo works inclusive (define the inclusion and discard the rest), in contrast the previous impl. that worked exclusive (define the exclusion and keep the remainder)
- the system property `user.dir` still works unreliably (i.e. in multi-module IDE setups)
- it's not clear how/if to process `target/classes`


heiko-braun:
@kenfinnigan I would appreciate some thoughts on this. In particular the changes in `SystemDependencyResolution#resolve()` which was previously in `ArtifactManager`

container/src/main/java/org/wildfly/swarm/internal/SystemDependencyResolution.java
```diff
@@ -0,0 +1,144 @@
...
+            WildFlySwarmClasspathConf classpathConf = new WildFlySwarmClasspathConf();
+            Set<String> classpathElements = new HashSet<>();
+            Set<String> providedGAVs = new HashSet<>();
+            List<String> testClasspathElements = testClasspath!=null ?
+                    testClasspath : Collections.EMPTY_LIST;
+
+            // filter classpath entries, makes some assumptions about maven
+            // TODO: not sure how this actually works with gradle. For now I assume gradle works on the same same local repository ...
+
+            final String m2Repo = "m2" + File.separator + "repository";
```
kenfinnigan: I don't think we can guarantee that location, as users _could_ have their repository in any directory

container/src/main/java/org/wildfly/swarm/internal/SystemDependencyResolution.java
```diff
@@ -0,0 +1,144 @@
...
+            final String projectDir = pwd + File.separatorChar;
+
+            for (final String element : classpath) {
+
+                // JAR files within CP scope
+                if(element.contains(m2Repo) && element.endsWith(JAR)) {
+                    // Read wildfly-swarm-classpath.conf entries
+                    if(!dryRun) {
```
kenfinnigan: Is dryRun meant to be set when executing in IDE only?

If so, how is it set for a user running a main()?

heiko-braun: dryRun is only used for test runs, but I realised it's probably just a quick fix.  A proper solution would require  to further refactor the code to split the classpath filtering (string values) from the actual archive parsing/creation (shrinkwrap archives).

container/src/main/java/org/wildfly/swarm/internal/SystemDependencyResolution.java
heiko-braun: right, so the default is `${user.home}/.m2/repository`, but i t can be overriden in either 

a) `settings.xml`
b) or using `-Dmaven.repo.local`

But considering that the algo so far is at best an approximation, I think a solution that works for the majority of users would be good enough. So one way to approach it would be to assume the majority uses the default location and still respect `-Dmaven.repo.local`, if given. I think the only case this would not cover is changes to `settings.xml`.

kenfinnigan: But the previous implementation worked wherever the m2repo was located, so wouldn't such an approach be a regression?

heiko-braun: Well, the main reason for looking into this was that it didn't work (See https://issues.jboss.org/browse/SWARM-565). It seems there a a few things that overlap each other and cause problems, primarily:
- the question of what deps actually get included in the IDE case
- the question what an IDE provides as `user.dir`
- the question if `target/classes` is included or not (i think this depends in the way the IDE builds maven projects)

kenfinnigan: I'm not sure what's going on as we made it work with multi module maven projects months ago.

heiko-braun:
Thinking further about the reasoning for my changes. I am coming from this end:
- `user.dir` seems to be unreliable (multi-module projects)
- `java.home` filtering seems to unreliable (on my box assorted JDK libs get included)
- the codepath that drives my changes was `deployment.addAllDeps()`

So, considering the above, in particular the last point, I was thinking that baselining on whatever information we have, or assumptions we can make about, the maven environment would probably lead to a more reliable resolution of dependencies. Overall, all we want with `deployment.addAllDeps()` is to refer to maven dependencies. And the approach in this PR thus tries to get away from `user.dir` as a criteria and `java.home` as a criteria. 

I hope this sheds some light on my line of thinking.

kenfinnigan:
The problem is that IDE and wildfly-swarm:run execution don't deal with Maven dependencies. They only deal with directories and files that are already resolved to a location

heiko-braun:
Right, at least not directly. Indirectly however, the classpath (as in `System.getProperty("java.class.path")` ) is determined from the maven deps within the environment that invokes the main class, i.e. the maven plugin or the IDE itself.

heiko-braun:
Actually I think that even in a gradle build, the classpath ultimately goes back to a file system structure that references a local maven repository, because we only ship maven artifacts. That's why all gradle builds tie into the maven central and the local maven repo: 

https://github.com/wildfly-swarm/wildfly-swarm-examples/blob/master/gradle/build.gradle#L4

As such I would expect, that even IDE execution that are results from imported gradle projects to reflect a similar classpath structure around a local maven repo.

container/src/main/java/org/wildfly/swarm/internal/SystemDependencyResolution.java
```diff
@@ -0,0 +1,144 @@
...
+                    }
+
+                    if (!excluded(exclusions, element)) {
+                        //System.out.println("Adding " + element);
+                        classpathElements.add(element);
+                    }
+                }
+
+                // classes Dir's within scope (i.e. target/classes in case of maven build)
+                else if(element.startsWith(projectDir) && element.endsWith(CLASSES)) {
```
mhofstetter: With such an approach you will never catch all cases of directories with relevant classes in it - without interpreting the definition of the build tool. In gradle the build output folders have this pattern "build/[classes|resources]/<sourceset-name>" - for example "build/classes/main. 

At best you can try to evaluate the name of the corresponding in-memory-JARArchive for better "debugging" - that's the current logic in ArtifactManager which you moved to a default-method in DependencyResolution.

mhofstetter:
Gradle uses the local m2-repo only for dependency-resolution of local builded artifacts. But in case of a artifact from a remote-repository gradle stores this artifact in it's own cache under ~/.gradle/caches/... .

In general I think the previous approach of grabbing all dependencies of the java.class.path and then exclude some jars was the right one. It should be easier to exclude some jars like JRE-libs (yes it sometimes doesn't work...) then interpreting the logic and conventions of multiple buildtools. The only thing I still don't get is why there is this filter for the "user.dir" directory.

heiko-braun:
Thanks @mhofstetter for the feedback. From this point of view I agree that the baseline on `.m2/repository` doesn't make much sense. I'll revert that to the previous of exclusions around the `java.class.path`. 

Regarding the question of classes directories, regardless of the build tool (could be `target/classes` in the maven case or `build/classes` in the gradle case):

Do we actually need to consider those or can we assume the build tool runs a full install cycle, resulting in jars in the local repo, when executed from within the IDE?

heiko-braun:
Regarding the `user.dir` property: This one seems to be really flaky, at least in intellij. Depending on the workspace setup it points to different project/module roots. This doesn't seem to be an issue when excluding `<target|build>/classes` as it is now. 

In my setup, for instance, I have core and examples in the same workspace. But my right click run on an example `Main()` uses the core module root as the `user.dir`. In other cases it was using the examples root as the `user.dir`.  It doesn't take much to see how an inclusion of classes with `deployment.addAllDeps()` can easily mess things up.

mhofstetter:
you're welcome @heiko-braun

In my opinion the addAllDeps should consider those directories because otherwise the classpath isn't complete. At least when talking about multiproject builds 'within the same build' (one swarm-app is composed of multiple projects). Building the code with the compiler from the IDE is the big advantage from running a swarm-app out of the IDE rather than using java -jar or mvn run when developing an application. And in the end from the JAR over the WAR approach. I mean, if i have to run a full-cycle build the overhead of building the WAR too isn't that big...

I think that this is also the common way of all maven and gradle IDE integrations when handling multiproject builds (one build has multiple java projects).

You're example with core and examples in the same eclipse workspace / idea project is another usecase where you combine multiple multiproject builds to a big build. But I think we can handle this the same way. I would stick with the classpath the buildtool is providing to the IDE.

The user.dir is very error prone - maybe deleting this filter would help. But this is just my point of view. I don't know if there is a special maven usecase which needs this filter.

kenfinnigan:
I've found that in Intellij the user.dir is dependent on both the project directory and classpath module as specified in the run settings.

If those aren't pointing to the right spots, you can easily get inconsistent behavior

mhofstetter:
@kenfinnigan That's why I have to manually set the "working directory" in my IDEA run config to a completely different folder to work around this 'user.dir' - filter.

kenfinnigan:
Interesting.

When I have it set to the directory of a project that contains a main(), or which I want to run Swarm() from, it works fine

mhofstetter:
It depends on the example / application you want to run. If you add your relevant classes to the archive via shrinkwrap in the main() it's fine. Then you just need the addAllDeps to add all 3rd party dependencies. Like in the jaxrs-shrinkwrap example. 

If the project is a little bit bigger and you don't want to add all classes manually, then the addAllDeps have to add all my dependencies. Otherwise the classpath isn't complete - which results in an deployment error when I'm lucky or in a frustrating search for the error.

mhofstetter:
ok - maybe a constellation like the one in the jaxrs-shrinkwrap example was the initial reason for the "user.dir" filter. If you add your necessary classes manually to the war-archive without the filter, the classes of the project with the main-class in it would be duplicated on the classpath. The project with the main-class in it would be treated like a JAR in the lib-folder of the WAR.

In the end it's up to you how you define the method "addAllDependencies". Making it somehow configurable or splitting it up into smaller pieces would maybe help. 
- add3rdPartyDeps (all JAR's in java.class.path except jre-libs)
- addProjectDeps (all directories in java.class.path except "main-project")
- addMainProject ("main-project" directory in java.class.path => currently identified by user.dir)

_UPDATE: ok with this approach it's a different behaviour if you run your swarm-app from IDE or with java -jar.._

If the user.dir - function is to mark the "main-project" - there are maybe other solutions to find and filter the corresponding build-directory.

Something like _MainClass_.class.getClassLoader().getResource(_MainClass_.class.getName().replace(".", "/") + ".class") to determine the project output dir where the main-class belongs to. It isn't very elegant but maybe better and less error prone.

Or maybe I'm just using your awesome piece of software the wrong way round and it's just meant to be used in a single java-project and by adding own classes manually via shrinkwrap ... I hope I don't confuse you.

heiko-braun:
ok, the latest PR just retains the outer refactorings, but keeps the resolution logic as it was before. Consider this an interim commit to create another baseline for further discussions

gastaldi:
There is a feature in Maven called `WorkspaceReader` which I believe it may help in this scenario: https://github.com/forge/furnace/blob/master/manager/resolver/maven/src/main/java/org/jboss/forge/furnace/manager/maven/ClasspathWorkspaceReader.java

heiko-braun:
retest this please

heiko-braun:
hm, looks like the PR bot doesn't retest

heiko-braun:
test this please

heiko-braun:
retest this please

heiko-braun:
retest this please

heiko-braun:
retest this please

heiko-braun:
retest this please

heiko-braun:
retest this please

heiko-braun:
retest this please

heiko-braun:
Regarding @mhofstetter's comment:

> ok - maybe a constellation like the one in the jaxrs-shrinkwrap example was the initial reason for the "user.dir" filter. If you add your necessary classes manually to the war-archive without the filter, the classes of the project with the main-class in it would be duplicated on the classpath. The project with the main-class in it would be treated like a JAR in the lib-folder of the WAR.

Maybe we need some agreement on classes versus dependencies. IMO addAllDeps() can mean both or one or the other. So far my understanding of that method was to add maven dependencies, not module contents. But I am not sure anymore. @kenfinnigan what do you think?

heiko-braun:
Another way to look at it, is to clearly separate the use cases.

If we extend the use cases (or make them more explicit as @mhofstetter suggested) then I can think of the following changes to the deployment API's:

a) addAllDependencies(): as before, just pulling in maven dep's
b) addProjectClasses(): pulling in `target/classes` based on `user.dir`
c) addProjectResources(): pulling in target/classes/<META-INF|WEB-INF> based on `user.dir`

a) with my current fixes solves SWARM-565.
b) and c) would address a separate issue, namely the inconvenience for packaging project source and resources into shrinkwrap archives.

bobmcwhirter:
+1

I dunno the current behavior but addAllDeps() originally meant 3rd party
deps, not the user's primary .jar.

Your proposed granularity seems to make sense to me.

Bonus points if we also have one that takes a Predicate filter.

addProjectClasses( e->{.... Return a bool} );

## 

Sent from my iPhone

On Jul 20, 2016, at 6:33 AM, Heiko Braun notifications@github.com wrote:

If we extend the use cases (or make them more explicit as @mhofstetter
https://github.com/mhofstetter suggested) then I can think of the changes
to the deployment API's:
a) addAllDependencies(): as before, just pulling in maven dep's
b) addProjectClasses(): pulling in target/classes based on user.dir
c) addProjectResources(): pulling in target/classes/ based on user.dir

a) with my current fixes solves SWARM-565.
b) and c) would address a separate issue, namely the inconvenience for
packaging project source and resources into shrinkwrap archives.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
https://github.com/wildfly-swarm/wildfly-swarm/pull/57#issuecomment-233913394,
or mute the thread

heiko-braun:
@bobmcwhirter good idea. Each project type (packaging, fractions, etc) brings certain contents (classes, META-INF/WEB-INF resources) and it would be good if we can simplify the shrinkwrap packaging with auto discovery and placement of those files. Maybe prebuild predicates can help here too?

heiko-braun:
ok, I am going to merge this one so I can close SWARM-565. And would suggest to proceed the discussion around further changes on a different thread.

heiko-braun:
The packaging utils issue (my point b) and c) from above) can be found here: https://issues.jboss.org/browse/SWARM-569

PR #59, opened 2016-08-02
Changed the logic which gradle archive-task gets wrapped in the swarm…

mhofstetter:
Changed the logic which gradle archive-task gets wrapped in the swarm-jar to a more controlled one.

With the current logic the latest added task with type 'Jar' wins. That's fine in the following scenarios:
- jar-packaging: just the task Jar with the name 'jar' from the java-plugin ==> Task 'jar' gets wrapped
- war-packaging: task with the 'jar' from the java-plugin and 'war' from the war-plugin ==> war-Tasks gets wrapped

In the following scenario the latest added jar-tasks results in an wrong swarm-jar.
- jar-packaging with a custom archive-task from type "Jar" e.g. source-jar ==> source-jar gets wrapped.

With the new logic the 'war' task wins over the 'jar' task. If the user likes to wrap another archive-task he has to add this explicitly in the swarmextension.

The following snipped illustrates how a source-jar from the type Jar is added to a project - which results in an wrong swarm-jar.
`task sourcesJar(type: Jar, dependsOn: classes) {
    classifier = 'sources'
    from sourceSets.main.allSource
}`

### DIFF
--- a/plugin/src/main/java/org/wildfly/swarm/plugin/gradle/PackagePlugin.java
+++ b/plugin/src/main/java/org/wildfly/swarm/plugin/gradle/PackagePlugin.java
@@ -15,8 +15,13 @@
  */
 package org.wildfly.swarm.plugin.gradle;
 
+import org.gradle.api.GradleException;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
+import org.gradle.api.plugins.JavaBasePlugin;
+import org.gradle.api.plugins.JavaPlugin;
+import org.gradle.api.plugins.WarPlugin;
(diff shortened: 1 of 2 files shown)


gastaldi:
That looks interesting. Could you create an issue in https://issues.jboss.org/browse/SWARM and specify in this PR? 

Thank you!

mhofstetter:
Here is the issue: https://issues.jboss.org/browse/SWARM-580

heiko-braun:
Ok to test

heiko-braun:
add to whitelist

heiko-braun:
@bobmcwhirter any thoughts on this? Can it be merged?

PR #60, opened 2016-08-04
SWARM-578: Introduce FileSystemLayout abstraction

heiko-braun:
Supports the IDE:run cases by providing abstraction to resolve gradle and maven specific file systems locations.

### DIFF
--- a/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/FileSystemLayout.java
+++ b/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/FileSystemLayout.java
@@ -1,0 +1,88 @@
+package org.wildfly.swarm.bootstrap.util;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.UUID;
+
+/**
+ * Build tool filesystem abstraction for use in IDE:run cases.
+ *
+ * @author Heiko Braun
+ * @since 02/08/16
+ */
+public abstract class FileSystemLayout {
+
+
+    public abstract String determinePackagingType();
+
+    public abstract Path resolveBuildClassesDir();
+
+    public abstract Path resolveBuildResourcesDir();
+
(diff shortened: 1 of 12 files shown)


heiko-braun:
@bobmcwhirter Here's the abstraction we talked about the other day. It's only used for the IDE:run cases. In the most simple case, it will implicitly use the 'user.dir' system property:

```
fsLayout = FileSystemLayout.create();
Path p = fsLayout.resolveBuildClassesDir();
[...]
```

kenfinnigan:
Would it make sense to move the FS layout classes from /bootstrap to /container?

there's no guarantee that the bootstrap module will be loaded in anything other than an uber jar situation, as container module only has an optional dep on bootstrap module. And every fraction definitely has a dependency on the container module.

bobmcwhirter:
Container does have a maven dep on bootstrap so it should transitive
around.

The module dep is optional only because the bootstrap module doesn't have a
real module.xml and that freaks out the fraction plugin. It's always
available.

## 

Sent from my iPhone

On Aug 4, 2016, at 5:56 AM, Ken Finnigan notifications@github.com wrote:

Would it make sense to move the FS layout classes from /bootstrap to
/container?

there's no guarantee that the bootstrap module will be loaded in anything
other than an uber jar situation, as container module only has an optional
dep on bootstrap module. And every fraction definitely has a dependency on
the container module.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
https://github.com/wildfly-swarm/wildfly-swarm/pull/60#issuecomment-237508404,
or mute the thread

heiko-braun:
I can't really remember why I moved it there. But I agree, `/container` seems to be more sensible.

heiko-braun:
retest this please

gastaldi:
:lgtm:

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/wildfly-swarm/wildfly-swarm/60#-:-KOKL6GWlY9HiwLDYHEN:1417088721)_

PR #68, opened 2016-08-11
SWARM-594 - Make DefaultDeploymentFactory into CDI components.

bobmcwhirter:
Since we're registering most runtime side components via CDI now
might as well remove another instance of ServiceLoader'ing.
Allows for removal/forgetting the related META-INF/services/
file.  Also may open up cross-injecting FSLayout bits as necessary
in order to build an archive from maven/gradle/whatever layout.


bobmcwhirter:
test this please

PR #70, opened 2016-08-11
Replaced @Singleton with @ApplicationScoped

gastaldi:
Based on https://gist.github.com/bobmcwhirter/6c271c6fbbb783df606ea4fa572df652, @Singleton should be replaced with @ApplicationScoped


bobmcwhirter:
Test this please

gastaldi:
Retest this please

gastaldi:
Ok to test

gastaldi:
Please retest

bobmcwhirter:
Yah, turns out we do _not_ want to do this.

Since some of these interfaces are in spi:api (not :runtime), we don't want to Proxy them (because @ApplicationScoped proxies) because then we have to mix weld3 into the spi:api.

We may want/need to move them to :runtime, but that could also affect plans for user-space usage of these interfaces in the future.

PR #72, opened 2016-08-12
SWARM-597 - Clean up RuntimeServer bootstrapping.

bobmcwhirter:
It's entirely possible to remove the UnmanagedInstance that
we used to bridge and subvert Weld.  So we did.

Additionally, clarified the Server and Deployer contracts,
letting the Deployer deploy, and the Server serve.  Even though
it's actually more like a container.

Additionally, use CDI to @Inject the RuntimeDeployer into
the RuntimeServer.

Additionally, use CDI to @Inject components needed by RuntimeDeployer.

Additionally, remove stuff from RuntimeServer that is now provided
directly to the RuntimeDeployer via CDI.

Additionally, remove the poorly-thought-out non-eager opening
Undertow stuff for the time being.

Additionally, be certain to use the word "additionally" often,
because Friday.


bobmcwhirter:
test this please

PR #73, opened 2016-08-15
SWARM-599: Remote JMX not working.

bobmcwhirter:
Add dependencies to jmx fraction to include both
remoting and management, as that's what's required in order
for remote JMX to function.

Additionally, ensure that the correct remotingjmx module is
included into the solution of the module tree.

Add a test to verify that a remote client can indeed contact
and use the remote JMX interface.

### DIFF
--- a/jmx/module.conf
+++ b/jmx/module.conf
@@ -1,1 +1,2 @@
 org.jboss.as.jmx
+org.jboss.remoting-jmx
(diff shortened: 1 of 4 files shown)


bobmcwhirter:
test this please

gastaldi:
Ok to test

PR #75, opened 2016-08-15
SWARM-600 and SWARM-599 - Remote JMX over (management/http/remoting)

bobmcwhirter:
Previous fix for SWARM-599 transitively brought in org.wildfly.swarm:management,
which then consistently opens port 9990, which is too eager.

Now, bringing in JMX only brings in the remoting subsystem, which by
default does not open any additional ports.

The remoting subsystem will/can use either the management or standard
interfaces for performing remoting over http for things such as
remote-jmx.

Therefore, to use remote-jmx, and if you desire to do it over the
management or HTTP interfaces, you must additionally specify the
dependency of org.wildfly.swarm:management or at least org.wildfly.swarm:undertow.

In the event you have an extremely lightweight application that has
no requirement for :management nor :undertow, and you have enabled
remote JMX, then the legacy remoting port, defined by the configuration
property of swarm.remoting.port, and defaults to 4777 will be opened.

In the event all of the above are available within your application,
no remote JMX will occur at all unless you enable the remote connector
on the JMXFraction.

If you generically enable it, it will prefer management port, falling
back to undertow, falling back to the legacy remoting port.

If you specifically enable it to use the management interface, and
management is not available, it will disable the remote connector for
JMX and will _not_ fall back to either undertow or the legacy remoting
port, because you were explicit in asking for management interface.

### DIFF
--- a/jmx/module.conf
+++ b/jmx/module.conf
@@ -1,2 +1,4 @@
 org.jboss.as.jmx
 org.jboss.remoting-jmx
+org.jboss.logging
+org.wildfly.swarm.remoting
--- a/jmx/pom.xml
+++ b/jmx/pom.xml
@@ -43,11 +43,21 @@
     </dependency>
     <dependency>
       <groupId>org.wildfly.swarm</groupId>
-      <artifactId>remoting</artifactId>
+      <artifactId>logging</artifactId>
     </dependency>
     <dependency>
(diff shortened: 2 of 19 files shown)


bobmcwhirter:
ok to test

tremes:
This could be good food for the doc (I can take a look because I think there's not much about JMX). Anyway thanks for explanation.

PR #78, opened 2016-08-15
Convert the container module to use i18n messages.

jamezp:
This adds JBoss Logging and JBoss Logging Tools dependencies to the container module. The messages have been converted to use the interface message bundle for internationalization support.

Messages that just wrapped and rethrew other messages I did not convert. In some use cases it shouldn't need to be done, but I didn't look extensively at all use cases where it was done.

### DIFF
--- a/container/src/test/java/org/wildfly/swarm/internal/ArtifactManagerTest.java
+++ b/container/src/test/java/org/wildfly/swarm/internal/ArtifactManagerTest.java
@@ -120,7 +120,7 @@ public class ArtifactManagerTest {
             manager.artifact(String.format("%s:%s:%s", ARTIFACT_INVALID_GROUP_ID, ARTIFACT_INVALID_ARTIFACT_ID, ARTIFACT_INVALID_VERSION));
             fail("RuntimeException should have been thrown");
         } catch (RuntimeException e) {
-            assertThat(e).hasMessage("Artifact 'no.such:thingy:1.0.0' not found.");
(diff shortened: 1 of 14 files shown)


container/src/test/java/org/wildfly/swarm/internal/ArtifactManagerTest.java
```diff
@@ -120,7 +120,7 @@ public void testArtifactInvalidGroupArtifactVersion() throws Exception {
             manager.artifact(String.format("%s:%s:%s", ARTIFACT_INVALID_GROUP_ID, ARTIFACT_INVALID_ARTIFACT_ID, ARTIFACT_INVALID_VERSION));
             fail("RuntimeException should have been thrown");
         } catch (RuntimeException e) {
-            assertThat(e).hasMessage("Artifact 'no.such:thingy:1.0.0' not found.");
+            assertThat(e).hasMessage("WFSWARM0013: Artifact 'no.such:thingy:1.0.0' not found.");
```
jamezp, line 123: Note tests like this will break once translations are available.

jamezp:
I forgot to add this does not solve the internationalization problem with the `BootstrapLogger`. That will need to be solved in a different way.

bobmcwhirter:
You rock!

kenfinnigan:
ok to test

kenfinnigan:
doh, ignore me, it already tested it

kenfinnigan:
Thanks for this @jamezp!

PR #79, opened 2016-08-16
update to keycloak 2.1.0.Final

sebastienblanc:


bobmcwhirter:
Have you manually verified that the standalone keycloak-swarm.jar continues to work?

sebastienblanc:
@bobmcwhirter yep, I have both tested that keycloak-server works and keycloak fraction with my test apps. But I will double check with Keycloak team if `module.conf` needs to be updated

sebastienblanc:
@stianst is there any updates in the modules between kc 1.9.x and 2.x ? If I quickly compare I can see the `keycloak-authz-x`  modules but these are added transitively with `keycoak-services`

stianst:
There was some new modules introduced by authz other than that I can't think of any

bobmcwhirter:
We do use their feature-pack, so hopefully they are all just available to us.  I think this is good to go.

PR #82, opened 2016-08-16
SWARM-608 - Display fraction stability at boot

bobmcwhirter:
- Motivation:

Since we do not communicate "quality" through release
versioning (no Alpha or Beta or Final), we maintain this
information on a per-fraction basis within each fraction.
- Modifications:

Upgrade the fraction-plugin which exposes the per-fraction
information via META-INF/fraction.properties.  Use this
information to display quality indicators upon process
startup.
- Result:

Each installed fraction, along with its quality is displayed
each time a -swarm process boots.

In the case of STABLE or above, the log message is INFO level.
Else, it's a WARN.


gastaldi:
Please retest this

PR #83, opened 2016-08-17
stageConfig url ignored

Athou:
build 2016.8.1 is ignoring the stage config file when calling swarm.withStageConfig(url)

### DIFF
--- a/container/src/main/java/org/wildfly/swarm/Swarm.java
+++ b/container/src/main/java/org/wildfly/swarm/Swarm.java
@@ -211,7 +211,7 @@ public class Swarm {
     public Swarm withStageConfig(URL url) {
         this.stageConfigUrl = Optional.of(url);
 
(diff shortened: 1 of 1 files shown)


kenfinnigan:
The presence check is like that so that setting a system property for the staging file will overwrite any other means of setting it.

You may even find that without calling swarm.withStageConfig() it's picked up your stages file off the classpath anyway.

If we incorporate this change, then doing swarm.withStageConfig() wins every time, which is not what is intended.

Maybe you could raise an issue https://issues.jboss.org/browse/SWARM about your use case and we can discuss if the original loading logic needs to be altered

Athou:
I removed the whole withStageConfig() call and the file is indeed picked up off the classpath, solving my issue. 

Thanks!

kenfinnigan:
No problem, glad its working

PR #86, opened 2016-08-17
SWARM-611 - Remove some (but not all) WELD circular warnings.

bobmcwhirter:
- Motivation:

Warnings are ugly.
- Modifications:

By denoting one end of the circular as a normal scope (and thus
proxy'd), instead of @Dependent, we can avoid the warning.
- Result:

Half as many warnings.


gastaldi:
Ok to test

PR #90, opened 2016-08-18
SWARM-609

kenfinnigan:
Added back some of the logging that had been removed as it was using println, but added using the @LogMessage format.

Also added "org.wildfly.swarm" category, as that's the logger category we're using, into logging.properties with a default INFO level but it is overridable with -Dswarm.logging={LEVEL}


gastaldi:
👍

PR #105, opened 2016-08-23
SWARM-619: Enable HTTP/2 in the HTTP/S connectors

gastaldi:

---
## Motivation

Enables HTTP/2 in the existing HTTP/S listeners.
## Modifications

Created an HTTP2Customizer to add the necessary parameters to the existing HTTP/S listeners
## Result

When HTTP/S is enabled, any resource accessed using Google Chrome or Mozilla Firefox should use the h2 protocol

### DIFF
--- a/undertow/src/main/java/org/wildfly/swarm/undertow/UndertowFraction.java
+++ b/undertow/src/main/java/org/wildfly/swarm/undertow/UndertowFraction.java
@@ -53,14 +53,14 @@ public class UndertowFraction extends Undertow<UndertowFraction> implements Frac
 
     public UndertowFraction applyDefaults() {
         server(new Server("default-server")
-                .httpListener("default", (listener) -> {
-                    listener.socketBinding("http");
-                })
-                .host(new Host("default-host")))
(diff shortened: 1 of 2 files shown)


undertow/src/main/java/org/wildfly/swarm/undertow/UndertowFraction.java
```diff
@@ -53,14 +53,14 @@ public void postConstruct() {
```
kenfinnigan, line 53: Are there changes in this file or is it just formatting?

Maybe update and remove formatting?

gastaldi: No changes, just formatting from IntelliJ IDEA. 
Could you do a little test by formatting this class in your IDE to check if it changes the indentation too? I think my IDE might be configured differently from the others.

kenfinnigan: Mine did the same, so we'll leave it

PR #107, opened 2016-08-24
SWARM-616 - Repacked .war shouldn't fail on Windows

bobmcwhirter:
## Motivation

Part of moving the .war to .war.original before repackaging
is failing on Windows.
## Modifications

Be more explicit about deleting the old .war.original and
the .war when moving/copying/repackaging.

Also, close streams when checksumming WEB-INF/lib contents.
## Result

Hopefully it works on Windows.


bobmcwhirter:
Superceded by https://github.com/wildfly-swarm/wildfly-swarm/pull/108

PR #113, opened 2016-08-25
SWARM-571: Arquillian does not include src/main/resource/modules

heiko-braun:

---
## Motivation

module.xml has not been discovered in arquillian tests so the build tool didn't package additonal module automatically.
## Modifications

Change the UberjarSimpleContainer to attempt an auto detection if an external reference to "swarm.build.modules" is not given"
## Result

modules declared in src/main/resources will now be automatically packaged into the uberJar created for the arq test execution.

### DIFF
--- a/arquillian/api/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
+++ b/arquillian/api/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
@@ -130,7 +130,15 @@ public class UberjarSimpleContainer implements SimpleContainer {
                 .fractionDetectionMode(BuildTool.FractionDetectionMode.never)
                 .bundleDependencies(false);
 
-        final String additionalModules = System.getProperty(SwarmInternalProperties.BUILD_MODULES);
+        String additionalModules = System.getProperty(SwarmInternalProperties.BUILD_MODULES);
+
+        // See https://issues.jboss.org/browse/SWARM-571
+        if(null==additionalModules) {
+            // see if we can find it
+            File modulesDir = new File("target/classes/modules");
+            additionalModules = modulesDir.exists() ? modulesDir.getAbsolutePath() : null;
+        }
+
         if (additionalModules != null) {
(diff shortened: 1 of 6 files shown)


arquillian/api/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
```diff
@@ -130,7 +130,15 @@ public void start(Archive<?> archive) throws Exception {
...
-        final String additionalModules = System.getProperty(SwarmInternalProperties.BUILD_MODULES);
+        String additionalModules = System.getProperty(SwarmInternalProperties.BUILD_MODULES);
+
+        // See https://issues.jboss.org/browse/SWARM-571
+        if(null==additionalModules) {
+            // see if we can find it
+            File modulesDir = new File("target/classes/modules");
+            additionalModules = modulesDir.exists() ? modulesDir.getAbsolutePath() : null;
+        }
+
```
kenfinnigan, line 141: Is it worth combining sys property with target/classes/modules?

heiko-braun: Not sure what you mean ...

kenfinnigan: Currently if the system property is set, nothing from target/classes/modules would be included.

Maybe there should be multiple additionalModule dirs added?

heiko-braun: I was thinking about that as well. But that would mean we need to prevent duplicate locations to be added through both means. Therefore I've decided to make it mutual exclusive for now.

heiko-braun: @kenfinnigan Do you want to hold back the PR for this or is it good to merge?

PR #114, opened 2016-08-25
SWARM-628: Introduce configuration to generate self-signed certificate

gastaldi:

---
## Motivation

This is more focused on quick-enabling HTTP/s for development purposes.
HTTP/2 support was introduced in SWARM-619 and is only enabled on HTTP/s listeners.
## Modifications

Introduced two new configuration options: `swarm.https.certificate.generate` and `swarm.https.certificate.generate.host`
## Result

When the new configuration option is set, it will generate a self-signed certificate and enable HTTP/s
for development and testing out-of-the-box

### DIFF
--- a/spi/src/main/java/org/wildfly/swarm/spi/api/SwarmProperties.java
+++ b/spi/src/main/java/org/wildfly/swarm/spi/api/SwarmProperties.java
@@ -58,6 +58,17 @@ public interface SwarmProperties {
     String HTTPS_PORT = "swarm.https.port";
 
     /**
+     * If true, generates a self-signed certificate for development purposes
+     */
+    String HTTPS_GENERATE_SELF_SIGNED_CERTIFICATE = "swarm.https.certificate.generate";
+
+    /**
+     * The host used in the self-signed certificate if {@link SwarmProperties#HTTPS_GENERATE_SELF_SIGNED_CERTIFICATE} is true
+     * Defaults to localhost
+     */
+    String HTTPS_GENERATE_SELF_SIGNED_CERTIFICATE_HOST = "swarm.https.certificate.generate.host";
+
+    /**
      * The context path to be used, defaults to /
      */
     String CONTEXT_PATH = "swarm.context.path";
--- a/testsuite/testsuite-undertow/src/test/java/org/wildfly/swarm/undertow/HTTPSCustomizerSelfSignedCertificateTest.java
+++ b/testsuite/testsuite-undertow/src/test/java/org/wildfly/swarm/undertow/HTTPSCustomizerSelfSignedCertificateTest.java
@@ -1,0 +1,61 @@
+/**
+ * Copyright 2015-2016 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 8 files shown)


spi/src/main/java/org/wildfly/swarm/spi/api/SwarmProperties.java
```diff
@@ -58,6 +58,11 @@
     String HTTPS_PORT = "swarm.https.port";
 
     /**
+     * If true, generates a self-signed certificate for development purposes
+     */
+    String HTTPS_GENERATE_SELF_SIGNED_CERTIFICATE = "swarm.https.generate_self_cert";
```
kenfinnigan: Should this use all dots instead of hyphens?

Following the same pattern as `swarm.ds.connection.url`

undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HTTPSCustomizer.java
```diff
@@ -44,38 +48,58 @@
...
+                            tmpDir = TempFileManager.INSTANCE.newTempDirectory("wildfly-swarm-data", ".d");
+                            System.setProperty("jboss.server.data.dir", tmpDir.getAbsolutePath());
+                        } catch (IOException e) {
+                            // Ignore
+                        }
+                    }
+                    management.securityRealm("SSLRealm", (realm) -> {
+                        realm.sslServerIdentity((identity) -> {
+                            identity.keystorePath("application.keystore")
+                                    .keystorePassword("password")
```
kenfinnigan: Should this be using fraction.keystorePath() and the same for password?

gastaldi: I was assuming that keystorePath() and password wouldn't be provided in this case

spi/src/main/java/org/wildfly/swarm/spi/api/SwarmProperties.java
gastaldi: I'm fine with any choice.

testsuite/testsuite-undertow/src/test/java/org/wildfly/swarm/undertow/HTTPSCustomizerSelfSignedCertificateTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @CreateSwarm
+    public static Swarm newSwarm() throws Exception {
+        System.setProperty(SwarmProperties.HTTPS_GENERATE_SELF_SIGNED_CERTIFICATE, "true");
+        return new Swarm()
+                .fraction(UndertowFraction.createDefaultFraction())
+                .fraction(ManagementFraction.createDefaultFraction());
+    }
+
+    @Test(expected = SSLHandshakeException.class)
```
kenfinnigan, line 55: I'm not an expert on SSL, but should this test throw a handshake exception if a self signed cert is installed?

gastaldi: Yes, because it is not part of the trusted known keystore

spi/src/main/java/org/wildfly/swarm/spi/api/SwarmProperties.java
```diff
@@ -58,6 +58,17 @@
     String HTTPS_PORT = "swarm.https.port";
 
     /**
+     * If true, generates a self-signed certificate for development purposes
+     */
+    String HTTPS_GENERATE_SELF_SIGNED_CERTIFICATE = "swarm.https.certificate.generate";
```
heiko-braun, line 63: Could this also become part of the UndertowFraction API instead of using properties? 

i.e. `UndertowFraction#generateCert(String host)`

PR #115, opened 2016-08-25
SWARM-615: CDI.current() fails to find provider

kenfinnigan:

---
## Motivation

Being able to access CDI.current() is critical for some use cases and this behavior is a regression from 1.0.0.Final to 2016.8
## Modifications

Add weld subsystem as an optional dependency to javax.enterprise.api module so that when user application uses CDI that module will be present and CDI.current() set.

Also added a test to verify its presence so that we don't get a regression
## Result

CDI.current() is available in a user application that uses CDI.


bobmcwhirter:
LGTM

PR #118, opened 2016-08-29
SWARM-639 - Unified meta-data for fractions/applications.

bobmcwhirter:

DO NOT MERGE YET.  This is completely dependent upon:

https://github.com/wildfly-swarm/wildfly-swarm-fraction-plugin/pull/12

Also, I have not fully-tested the DeltaSpike or Microprofile issues, which I'd like to do.  But I offer this PR as more than likely "complete" modulo fraction-plugin release and adjustment to the root pom.xml.

---
## Motivation

Beause of the variety of .conf files, we have a difficult time
actually keeping up with all the bootstrap artifacts, bootstrap
modules, and which dependencies are actually called for by the
application (instead of simply by WFS internally).
## Modifications

All of the .conf files have been replaced in the uber-jar with
a singular META-INF/wildfly-swarm-application.yaml.  It is a yaml
file with the following useful bits of information:
- a list of bootstrap artifacts (our fraction jars as mvn coordinates)
- a list of bootstrap modules to mix into swarm.application:main, etc
- a list of actual, bonafide dependencies of the application, minus any
  transitive dependencies only required by WFS itself.

Additionally, to accomodate both the uberjar and classpath modes of executions,
similar information is held within each fraction (under META-INF/fraction-manifest.yaml)
to help make these same decisions when a META-INF/wildfly-swarm-application.yaml
is not actually present (because no uberjar).

Collectively, all of the above bits of data are glommed into a new class,
ApplicationEnvironment, which can answer the specific questions regardless
of execution mode:
- What are the app's dependencies?
- What are the bootstrap modules to add?
## Result

In general, this should result in a no-op visible change, but
simply a cleaner, more organized internal architecture.
Also, addAllDependencies() not adding extraneous items in any
execution mode, and WEB-INF/lib/ of any maven (or gradle) -created
.war being clean of anything except application dependencies.


bootstrap/src/main/java/org/wildfly/swarm/bootstrap/Main.java
```diff
@@ -40,9 +39,14 @@ public Main(String... args) throws Throwable {
     }
 
     public static void main(String... args) throws Throwable {
-        //TODO Move property key to -spi
-        System.setProperty("swarm.isuberjar", Boolean.TRUE.toString());
-        new Main(args).run();
+        try {
+            //TODO Move property key to -spi
+            System.setProperty("swarm.isuberjar", Boolean.TRUE.toString());
```
kenfinnigan: Move to SwarmInternalProperties? or whatever its called. Should be easier with -spi in main code

bobmcwhirter: Ah, yes, thank you.  I forget you'd merged SPI in.  Will do.

kenfinnigan:
@heiko-braun, do you have any comments on this? We'd like to get it merged today as its a large change.

There's also https://github.com/wildfly-swarm/wildfly-swarm-fraction-plugin/pull/12

arquillian/daemon/pom.xml
```diff
@@ -9,22 +9,101 @@
 
   <parent>
     <groupId>org.wildfly.swarm</groupId>
-    <artifactId>arquillian-parent</artifactId>
+    <artifactId>arquillian-daemon-parent</artifactId>
```
gastaldi: This should be arquillian-parent.

heiko-braun:
Thanks @kenfinnigan . As far I as I can tell, it looks good. But to better understand how things relate to each other:  Does each `fraction-manifest.yaml` contribute to `wildfly-swarm-application.yaml`? And in what way?

heiko-braun:
@bobmcwhirter @kenfinnigan But regardless of my prior question, feel free to merge it when needed. I'll need some more time to fully understand the impact of this, but that should not hold back your progress.

heiko-braun:
@bobmcwhirter Can you explain what concepts are new and which ones are merely consolidated?

bobmcwhirter:
From the first/only commit, which is more verbose, particular the last paragraph:

## Motivation

Beause of the variety of .conf files, we have a difficult time
actually keeping up with all the bootstrap artifacts, bootstrap
modules, and which dependencies are actually called for by the
application (instead of simply by WFS internally).

## Modifications

All of the .conf files have been replaced in the uber-jar with
a singular META-INF/wildfly-swarm-application.yaml.  It is a yaml
file with the following useful bits of information:
- a list of bootstrap artifacts (our fraction jars as mvn coordinates)
- a list of bootstrap modules to mix into swarm.application:main, etc
- a list of actual, bonafide dependencies of the application, minus any
  transitive dependencies only required by WFS itself.

Additionally, to accomodate both the uberjar and classpath modes of executions,
similar information is held within each fraction (under META-INF/fraction-manifest.yaml)
to help make these same decisions when a META-INF/wildfly-swarm-application.yaml
is not actually present (because no uberjar).

Collectively, all of the above bits of data are glommed into a new class,
ApplicationEnvironment, which can answer the specific questions regardless
of execution mode:
- What are the app's dependencies?
- What are the bootstrap modules to add?

## Result

In general, this should result in a no-op visible change, but
simply a cleaner, more organized internal architecture.
Also, addAllDependencies() not adding extraneous items in any
execution mode, and WEB-INF/lib/ of any maven (or gradle) -created
.war being clean of anything except application dependencies.

This is has resulted in simplifying access to all of this metadata
by wrapping it up in FractionManifest (per fraction) and
WildFlySwarmMaifest (per -swarm.jar).  These two classes are still
not used directly, but rather contribute to the user-facing singleton
of ApplicationEnvironment.

ApplicationEnvironment has also replaced Layout.

bobmcwhirter:
For the most part, there is nothing particularly _new_, but rather better organized, with duplicates removed, and a more coherent API (ApplicationEnvironment) for accessing information.

ApplicationEnvironment serves to be the switchboard to answer questions that may vary based on Uberjar-vs-Not-Uberjar scenarios.

If UberJar, it mostly is supported by `wildfly-swarm-manifest.yaml` in the -swarm.jar.

If not UberJar, it's mostly supported by the collection of `fraction-manifest.yaml` coming from the fraction jars.

heiko-braun:
@bobmcwhirter isHollow? https://github.com/bobmcwhirter/wildfly-swarm/blob/b63c9c80f65f35399dad7e923d531ff9a71a4ccb/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/env/ApplicationEnvironment.java#L227

heiko-braun:
Should this throw instead of returning an empty collection? https://github.com/bobmcwhirter/wildfly-swarm/blob/b63c9c80f65f35399dad7e923d531ff9a71a4ccb/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/env/ApplicationEnvironment.java#L171

Similar question here (and maybe other cases): https://github.com/bobmcwhirter/wildfly-swarm/blob/b63c9c80f65f35399dad7e923d531ff9a71a4ccb/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/env/ApplicationEnvironment.java#L208

kenfinnigan:
isHollow is for handling an uber jar of just fractions where a WAR is then deployed to it

heiko-braun:
So isHollow() reflects the case of `java -jar *-swarm.jar myApp.war?`

kenfinnigan:
yeah

heiko-braun:
Is this an expression of thread safe singleton access? https://github.com/bobmcwhirter/wildfly-swarm/blob/b63c9c80f65f35399dad7e923d531ff9a71a4ccb/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/env/ApplicationEnvironment.java#L43

bobmcwhirter:
@heiko-braun wrt throw instead of null/empty -- no.  Given the code-paths, the `getDependencies()` is used in the `resolveDependencies()` which defers to your `SystemDependencyResolution` or `MavenDependencyResolution`.  The `SystemDependencyResolution` doesn't use `ApplicationEnvironment.getDependencies()` but rather the previous classpath-based bits as usual.

The `MavenDependencyResolution`, which only fires when uber-jar'd, does use the `getDependencies()` which is answered by the `wildfly-swarm-manifest.yaml` which is only available when uber-jar'd.

bobmcwhirter:
@heiko-braun wrt to the `AtomicReference<>.updateAndGet()`, yes it's basically a thread-safe, atomic singleton pattern to prevent double initialization of the environment.

heiko-braun:
@kenfinnigan @bobmcwhirter cool, thanks for the clarification.

PR #123, opened 2016-08-31
SWARM-575 - stateless-ejb javaee7 sample doesn't work

bobmcwhirter:

---
## Motivation

When trying to apply swarmtool commandline tool to the
stateless-ejb example, it would fail, being unable to
resolve the jta-1.0.1B artifact.  Meanwhile, the plugin
would successfully build the -swarm.jar.  Both modes
of turning a .war into a -swarm.jar should work the
same, so having a difference is bad.
## Modifications

The jta-1.0.1B artifact was a symptom of a larger problem.
We were asking our ResolvingHelpers (maven, gradle and shrinkwrap)
to _transitively_ resolve all the dependencies that BuildTool
figured we needed.  Which was gratuitous and ignored the fact
that the dependencies between module.xml are already handled and
solved, and did not require an additional transitive resolution
step.

Therefore, I added an additional helper.resolveAll(deps, boolean isTransitive)
that allows us to disable transitive resolution when provided with a
list of dependencies to resolve.

By default, the single-arg helper.resolveAll(deps) does work
transitively, still. The interface provides a default implementation
that calls resolveAll(deps, true) to provide the existing, legacy
behaviour of the single-parameter version.

In BuildTool's populateWhateverRepository() methods, we now
take our already-solved list of dependencies, and call resolveAll(deps, false)
to simply ensure that they are present on disk, in the uberjar
or in the user's local M2REPO.
## Result

SwarmTool now can be applied to the stateless-ejb example without
throwing a DependencyResolutionException on jta-1.0.1B.

### DIFF
--- a/arquillian/resolver/src/main/java/org/wildfly/swarm/arquillian/resolver/ShrinkwrapArtifactResolvingHelper.java
+++ b/arquillian/resolver/src/main/java/org/wildfly/swarm/arquillian/resolver/ShrinkwrapArtifactResolvingHelper.java
@@ -34,6 +34,9 @@ import org.jboss.shrinkwrap.resolver.api.maven.repository.MavenChecksumPolicy;
 import org.jboss.shrinkwrap.resolver.api.maven.repository.MavenRemoteRepositories;
 import org.jboss.shrinkwrap.resolver.api.maven.repository.MavenRemoteRepository;
 import org.jboss.shrinkwrap.resolver.api.maven.repository.MavenUpdatePolicy;
+import org.jboss.shrinkwrap.resolver.api.maven.strategy.MavenResolutionStrategy;
+import org.jboss.shrinkwrap.resolver.api.maven.strategy.NonTransitiveStrategy;
+import org.jboss.shrinkwrap.resolver.api.maven.strategy.TransitiveStrategy;
 import org.jboss.shrinkwrap.resolver.impl.maven.ConfigurableMavenWorkingSessionImpl;
 import org.jboss.shrinkwrap.resolver.impl.maven.MavenWorkingSessionContainer;
 import org.wildfly.swarm.spi.api.internal.SwarmInternalProperties;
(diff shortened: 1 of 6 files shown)


arquillian/resolver/src/main/java/org/wildfly/swarm/arquillian/resolver/ShrinkwrapArtifactResolvingHelper.java
```diff
@@ -110,16 +118,48 @@ public ArtifactSpec resolve(ArtifactSpec spec) {
...
+                                        return null;
+                                    }
+
+                                    @Override
+                                    public MavenResolutionFilter[] getResolutionFilters() {
+                                        return new MavenResolutionFilter[]{
+                                                new MavenResolutionFilter() {
+                                                    @Override
+                                                    public boolean accepts(MavenDependency dependency, List<MavenDependency> dependenciesForResolution, List<MavenDependency> dependencyAncestors) {
+                                                        if ( dependency.getVersion().equals( "1.0.1B" ) ) {
```
kenfinnigan: Are these printlns leftover from testing?

bobmcwhirter: Ah dang it. Yes.

kenfinnigan: if you've "left the building" I can fix

bobmcwhirter: I’m here, about to push -f a fix to my PR.

bobmcwhirter: Okay, squashed in another commit that should remove the printlns.

/me double-checks.

PR #126, opened 2016-09-05
SWARM-657: can't run tests using the Arquillian Daemon when another application listens on port 12345

Ladicek:

---
## Motivation

Running tests using the Arquillian Daemon is impossible when another
application listens on port 12345. In my case, that application is
the HipChat client.
## Modifications

Make the Arquillian Daemon's port configurable using a system property.
Change all the arquillian.xml files to use the system property as well.
On both places, 12345 is used as the default value.
## Result

By default, nothing changes. Passing `-Dswarm.arquillian.daemon.port=34567`
to the Maven invocation makes the Arquillian Daemon use a different port.
The tests will pick up the new port number as well, so they run correctly.

### DIFF
--- a/arquillian/daemon/src/main/java/org/wildfly/swarm/arquillian/daemon/DaemonServiceActivator.java
+++ b/arquillian/daemon/src/main/java/org/wildfly/swarm/arquillian/daemon/DaemonServiceActivator.java
@@ -15,6 +15,10 @@
  */
 package org.wildfly.swarm.arquillian.daemon;
 
+import java.net.BindException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 import org.jboss.as.server.deployment.DeploymentUnit;
 import org.jboss.as.server.deployment.Phase;
 import org.jboss.as.server.deployment.Services;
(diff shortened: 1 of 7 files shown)


gastaldi:
Ok to test

arquillian/daemon/src/main/java/org/wildfly/swarm/arquillian/daemon/DaemonServiceActivator.java
```diff
@@ -41,16 +45,27 @@ public void activate(ServiceActivatorContext serviceActivatorContext) throws Ser
     }
 
     static class DaemonService implements Service<Void> {
+        private static final Logger log = Logger.getLogger(DaemonService.class.getName());
+
+        private static final String SWARM_ARQUILLIAN_DAEMON_PORT = "swarm.arquillian.daemon.port";
```
gastaldi: Perhaps it would be nice to move this to SwarmProperties

Ladicek: Makes sense, thanks.

Ladicek: Fixed.

gastaldi:
Looks good to me. Thank you!

PR #131, opened 2016-09-09
SWARM-605 Clustering/Session Replication not working

nekop:

---
## Motivation

The undertow clustering is not working because it's missing required module and cannot create clustered session manager.
## Modifications

Add org.wildfly.clustering.web.undertow module to undertow/module.conf
## Result

Undertow can create clustered session manager for session replication

### DIFF
--- a/undertow/module.conf
+++ b/undertow/module.conf
@@ -1,5 +1,6 @@
 javax.servlet.api export=true
 org.wildfly.extension.undertow
+org.wildfly.clustering.web.undertow
 io.undertow.core
 io.undertow.servlet
 org.ow2.asm


kenfinnigan:
Thanks for taking a look at solving this issue.

Unfortunately we're not entirely sure whether its something we need to solve in WF Swarm or not. We have another issue, https://issues.jboss.org/browse/SWARM-400, to investigate whether it makes sense for us to provide a means of clustering and session replication.

If we do decide to include clustering/session replication, we would want to consider a different way to enable it as this approach would mean anyone using Undertow gets the module for clustering present in their uber jar and enabled whether they wanted it or not.

For now I'm placing a "hold" on this until we decide the best approach.

nekop:
Yes I completely agree that we need to discuss the approach.

heiko-braun:
So this effectively adds:

```
<module xmlns="urn:jboss:module:1.3" name="org.wildfly.clustering.web.undertow">
    <properties>
        <property name="jboss.api" value="private"/>
    </properties>

    <resources>
        <artifact name="org.wildfly:wildfly-clustering-web-undertow:<version>"/>
    </resources>

    <dependencies>
        <module name="javax.servlet.api"/>

        <module name="io.undertow.core"/>
        <module name="io.undertow.servlet"/>
        <module name="org.jboss.as.web-common"/>
        <module name="org.jboss.logging"/>
        <module name="org.jboss.metadata.web"/>
        <module name="org.jboss.msc"/>
        <module name="org.wildfly.clustering.ee.spi"/>
        <module name="org.wildfly.clustering.service"/>
        <module name="org.wildfly.clustering.web.api"/>
        <module name="org.wildfly.clustering.web.spi"/>
        <module name="org.wildfly.extension.undertow"/>
    </dependencies>
</module>
```

heiko-braun:
Which boils down to ~50k:

```
hbraun:~/dev/prj/wildfly-core (WFCORE-1797)$ ll -h /Users/hbraun/.m2/_repository/org/wildfly/wildfly-clustering-web-undertow/10.0.0.Final-SNAPSHOT/wildfly-clustering-web-undertow-10.0.0.Final-SNAPSHOT.jar 
-rw-r--r--  1 hbraun  staff    50K Jan 11  2016 /Users/hbraun/.m2/_repository/org/wildfly/wildfly-clustering-web-undertow/10.0.0.Final-SNAPSHOT/wildfly-clustering-web-undertow-10.0.0.Final-SNAPSHOT.jar
```

heiko-braun:
@kenfinnigan Considering the actual number of bytes added (50K) and the fact that we didn't conclude yet on how to proceed I would suggest to include this change until we deliberately decide to remove or disable the clustering capabilities. But along the lines of compatibility with WF for users migrating from WF proper to Swarm, this change could be seen as an enhancement.

heiko-braun:
test this please

kenfinnigan:
It's not about the size of the dependency that's added.

It's more a question of whether WF Swarm should support any clustering.

SWARM-400 needs to be evaluated before anything further happens with this and the infinispan PR

kenfinnigan:
As per https://groups.google.com/d/msg/wildfly-swarm/tHJ6e0ifKGg/K8n2DSBMAQAJ, we've decided not to incorporate clustering support into WildFly Swarm

PR #140, opened 2016-09-12
Refactor code and remove the comments (swarm-674)

cmoulliard:
- Refactor code to use Swarm instead of Container
- Remove new Fraction() as the method "public void initialize(InitContext initContext) {" is not longer available
- Add test case for withStageConfigUrl


gastaldi:
Ok to test

PR #141, opened 2016-09-13
SWARM-667: Dependencies not found in Maven Profile

kenfinnigan:

---
## Motivation

When WildFly Swarm dependencies are present only in a Maven Profile, they are unable to be found when building the container for Arquillian tests.
## Modifications

Check for presence of `env.MAVEN_CMD_LINE_ARGS` which contains the Maven command line. Then extract the parts relating to profiles to pass it into ShrinkwrapResolver when loading pom.xml.
## Result

When dependencies are present in a Maven profile, Arquillian tests will find them when run on the command line.

### DIFF
--- a/arquillian/api/src/main/java/org/wildfly/swarm/arquillian/adapter/MavenProfileLoader.java
+++ b/arquillian/api/src/main/java/org/wildfly/swarm/arquillian/adapter/MavenProfileLoader.java
@@ -1,0 +1,52 @@
+/**
+ * Copyright 2015-2016 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.arquillian.adapter;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.jboss.shrinkwrap.resolver.api.maven.ConfigurableMavenResolverSystem;
+import org.jboss.shrinkwrap.resolver.api.maven.PomEquippedResolveStage;
+
+/**
+ * @author Ken Finnigan
+ */
+public final class MavenProfileLoader {
+
+    private static Pattern profilePattern = Pattern.compile("-P([\\w\\-,]+)");
+
+    private static String[] profiles = new String[]{};
+
+    private static boolean profilesDiscovered = false;
+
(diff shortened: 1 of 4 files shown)


arquillian/api/src/main/java/org/wildfly/swarm/arquillian/adapter/MavenProfileLoader.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.jboss.shrinkwrap.resolver.api.maven.ConfigurableMavenResolverSystem;
+import org.jboss.shrinkwrap.resolver.api.maven.PomEquippedResolveStage;
+
+/**
+ * @author Ken Finnigan
+ */
+public final class MavenProfileLoader {
+
+    private static Pattern profilePattern = Pattern.compile("-P([\\w\\-,]+)");
```
gastaldi: private static final

arquillian/api/src/main/java/org/wildfly/swarm/arquillian/adapter/MavenProfileLoader.java
```diff
@@ -0,0 +1,52 @@
...
+    private static boolean profilesDiscovered = false;
+
+    private MavenProfileLoader() {
+    }
+
+    public static PomEquippedResolveStage loadPom(ConfigurableMavenResolverSystem resolver) {
+        if (!profilesDiscovered) {
+            String mavenArgs = System.getProperty("env.MAVEN_CMD_LINE_ARGS");
+
+            final Matcher matcher = profilePattern.matcher(mavenArgs);
```
gastaldi: What if mavenArgs is null?

arquillian/api/src/main/java/org/wildfly/swarm/arquillian/adapter/MavenProfileLoader.java
```diff
@@ -0,0 +1,52 @@
...
+import org.jboss.shrinkwrap.resolver.api.maven.PomEquippedResolveStage;
+
+/**
+ * @author Ken Finnigan
+ */
+public final class MavenProfileLoader {
+
+    private static Pattern profilePattern = Pattern.compile("-P([\\w\\-,]+)");
+
+    private static String[] profiles = new String[]{};
```
gastaldi: No difference but `new String[0]` or just `{}` looks nicer

arquillian/api/src/main/java/org/wildfly/swarm/arquillian/adapter/MavenProfileLoader.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private static boolean profilesDiscovered = false;
+
+    private MavenProfileLoader() {
+    }
+
+    public static PomEquippedResolveStage loadPom(ConfigurableMavenResolverSystem resolver) {
```
gastaldi, line 38: Why not moving this method to `org.wildfly.swarm.arquillian.resolver.ShrinkwrapArtifactResolvingHelper` ?

kenfinnigan: It might be worth doing, but couldn't think of a good way to do it that wasn't more ugly than what I have

PR #146, opened 2016-09-19
SWARM-695: Bootstrap fails due to empty default deployment

heiko-braun:

---
## Motivation

The DeploymentFactories have been moved into each respective fraction (i.e. Undertow -> WARDeploymentFactory). This means that without a corresponding fraction, the factory will not be available and in some cases the default deployment will be created by the EmptyJARDeplymentFactory
## Modifications

EmptyJARDeplymentFactory contains a single META-INF/MANIFEST.MF to prevent the error in ZipExporterDelegate.
## Result

Although useless, empty deployments now “deploy” correctly.

### DIFF
--- a/container/src/main/java/org/wildfly/swarm/container/runtime/deployments/DefaultDeploymentCreator.java
+++ b/container/src/main/java/org/wildfly/swarm/container/runtime/deployments/DefaultDeploymentCreator.java
@@ -90,9 +90,9 @@ public class DefaultDeploymentCreator {
             jarArchive.addAsManifestResource(new Asset() {
                 @Override
                 public InputStream openStream() {
-                    return new ByteArrayInputStream(new String("Produced by EmptyJARArchiveDeploymentFactory").getBytes());
+                    return new ByteArrayInputStream(new String("Created-By: WildFly Swarm\n").getBytes());
                 }
-            }, ArchivePaths.create("readme.txt"));
+            }, ArchivePaths.create("MANIFEST.MF"));
             return jarArchive;
         }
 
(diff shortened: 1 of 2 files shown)


container/src/main/java/org/wildfly/swarm/container/runtime/deployments/DefaultDeploymentCreator.java
```diff
@@ -82,7 +86,14 @@ public String getType() {
 
         @Override
         public Archive create() throws Exception {
-            return ShrinkWrap.create(JARArchive.class, UUID.randomUUID().toString() + "." + this.type );
+            JARArchive jarArchive = ShrinkWrap.create(JARArchive.class, UUID.randomUUID().toString() + "." + this.type);
+            jarArchive.addAsManifestResource(new Asset() {
```
gastaldi: You can use new StringAsset() instead.

heiko-braun: It's the same, just more bytes. I'll leave it like that.

gastaldi (approved):

container/src/main/java/org/wildfly/swarm/container/runtime/deployments/DefaultDeploymentCreator.java
```diff
@@ -82,7 +86,14 @@ public String getType() {
...
+                @Override
+                public InputStream openStream() {
+                    return new ByteArrayInputStream(new String("Created-By: WildFly Swarm\n").getBytes());
```
gastaldi: Why new String() here?

gastaldi (changes requested):
Please add a test case too

gastaldi (approved):

PR #147, opened 2016-09-19
SWARM-697: Fix StackOverflowError during gradle artifactresolving

mhofstetter:

---
## Motivation

There are 3rd-party libs (even on mavencentral) which have circular dependencies. This shouldn't break the build when building the swarm-jar. It should also be possible to force versions of transitive dependencies.
## Modifications

Used gradle-features to resolve artifacts to a given configuration. This is much faster and can handle the circular dependencies. The detached configuration to resolve dependencies copies the forced modules from the compile-configuration.
## Result

The wildfly-swarm-package task can handle circular 3rd-party libs and respects forced versions from the ResolutionStrategy from the configuration compile.

### DIFF
--- a/plugin/src/main/java/org/wildfly/swarm/plugin/gradle/GradleArtifactResolvingHelper.java
+++ b/plugin/src/main/java/org/wildfly/swarm/plugin/gradle/GradleArtifactResolvingHelper.java
@@ -22,14 +22,11 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
-import org.gradle.api.Action;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.DependencySet;
-import org.gradle.api.artifacts.ResolvedDependency;
-import org.gradle.api.artifacts.repositories.MavenArtifactRepository;
+import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.internal.artifacts.dependencies.DefaultDependencyArtifact;
 import org.gradle.api.internal.artifacts.dependencies.DefaultExternalModuleDependency;
 import org.gradle.api.internal.artifacts.dependencies.DefaultProjectDependency;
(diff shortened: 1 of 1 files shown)


gastaldi:
Can you provide a test case reproducing this issue?

plugin/src/main/java/org/wildfly/swarm/plugin/gradle/GradleArtifactResolvingHelper.java
```diff
@@ -122,22 +110,19 @@ public ArtifactSpec resolve(final ArtifactSpec spec) {
...
-    private Stream<ResolvedDependency> subTree(ResolvedDependency root) {
-        return root.getChildren().stream()
-                .flatMap( e-> fullTree( e ));
+        return config
+            .getResolvedConfiguration()
+            .getFirstLevelModuleDependencies()
+            .stream()
+            .map(dep -> dep.getModuleArtifacts())
+            .flatMap(artifacts -> artifacts.stream())
+            .collect(Collectors.toList());
```
gastaldi, line 125: Why a List and not a Set to prevent duplicates?

plugin/src/main/java/org/wildfly/swarm/plugin/gradle/GradleArtifactResolvingHelper.java
```diff
@@ -87,27 +77,25 @@ public ArtifactSpec resolve(final ArtifactSpec spec) {
...
+            .map(artifact -> new ArtifactSpec("default",
+                                artifact.getModuleVersion().getId().getGroup(),
+                                artifact.getModuleVersion().getId().getName(),
+                                artifact.getModuleVersion().getId().getVersion(),
                                 artifact.getExtension(),
                                 artifact.getClassifier(),
-                                artifact.getFile()))));
-
-        return resolvedSpecs.stream()
-                .filter(a -> !"system".equals(a.scope))
```
mhofstetter, line 103: Filtering out all system-scoped dependencies was unnecessary. All artifacts will have scope "default", the default-name of all detached configurations which the GradleArtifactResolvingHelper is using to resolve artifacts for the dependencies.

I think this is a leftover from the maven-implementation.

plugin/src/main/java/org/wildfly/swarm/plugin/gradle/GradleArtifactResolvingHelper.java
mhofstetter: Gradle will not return duplicates then resolving all the artifacts for the resolved dependency in a previous step. this step from dependency to module-artifact is mostly a 1:1 conversion - because there aren't many modules with multiple artifacts.

mhofstetter:
If by "testcase" you mean a corrupt 3rd-party-dependency with circular dependencies it's the following for example.

`compile group: 'org.apache.xmlgraphics', name: 'batik-anim', version: '1.7'`

Version 1.8 fixed this issue - but as I described in the JIRA-ticket it's not possible to force the version from outside of the plugin.

gastaldi (approved):
Looks good to me

mhofstetter:
Testing gradle-plugins would be much easier when developing the gradle-plugin with gradle. This way you have all the test-support (there are binaries which aren't available on a public maven repo - so you can't include them in a maven build for proper testing). But I'll respect that you are maven-oriented and want your whole build in maven.

heiko-braun (approved):

PR #149, opened 2016-09-21
SWARM-696: Transitive Swarm dependencies in multi module setups

heiko-braun:
## Motivation

In multi-module setups (maven) Swarm fractions deps are brought trough a transtive dependency graph. This pollutes the WEB-INF/lib again and causes the same issues aas reported in SWARM-616
## Modifications

Change the resolver API's to specifiy whether Swarm should be traversed
## Result

Transitive Swarm deps will not be traveresed.

### DIFF
--- a/plugin/src/main/java/org/wildfly/swarm/plugin/gradle/GradleArtifactResolvingHelper.java
+++ b/plugin/src/main/java/org/wildfly/swarm/plugin/gradle/GradleArtifactResolvingHelper.java
@@ -72,7 +72,7 @@ public class GradleArtifactResolvingHelper implements ArtifactResolvingHelper {
     }
 
     @Override
-    public Set<ArtifactSpec> resolveAll(final Set<ArtifactSpec> specs, boolean transitive) throws Exception {
+    public Set<ArtifactSpec> resolveAll(final Set<ArtifactSpec> specs, boolean transitive, boolean defaultExcludes) throws Exception {
         if (specs.isEmpty()) {
             return specs;
         }
--- a/plugin/src/main/java/org/wildfly/swarm/plugin/maven/RepositorySystemSessionWrapper.java
+++ b/plugin/src/main/java/org/wildfly/swarm/plugin/maven/RepositorySystemSessionWrapper.java
@@ -22,11 +22,13 @@ import org.eclipse.aether.RepositoryListener;
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.SessionData;
 import org.eclipse.aether.artifact.ArtifactTypeRegistry;
+import org.eclipse.aether.collection.DependencyCollectionContext;
 import org.eclipse.aether.collection.DependencyGraphTransformer;
 import org.eclipse.aether.collection.DependencyManager;
 import org.eclipse.aether.collection.DependencySelector;
 import org.eclipse.aether.collection.DependencyTraverser;
 import org.eclipse.aether.collection.VersionFilter;
+import org.eclipse.aether.graph.Dependency;
 import org.eclipse.aether.repository.AuthenticationSelector;
 import org.eclipse.aether.repository.LocalRepository;
 import org.eclipse.aether.repository.LocalRepositoryManager;
@@ -42,9 +44,14 @@ import org.eclipse.aether.transfer.TransferListener;
  */
 final class RepositorySystemSessionWrapper implements RepositorySystemSession {
 
-    RepositorySystemSessionWrapper(RepositorySystemSession delegate, DependencyGraphTransformer transformer) {
+    public RepositorySystemSessionWrapper(RepositorySystemSession delegate, DependencyGraphTransformer transformer) {
+        this(delegate, transformer, false);
+    }
+
+    RepositorySystemSessionWrapper(RepositorySystemSession delegate, DependencyGraphTransformer transformer, boolean excludeSwarm) {
         this.delegate = delegate;
         this.transformer = transformer;
+        this.excludeSwarm = excludeSwarm;
     }
 
     @Override
@@ -139,7 +146,7 @@ final class RepositorySystemSessionWrapper implements RepositorySystemSession {
 
     @Override
     public DependencyTraverser getDependencyTraverser() {
-        return delegate.getDependencyTraverser();
+        return excludeSwarm ? new SwarmExcludedTraverser() : delegate.getDependencyTraverser();
     }
 
     @Override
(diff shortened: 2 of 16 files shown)


plugin/src/main/java/org/wildfly/swarm/plugin/gradle/GradleArtifactResolvingHelper.java
```diff
@@ -82,7 +82,7 @@ public ArtifactSpec resolve(final ArtifactSpec spec) {
     }
 
     @Override
-    public Set<ArtifactSpec> resolveAll(final Set<ArtifactSpec> specs, boolean transitive) throws Exception {
+    public Set<ArtifactSpec> resolveAll(final Set<ArtifactSpec> specs, boolean transitive, boolean defaultExcludes) throws Exception {
```
mhofstetter, line 75: The gradle-implementation doesn't respect this parameter?

I ask because I faced the same problems when having swarm-libs in transitive dependencies. The swarm-lib finds it's place in the section "dependencies" in the META-INF/wildfly-swarm-manyfest.yaml and later it results in the WEB-INF/lib of the WAR created by shrinkwrap. (Yes I'm using the JAR-mode and Gradle)

This resulted in strange deployment-errors like 

> 2016-09-21 10:55:03,170 ERROR [stderr](main) org.jboss.weld.exceptions.DeploymentException: WELD-001410: The injection point has non-proxyable dependencies: [BackedAnnotatedField] @Inject private org.wildfly.swarm.container.runtime.RuntimeDeployer.defaultDeploymentCreator

heiko-braun: I wanted to discuss and verify the general approach first before touching the other resolver implementations. But it's good you mention it, I'll make a note to look into it.

But didn't you mention on another thread that the gradle impl. should not be using the maven resolver, but some gradle specific components? Would that work similarly ?

mhofstetter: OK, that makes sense. What I thought I mentioned in another thread, was that the gradle implementation shouldn't rely on the m2 repo. And also in this case, using maven-related classes to solve the problem in the gradle-implementation isn't the best approach. The filtering should happen with the gradle-api.

This and my pull-request https://github.com/wildfly-swarm/wildfly-swarm/pull/147 doesn't relate to each other apart from the fact that I also tried to solve the problem with the gradle-api and not with a custom implementation.

heiko-braun:
rebased with current master that now includes #147

heiko-braun:
As for the pending changes to the gradle resolver, @mhofstetter suggested:

https://docs.gradle.org/current/userguide/dependency_management.html#sub:exclude_transitive_dependencies

> maybe you can programatically add this exclude pattern to the detachedConfig which handles resolving all the dependencies

heiko-braun:
The last commit adds a test case that verifies the multi-module setup works correctly.

heiko-braun:
Thinking further about the Gradle questions above, I think I would like to keep this separate and move on with this fix in just the context of maven (which is what SWARM-696 is really about).  

The current Gradle functionality should not be impacted by these changes, therefore I would suggest to merge this and look into further Gradle enhancements separately.

plugin/src/main/java/org/wildfly/swarm/plugin/maven/RepositorySystemSessionWrapper.java
```diff
@@ -175,4 +190,24 @@ public RepositoryCache getCache() {
...
+
+    private boolean excludeSwarm;
+
+    public class SwarmExcludedTraverser implements DependencyTraverser
+    {
+
+       @Override
+       public boolean traverseDependency(Dependency dependency)
+       {
+          return !SWARM_GROUP_ID.equals(dependency.getArtifact().getGroupId());
```
bobmcwhirter, line 202: Can we do this without playing with groupId string comparison?  I've tried hard in BuildTool/DependencyManager and the fraction-plugin to not assume any particular groupId for a fraction type of thing.

testsuite/testsuite-buildtool/.gitignore
```diff
@@ -0,0 +1,203 @@
+# Created by .ignore support plugin (hsz.mobi)
```
bobmcwhirter: Can we not have a .gitignore deep in the tree, and just promote anything reasonable to the root .gitignore?

plugin/src/main/java/org/wildfly/swarm/plugin/maven/RepositorySystemSessionWrapper.java
heiko-braun: That's a valid concern, especially considering that we change the GAV when moving to product. What would you suggest instead?

testsuite/testsuite-buildtool/.gitignore
heiko-braun: I have to remove that. It was copied over from a reproducer contributed by a user.

plugin/src/main/java/org/wildfly/swarm/plugin/maven/RepositorySystemSessionWrapper.java
bobmcwhirter: In the other places, we've done the possibly-sub-optimal route of figuring out if a jar is a swarm fraction jar by the presence of META-INF/fraction-manfiest.yml at this point.

heiko-braun: @bobmcwhirter At this stage during maven processing the file is not available. Any other suggestions?

bobmcwhirter: I'll ponder, but if the resolved file artifact isn't available at this phase, then we probably have to do yet-another-phase.  In DependencyManager (called from BuildTool), for instance, I do the full transitive solution in phase 1.  Then I remove any dependency that contains fraction-manifest.yml and perform the full resolution again in phase 2.

That ultimately helps me (probably poorly, given this issue) know which set of artifacts are brought in purely because of fractions, and which might've been asked for directly by the project by subtracting one solution from another.

Once there's another dependency though, that transitively brings in fractions, obviously we have a problem.

heiko-braun:
Without this fix, any multi-module maven project will break at runtime, when a module that contains our plugin has a dependency that transitively refers to swarm fractions. A prominent example is booker, but also users have complained about this already.

I know that depending on GAV's bears the risk of braking when we change the main swarm group id's. Obviously not perfect, but from my point of view good enough at this stage to be included in the next release.

testsuite/testsuite-buildtool/.gitignore
heiko-braun: removed

heiko-braun:
@bobmcwhirter Do you think we can move on and merge it for the lack of a better idea at this point? I think the risk of using groupId's is acceptable.

bobmcwhirter:
Looks good to me. Plus if we move to fractions in a special groupId this would be super fine. Or if we must match any groupId of *.fractions maybe to be general and flexible but still string comparison on groupId. 

Go for it.

bobmcwhirter:
I can't merge via phone apparently.

PR #150, opened 2016-09-21
Fix #SWARM-705 - Unable to create -swarm.jar

AndyGee:
This resolves the issue as described by https://issues.jboss.org/browse/SWARM-705

### DIFF
--- a/tools/src/main/java/org/wildfly/swarm/tools/BuildTool.java
+++ b/tools/src/main/java/org/wildfly/swarm/tools/BuildTool.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2015-2016 Red Hat, Inc, and individual contributors.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


gastaldi:
It seems that nothing has changed except having some code extracted as methods, can you elaborate on how is that supposed to fix the issue?

gastaldi:
Ah @emag pointed out the `synchronized` keyword on the methods, so this seems to be a racing issue? Does it happen only in Windows?

gastaldi:
Ok to test

gastaldi (approved):

AndyGee:
Hi guys, hold on this. It has improved somewhat, but I'm still getting this issue.

AndyGee:
Still not found what is holding the file lock on delete, however, the overwrite succeeds at:

repackaged.as(ZipExporter.class).exportTo(file, true);

So the strategy is to ensure we have the 'backup', try and delete, and then write/overwrite the target - This does not break the intention.

gastaldi:
Can you use YourKit or a profiler to identify what's holding the file? Maybe we forgot to close a stream?

AndyGee:
Found and reported the culprit: https://issues.jboss.org/browse/SHRINKWRAP-511

I don't know what the release turnaround is for ShrinkWrap, so I would suggest merging this as a workaround until it is resolved - It provides a consistent build for Windows until then.

gastaldi:
I think ShrinkWrap needs to be updated too?

kenfinnigan:
This is a workaround until ShrinkWrap is released

PR #151, opened 2016-09-21
Add test case covering project with DefaultLoading

cmoulliard:
- Add test case covering project created using DefaultLoading (no config file passed as parameter)


gastaldi:
ProjectStagesTest is marked as @Ignored, so I don't think this test will run

cmoulliard:
> ProjectStagesTest is marked

Yep. it will not work till we have resolved the issue with Weld that I have reported

kenfinnigan:
ok to test

PR #153, opened 2016-09-22
SWARM-710: RuntimeDeployer should perform cleanup when destroyed

gastaldi:

---
## Motivation

RuntimeServer calls deployer.get().stop(), but that doesn't work if the CDI engine is down (when Ctrl+C)
## Modifications

Added a @PreDestroy annotation to RuntimeDeployer.stop() and removed the call to the Instance deployer object in RuntimeServer.stop
## Result

RuntimeDeployer.stop is called when the object is destroyed accordingly


heiko-braun (approved):

heiko-braun:
lgtm. an alternative might be to explicitly control the shutdown of the weld container, rather then leaving it to it's own shutdown hooks. That way we might be able to prevent similar problems in the future.

PR #156, opened 2016-09-28
Swarm 720 - Update Logging Fraction to support the creation of a default File Logging handler

cmoulliard:
- Update Logging Fraction to support the creation of a default File Logging handler
- Add new default values : DEFAULT_FILE_HANDLER_NAME, DEFAULT_LOGGING_FILE_NAME, DEFAULT_LOGGING_DIR
- Design new test cases to support the DSL
- Include package name to also configure the Logger with the level passed

### DIFF
--- a/logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
+++ b/logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
@@ -15,6 +15,7 @@
  */
 package org.wildfly.swarm.logging;
 
+import java.io.File;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.List;
@@ -28,6 +29,7 @@ import org.wildfly.swarm.config.logging.CustomFormatter;
 import org.wildfly.swarm.config.logging.CustomHandler;
 import org.wildfly.swarm.config.logging.FileHandler;
 import org.wildfly.swarm.config.logging.Level;
+import org.wildfly.swarm.config.logging.Logger;
 import org.wildfly.swarm.config.logging.PatternFormatter;
 import org.wildfly.swarm.config.logging.RootLogger;
 import org.wildfly.swarm.config.logging.SyslogHandler;
@@ -35,10 +37,17 @@ import org.wildfly.swarm.spi.api.Fraction;
 import org.wildfly.swarm.spi.api.annotations.MarshalDMR;
 import org.wildfly.swarm.spi.api.annotations.WildFlyExtension;
 
+import static org.wildfly.swarm.logging.LoggingProperties.DEFAULT_COLOR_PATTERN;
+import static org.wildfly.swarm.logging.LoggingProperties.DEFAULT_FILE_HANDLER_NAME;
+import static org.wildfly.swarm.logging.LoggingProperties.DEFAULT_LOGGING_DIR;
+import static org.wildfly.swarm.logging.LoggingProperties.DEFAULT_LOGGING_FILE_NAME;
+import static org.wildfly.swarm.logging.LoggingProperties.DEFAULT_PATTERN;
+
 /**
  * @author Bob McWhirter
  * @author Ken Finnigan
  * @author Lance Ball
+ * @author Charles Moulliard
  */
 @SuppressWarnings("unused")
 @WildFlyExtension(module = "org.jboss.as.logging")
@@ -51,6 +60,9 @@ public class LoggingFraction extends Logging<LoggingFraction> implements Fractio
 
     public static final String COLOR_PATTERN = "COLOR_PATTERN";
 
+    public static LoggingFraction loggingFraction;
+
+    public RootLogger rootLogger = new RootLogger();
 
     public LoggingFraction applyDefaults() {
         Level level = Level.INFO;
@@ -74,6 +86,13 @@ public class LoggingFraction extends Logging<LoggingFraction> implements Fractio
                 .rootLogger(level, CONSOLE);
 
         return this;
+    }
+
+    public static LoggingFraction getInstance() {
+        if (loggingFraction == null) {
+            loggingFraction = new LoggingFraction();
+        }
+        return loggingFraction;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


testsuite/testsuite-logging/src/test/java/org/wildfly/swarm/logging/DefaultFileLoggingArquillianTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Test
+    public void doLogging() throws FileNotFoundException {
+        String message = "testing: " + UUID.randomUUID().toString();
+        Logger logger = Logger.getLogger("org.wildfly.swarm.logging");
+        logger.info(message);
+        assertTrue("File not found: " + logFile, new File(logFile).exists());
+
+        BufferedReader reader = new BufferedReader(new FileReader(logFile));
+        List<String> lines = reader.lines().collect(Collectors.toList());
```
gastaldi: You need to close this reader or use Files.readAllLines(Path) instead

gastaldi:
Add to whitelist

heiko-braun (approved):

gastaldi (approved):

bobmcwhirter:
retest this please

bobmcwhirter:
Would you mind squashing and rebasing this before we merge it?

bobmcwhirter (commented):
This seems to improve things if you have exactly 1 file and 1 package you want logged to it.

Do you reckon we could do something non-static on LoggingFraction to make setting up N files for N packages easier?  Would be chainable from existing logging-fractions setup however.

e.g.

```
myLoggingFraction.logToFile( "foo.txt", "com.foo" ).logToFile( "bar.txt", "org.bar" )
```

etc

cmoulliard:
> Would you mind squashing and rebasing this before we merge it?

No problem.

cmoulliard:
> do something non-static on LoggingFraction

We can do that for sure but that should maybe done globally as all the factory methods are defined statically (createDefaultLoggingFraction, createErrorLoggingFraction, createDebugLoggingFraction, ...) ?

cmoulliard:
> Do you reckon we could do something non-static on LoggingFraction to make setting up N files for N packages easier?

I will work on that and refactor LoggingFraction to support the chaining

logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
```diff
@@ -81,7 +87,7 @@ public LoggingFraction applyDefaults(Level level) {
      *
      * @return The fully-configured fraction.
      */
-    public static LoggingFraction createTraceLoggingFraction() {
```
gastaldi, line 84: Removing static here would break backward compatibilty

logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
```diff
@@ -90,7 +96,7 @@ public static LoggingFraction createTraceLoggingFraction() {
      *
      * @return The fully-configured fraction.
      */
-    public static LoggingFraction createDebugLoggingFraction() {
+    public LoggingFraction createDebugLoggingFraction() {
```
bobmcwhirter: This should remain static.

logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
```diff
@@ -99,7 +105,7 @@ public static LoggingFraction createDebugLoggingFraction() {
      *
      * @return The fully-configured fraction.
      */
-    public static LoggingFraction createErrorLoggingFraction() {
+    public LoggingFraction createErrorLoggingFraction() {
```
bobmcwhirter: As should this.  I think something is decided askew.

logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
cmoulliard: I will switch to use static method & resubmit

logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
cmoulliard: No problem

logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
```diff
@@ -51,6 +60,9 @@
 
     public static final String COLOR_PATTERN = "COLOR_PATTERN";
 
+    public static LoggingFraction loggingFraction;
```
gastaldi, line 63: Why have a static reference here?

logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
```diff
@@ -76,6 +88,13 @@ public LoggingFraction applyDefaults(Level level) {
         return this;
     }
 
+    public static LoggingFraction getInstance() {
```
gastaldi, line 91: I don't think this is needed due to the existing static createXX() methods

gastaldi (changes requested):

logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
cmoulliard: To have one singleton / class of the LoggingFraction. Otherwise, we can't chain the methods.

logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
cmoulliard: To have one singleton / class of the LoggingFraction. Otherwise, we can't chain the methods.

logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
kenfinnigan: How so?

Ideally we want the methods of LoggingFraction to be chainable on an instance, but we don't need to hold an instance to make that happen.

Maybe I'm not understanding how this is meant to work

bobmcwhirter:
It may also help to fully build the wildfly-swarm-examples repository, to witness what does or doesn't break when you make fundamental changes such as this.

logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java
cmoulliard: If you look to the code of the master branch, you will see that an instance of the LogginFraction is created within the static method "createDefaultLoggingFraction" - https://github.com/wildfly-swarm/wildfly-swarm/blob/master/logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java#L121

As now we have a new static method where we will create a defaultFileLoggingFraction, then we should be able to reuse an existing instance of the LoggingFraction instead of creating a new object otherwise when we will chain the 2 static methods within (https://github.com/cmoulliard/wildfly-swarm-core/blob/83ae9cdb1bd3940805e07cf6b705e11a5cfdac8a/testsuite/testsuite-logging/src/test/java/org/wildfly/swarm/logging/FileLoggingInfoLevelArquillianTest.java#L58-L61), then the console will not be created. 

Do you have a better idea to code that ?

bobmcwhirter: I think this is where the hang-up is.

There's two flavors of configuration methods for a given fraction.

The static methods, such as createDebugLoggingFraction() is a helper factory method.  It's a starting point, so you're not starting from scratch with `new LoggingFraction()` and having to do everything by-hand, from nil.

Ultimately, these static methods are doing `new LoggingFraction()`, and then calling a series of _instance_ methods on that object to do work.  Ultimately the fraction is returned to the user (or the container), and no static state is or should be held.  Ever.

When we speak of chaining, that means _instance_ methods, which return _LoggingFraction_ and do work.

Start there.

Once you have friendly instance methods that can be applied to _any_ instance of a `LoggingFraction` object, _then_ you can think about a handy-dandy helper static factory method just to make things easier.

Bottom line:
- Do work in an instance method
- Have that instance method return `this`
- Once you can work with instance methods, _then_ ponder helper factory methods to help people get started for the 80% case.
- Never hold static state. Ever.  Never.  Ever.  It's a smell if you do.

bobmcwhirter: Basically, if a user starts with

```
LoggingFraction logging = new LoggingFraction()
```

or

```
LoggingFraction logging = LoggingFraction.createDebugLoggingFraction()
```

you should be able work with _either_ of those starting points, allowing for

```
logging.logToFile(...) // or whatever
```

heiko-braun:
If this doesn't make progress I'd suggest we close the PR.

kenfinnigan:
@cmoulliard, are there plans to complete the requested changes?

cmoulliard:
We can delete the PR & close the ticket. I propose that we have a more general discussion later about the conventions to follow to use the FluentApi about the fractions.

gastaldi:
Closing PR for now.

PR #158, opened 2016-09-28
SWARM-438 - Programttic API for advertising services.

bobmcwhirter:

---
## Motivation

In the case where a user desires run-time advertisement of
services, as opposed to using archive.advertise() or the
@Advertise annotation, we should provide a way to accomplish
it.
## Modifications

Added a method to the Topology interface, which can be obtained
through JNDI.  topology.advertise(name) will advertise for
http and https (if available) for the current IP address of the
process, the given service-name provided as an argument.

The return value is a handle which may be used to unadvertise()
the service.  Once unadvertised, the service may not be re-used
without obtaining a new handle.
## Result

A user may now, from within a running deployment, advertise
arbitrary services.

### DIFF
--- a/topology/src/main/java/org/wildfly/swarm/topology/AdvertisemetHandle.java
+++ b/topology/src/main/java/org/wildfly/swarm/topology/AdvertisemetHandle.java
@@ -1,0 +1,8 @@
+package org.wildfly.swarm.topology;
+
+/**
+ * @author Bob McWhirter
+ */
+public interface AdvertisemetHandle {
+    void unadvertise();
+}
--- a/topology/src/main/java/org/wildfly/swarm/topology/Topology.java
+++ b/topology/src/main/java/org/wildfly/swarm/topology/Topology.java
@@ -36,6 +36,8 @@ public interface Topology {
 
     void removeListener(TopologyListener listener);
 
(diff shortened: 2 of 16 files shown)


kenfinnigan (approved):

topology/src/main/java/org/wildfly/swarm/topology/Topology.java
```diff
@@ -36,6 +36,8 @@ static Topology lookup() throws NamingException {
 
     void removeListener(TopologyListener listener);
 
+    AdvertisemetHandle advertise(String name);
```
gastaldi, line 39: Small typo in class name? Advertisement

topology/src/main/java/org/wildfly/swarm/topology/AdvertisemetHandle.java
```diff
@@ -0,0 +1,8 @@
+package org.wildfly.swarm.topology;
+
+/**
+ * @author Bob McWhirter
+ */
+public interface AdvertisemetHandle {
```
gastaldi, line 6: Here is the typo: Should read `AdvertisementHandle` (missing the "n")

bobmcwhirter:
Damnit. Will PR a fix tomorrow. Thanks!

PR #168, opened 2016-10-04
SWARM-688 don't list all fractions in info log.

tremes:

---
## Motivation

Currently all fractions are output at the startup of the container in the log.
## Modifications

Log list of fractions under DEBUG level
## Result

List of fractions is logged/available under DEBUG level only

### DIFF
--- a/container/src/main/java/org/wildfly/swarm/container/runtime/ServerBootstrapImpl.java
+++ b/container/src/main/java/org/wildfly/swarm/container/runtime/ServerBootstrapImpl.java
@@ -131,7 +131,7 @@ public class ServerBootstrapImpl implements ServerBootstrap {
         if ( stabilityIndex < 3 ) {
             LOG.warn(SwarmMessages.MESSAGES.availableFraction(manifest.getName(), manifest.getStabilityLevel(), manifest.getGroupId(), manifest.getArtifactId(), manifest.getVersion()));
         } else {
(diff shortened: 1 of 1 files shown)


kenfinnigan:
@tremes the issue linked to wasn't so much about logging all the fractions as DEBUG, but logging the fractions that are internal as DEBUG.

Apologies if the JIRA wasn't clear

tremes:
Ah ok..will update

kenfinnigan:
Thanks!

bobmcwhirter:
09:18 bobmcw: tremes: so, I think if there's not an internal flag on fraction-manifest.yml, you can tell it's an internal fraction (maybe) by lack of a bootstrap-module or other key typically present
09:18 bobmcw: at least that'd catch the module-only fractions that don't have a matching FooFraction class
09:18 bobmcw: but not so much things like :ee or :io
09:19 bobmcw: if you need another flag to denote which fractions are internal, then wildfly-swarm-fraction-plugin can be modified
09:19 bobmcw: so that it adds it to fraction-manifest.yaml

tremes:
I think we can leave it as it is and close this because fractions like `jaxrs-jsonp`, `jaxrs-cdi`, etc are not listed anyway. 

If I understand it correctly we load fraction manifests only for bootstrap-modules in `ApplicationEnvironment#loadFractionManifestsFromUberjar` and e.g `jaxrs-jsonp` is not in bootstrap-modules because its manifest doesn't have any `module` attribute (checked in `DependencyManager#analyzeFractionManifests`). Question aside: Is that intended?

Please correct me if I am wrong.

PR #175, opened 2016-10-05
SWARM-752: Enable Module's MBeanServer

gastaldi:

---
## Motivation

Debugging Module classloaders can be a pain most of the times. JBoss Modules offers some MBeans to help in this (awful) task.
## Modifications

Called ModuleLoader#installMBeanServer using reflection since the method is not public
## Result

Running jconsole you should see the jboss.modules mbeans in a running swarm app

### DIFF
--- a/container/src/main/java/org/wildfly/swarm/Swarm.java
+++ b/container/src/main/java/org/wildfly/swarm/Swarm.java
@@ -17,6 +17,7 @@ package org.wildfly.swarm;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


bobmcwhirter (approved):
Just formatting issues, else, good.

kenfinnigan:
Should this be done in Swarm or lower down?

bobmcwhirter:
Ah, good point.  Maybe in our BootModuleLoader's ctor, so it occurs early on?

gastaldi:
That should be executed just once, is the BootModuleLoader's ctor executed once too?

gastaldi:
Preferably before any JBoss Modules class is used

bobmcwhirter:
our BootModuleLoader ctor should occur even earlier than Swarm.

jboss-modules bootstraps by looking for the boot.module.loader property, and instantiating it to get started.  That's the earliest we can jack in, besides bootstrap.Main, which isn't used when non-uberjar.

And yah, only once for that ctor.

gastaldi:
When calling `ModuleLoader.installMBeanServer` in the `BootModuleLoader` ctor, some log.trace() statements are performed, thus initializing JUL prematurely and screwing up the log output. Any suggestions?

bobmcwhirter:
Ah, then definitely _after_ we do the logmanager setup.

PR #181, opened 2016-10-13
SWARM-768: Honor swarm.project.stage as an environment variable

gastaldi:

---
## Motivation

Docker environments use env vars to pass parameters to docker container
## Modifications

SwarmProperties.PROJECT_STAGE is resolved in System.getenv() if System.getProperty() doesn't return null
## Result

Swarm uses SwarmProperties.PROJECT_STAGE

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/Swarm.java
+++ b/core/container/src/main/java/org/wildfly/swarm/Swarm.java
@@ -464,7 +464,14 @@ public class Swarm {
 
     private void loadStageConfiguration(URL url) {
         List<ProjectStage> projectStages = new ProjectStageFactory().loadStages(url);
-        String stageName = System.getProperty(SwarmProperties.PROJECT_STAGE, "default");
+        String stageName = System.getProperty(SwarmProperties.PROJECT_STAGE);
+        if (stageName == null) {
+            // Try with SWARM_PROJECT_STAGE
+            stageName = System.getenv(SwarmProperties.PROJECT_STAGE.replace('.', '_').toUpperCase());
+        }
+        if (stageName == null) {
+            stageName = "default";
+        }
         ProjectStage stage = null;
         for (ProjectStage projectStage : projectStages) {
             if (projectStage.getName().equals(stageName)) {


core/container/src/main/java/org/wildfly/swarm/Swarm.java
```diff
@@ -464,7 +464,13 @@ private static boolean isFatJar() throws IOException {
 
     private void loadStageConfiguration(URL url) {
         List<ProjectStage> projectStages = new ProjectStageFactory().loadStages(url);
-        String stageName = System.getProperty(SwarmProperties.PROJECT_STAGE, "default");
+        String stageName = System.getProperty(SwarmProperties.PROJECT_STAGE);
```
lordofthejars, line 467: I like the if/else approach but it seems that in Java 8 it is better to use Optionals.

```
Optional.ofNullable(System.getenv(PROJECT_STAGE))
                .orElse(Optional.ofNullable(System.getProperty(PROJECT_STAGE))
                .orElse("default"));
```

core/container/src/main/java/org/wildfly/swarm/Swarm.java
```diff
@@ -464,7 +464,13 @@ private static boolean isFatJar() throws IOException {
...
+        if (stageName == null) {
+            stageName = System.getenv(SwarmProperties.PROJECT_STAGE);
```
lordofthejars: In this case environment variable must be called `swarm.project.stage`, what I would do is a `PROJECT_STAGE.replace('.', '_')` to be more environment fashion.

core/container/src/main/java/org/wildfly/swarm/Swarm.java
gastaldi: Thank you for your feedback, but I believe that the if/else approach is good in this case :+1:

core/container/src/main/java/org/wildfly/swarm/Swarm.java
gastaldi: I agree, I changed it to be `SWARM_PROJECT_STAGE`

core/container/src/main/java/org/wildfly/swarm/Swarm.java
kenfinnigan: What's the reason that Optional isn't a good approach here?

I'm a bit confused as I know @gastaldi has pushed the use of Optionals in other areas.

gastaldi: Not saying that Optional isn't a good approach, I just didn't want to spend much time on rewriting the whole method to use Optional (specially considering it does not belong in this Sprint 😜)

gastaldi: Also the suggested approach makes unnecessary calls if the property value exists, for example, so it should use orElseGet instead if converted

kenfinnigan (approved):

PR #183, opened 2016-10-14
SWARM-762: Use mirror and full proxy configuration in Maven resolving…

dmandalidis:
… helper
## Motivation

Maven resolving helper does not take mirror configuration into account
while it only uses the first proxy from the Maven configuration.
## Modifications

AbstractSwarmMojo was modified to avoid passing the now redundant first proxy
entry. MavenArtifactResolvingHelper was changed to use the mirror
configuration and the correct proxy
## Result

Swarm plugin will correctly use mirror and proxy configuration of Maven
settings.


kenfinnigan:
ok to test

kenfinnigan (approved):

kenfinnigan:
Thanks for providing a fix!

Greatly appreciated

PR #188, opened 2016-10-20
SWARM-478: Add Flyway fraction

gastaldi:

---
## Motivation

Flyway (https://flywaydb.org/) is an open-source database migration tool. It strongly favors simplicity and convention over configuration.
## Modifications

Created a flyway fraction that automatically bounds to the primary DataSource if not configured or uses the provided URL/user/pass otherwise.
It registers a ServletContextListener in the deployed app to perform the migration. Scripts must be placed under src/main/resources/db/migration
## Result

Flyway can now be activated in Swarm applications by simply adding the flyway fraction and dropping the scripts in the specified place.

### DIFF
--- a/fractions/flyway/src/main/java/org/wildfly/swarm/flyway/FlywayFraction.java
+++ b/fractions/flyway/src/main/java/org/wildfly/swarm/flyway/FlywayFraction.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2016 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.flyway;
+
+import org.wildfly.swarm.spi.api.Fraction;
+import org.wildfly.swarm.spi.api.annotations.DeploymentModule;
+
+/**
+ * Flyway Fraction
+ *
+ * @author <a href="mailto:ggastald@redhat.com">George Gastaldi</a>
+ */
+@DeploymentModule(name = "org.wildfly.swarm.flyway", slot = "deployment")
+public class FlywayFraction implements Fraction<FlywayFraction> {
+
+    /**
+     * Uses the specified connection info if not <code>null</code>. Otherwise
+     * use primary Datasource
+     */
+    private String jdbcUrl;
+    private String jdbcUser;
+    private String jdbcPassword;
+
+    public String jdbcPassword() {
+        return jdbcPassword;
+    }
+
+    public String jdbcUrl() {
+        return jdbcUrl;
+    }
+
+    public String jdbcUser() {
+        return jdbcUser;
+    }
+
+    public FlywayFraction jdbcPassword(String jdbcPassword) {
+        this.jdbcPassword = jdbcPassword;
+        return this;
+    }
+
+    public FlywayFraction jdbcUser(String jdbcUser) {
+        this.jdbcUser = jdbcUser;
+        return this;
+    }
+
+    public FlywayFraction jdbcUrl(String jdbcUrl) {
+        this.jdbcUrl = jdbcUrl;
+        return this;
+    }
+
+    public boolean usePrimaryDataSource() {
+        return this.jdbcPassword == null;
+    }
+}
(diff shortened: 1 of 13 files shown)


gastaldi:
test this please

gastaldi:
Retest this please

fractions/flyway/src/main/java/org/wildfly/swarm/flyway/FlywayFraction.java
```diff
@@ -0,0 +1,71 @@
...
+import org.wildfly.swarm.spi.api.annotations.DeploymentModule;
+import org.wildfly.swarm.spi.api.annotations.DeploymentModules;
+
+/**
+ * Flyway Fraction
+ * 
+ * @author <a href="mailto:ggastald@redhat.com">George Gastaldi</a>
+ */
+@DeploymentModules({
+        @DeploymentModule( name = "org.flywaydb"),
```
bobmcwhirter: I'm not completely sure, and not worth redoing this PR, but I think since you have export=true in the module.conf, you don't need to add org.flywaydb module to the export.

But I could be wrong.  Worth experimenting in a subsequent round.  But as said, not worth refiguring this PR.

fractions/flyway/src/main/java/org/wildfly/swarm/flyway/runtime/FlywayMigrationArchivePreparer.java
```diff
@@ -0,0 +1,71 @@
...
+import org.wildfly.swarm.undertow.descriptors.WebXmlContainer;
+
+/**
+ *
+ * @author <a href="mailto:ggastald@redhat.com">George Gastaldi</a>
+ */
+public class FlywayMigrationArchivePreparer implements ArchivePreparer {
+
+    @Inject
+    private Instance<DatasourcesFraction> dsFractionInstance;
```
bobmcwhirter, line 37: Also, for a future iteration, @Inject @ConfigurationValue for some FlywayProperties.USERNAME and .PASSWORD.  

This would allow it to be used without having a main() configuring the FlywayFraction.  Akin to how we deal with other datasources and such.

```
String thingy = this.injectedThingy.orElse( this.fraction.thingy() )
```

Or whatnot.  So that `project-stages.yml` or properties would suffice to configure in lieu of a `main()`.

bobmcwhirter (approved):
Comments only for future iteration, not to block this PR at all.  Looks great!

fractions/flyway/src/main/java/org/wildfly/swarm/flyway/FlywayFraction.java
bobmcwhirter: Of course, I say "you don't need to add org.flywaydb module to the export" but I mean "you don't need to add org.flywaydb module to the _deployment_"

fractions/flyway/src/main/java/org/wildfly/swarm/flyway/runtime/FlywayMigrationArchivePreparer.java
bobmcwhirter: and I might actually put all of that into a Customizer, to sort through your various options and maybe @Produce some `FlywayConfiguration` that gets injected into a simpler `ArchivePreparer` with less logic.  But I'm not Monday quarterbacking at this point.

fractions/flyway/src/main/java/org/wildfly/swarm/flyway/FlywayFraction.java
```diff
@@ -0,0 +1,70 @@
...
+import org.wildfly.swarm.spi.api.Fraction;
+import org.wildfly.swarm.spi.api.annotations.DeploymentModule;
+import org.wildfly.swarm.spi.api.annotations.DeploymentModules;
+
+/**
+ * Flyway Fraction
+ * 
+ * @author <a href="mailto:ggastald@redhat.com">George Gastaldi</a>
+ */
+@DeploymentModules({
```
bobmcwhirter: Also, `@DeploymentModule` is a repeatable annotation.  You never actually have to explicitly use the `@DeploymentModules` plural form.  That's just useful to the JVM to gather up the repeated singular forms.

pom.xml
```diff
@@ -13,7 +13,6 @@
     <version>7</version>
   </parent>
 
-  <groupId>org.wildfly.swarm</groupId>
```
bobmcwhirter, line 16: The `groupId` is useful to have in order to avoid irritating the `maven-release-plugin` if I recall.  Or maybe just the `maven-version-plugin`.  But one gets confused if `groupId` is implicit and not explicit.

testsuite/testsuite-flyway/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.wildfly.swarm</groupId>
+    <artifactId>testsuite</artifactId>
+    <version>2016.11.0-SNAPSHOT</version>
+    <relativePath>../</relativePath>
+  </parent>
+
+  <artifactId>testsuite-flyway</artifactId>
```
bobmcwhirter, line 17: likewise `groupId` explicitness is good

fractions/flyway/src/main/java/org/wildfly/swarm/flyway/runtime/FlywayMigrationArchivePreparer.java
gastaldi: I agree could you @Issue that enhancement?

PR #190, opened 2016-10-21
SWARM-781 - Implicitly add beans.xml if required.

bobmcwhirter:

---
## Motivation

A user includes our :cdi fraction, they shouldn't also be
required to ensure there's an empty beans.xml.

But if they do provide one, we should respect it.
## Modifications

An archive-preparer that checks for and adds a beans.xml
if one is missing from the deployment, if the CDI fraction
is involved.
## Result

No more requirement of an empty beans.xml.

### DIFF
--- a/fractions/javaee/cdi/src/main/java/org/wildfly/swarm/cdi/runtime/CDIArchivePreparer.java
+++ b/fractions/javaee/cdi/src/main/java/org/wildfly/swarm/cdi/runtime/CDIArchivePreparer.java
@@ -1,0 +1,38 @@
+package org.wildfly.swarm.cdi.runtime;
+
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.Node;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.wildfly.swarm.spi.api.ArchivePreparer;
+
+/** Ensures that a <code>beans.xml</code> is present in the archive.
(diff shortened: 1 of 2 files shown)


fractions/javaee/cdi/src/main/java/org/wildfly/swarm/cdi/runtime/CDIArchivePreparer.java
```diff
@@ -0,0 +1,41 @@
...
+import org.wildfly.swarm.spi.api.ArchivePreparer;
+import org.wildfly.swarm.spi.api.JARArchive;
+
+/**
+ * @author Bob McWhirter
+ */
+public class CDIArchivePreparer implements ArchivePreparer {
+
+    public static final String META_INF_BEANS_XML = "META-INF/beans.xml";
+    public static final String WEB_INF_CLASSES_BEANS_XML = "WEB-INF/classes/beans.xml";
```
kenfinnigan: I think the spec says it should be WEB-INF/beans.xml

bobmcwhirter: Ah, I was going from http://stackoverflow.com/questions/8991561/where-should-beans-xml-be-placed

bobmcwhirter:
retest this please

kenfinnigan (approved):

PR #191, opened 2016-10-24
SWARM-491: DeclaredDependencies, ResolvedDependencies ( and GradleBuildAdapter)

heiko-braun:
## Motivation

Arquillian test did no work in gradle builds, this led to larger review of the BuildTool, DependencyManager, Maven / Gradle Plugin and ArquillianAdapter components
## Modifications
- The contract between DeclaredDependencies and ResolvedDependencies has been made more explicit (but requires further work)
- A gradle build adapter has been introduced to produce the DeclaredDependencies of gradle project (fixes the root cause)
- Archive.getAllDependencies() has been moved to the runtime phase. The API is the same, but the actual artifact resolution happens dring the runtime stages
- DependencyManager has been refactored to make the different processing stages more obvious with regard to the pre/post conditions and invariants. It's in a reasonable shape, but requires further work to cleanly split Declared/Resolved dependencies and eleminate the recursive processing of that data
- Numerous smaller changes to cleanup and improve readability and maintenance (i.e. class visibility levels, terms used, etc)
## Result

Everything works as before, but the code is more structured (hopefully) and in addition to this gradle build can now leverage arquillian

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # ignore .svn metadata files
+wip-notes.txt
 .svn
 # ignore Maven generated target folders
 ~
@@ -56,3 +57,4 @@ tmp
 log
 *.log
 embedded-cassandra.log*
+diagrams
--- a/arquillian/adapter/pom.xml
+++ b/arquillian/adapter/pom.xml
@@ -95,6 +95,11 @@
       <artifactId>arquillian-config-impl-base</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.gradle</groupId>
+      <artifactId>gradle-tooling-api</artifactId>
+      <version>3.0</version>
+    </dependency>
 
     <dependency>
       <groupId>junit</groupId>
@@ -116,5 +121,15 @@
 
   </dependencies>
 
+
+  <repositories>
+    <repository>
+      <id>gradle-tooling</id>
+      <url>http://repo.gradle.org/gradle/libs-releases-local</url>
+      <releases>
+        <enabled>true</enabled>
+      </releases>
+    </repository>
+  </repositories>
 
 </project>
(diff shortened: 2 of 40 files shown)


heiko-braun:
arq

heiko-braun:
package

heiko-braun:
retest this please

heiko-braun:
@bobmcwhirter I think based on this, we can start a review.

heiko-braun:
@bobmcwhirter I'll leave it to you to review/merge.

kenfinnigan:
Looks like there's some variation from the code styles in some files, could that be fixed?

Also, I think we need to add license header to new files

heiko-braun:
@kenfinnigan I've added the license headers, but regarding the codestyles I need some pointers. Skimming through the changeset nothing jumps to my eye.

kenfinnigan:
Couple of ones I noticed were https://github.com/wildfly-swarm/wildfly-swarm/pull/191/files#diff-19ea8f5ef8720c17e258f006837c82b1R17 and https://github.com/wildfly-swarm/wildfly-swarm/pull/191/files#diff-7497a7e221ac79d07edb8b9cd6eee7cbR87

bobmcwhirter (commented):
So, I'd like to approve this, but having merged it against master locally, while the core repo builds, I get a failure on -examples.

The `camel/camel-mail` run in non-uberjar format gives:

`CamelMailIT.testIt:38 [Should contain 'Bound camel naming object: java:jboss/camel/context/mail-context'] actual value:<0> should be greater than:<0>`

I haven't dug into as to why, and this could be something specific to me.  But worth verifying.

I do realize it's Camel, which is always a yak-hole.

heiko-braun:
@bobmcwhirter thanks, i'll have a look. That test did fail for me every once in a while i was working on this.

heiko-braun:
@bobmcwhirter I think the error in camel-mail is unrelated, but brought forward by my changes. It looks like the test driver executes before the swarm server has been started (i.e. you can see that when running with `-Dmaven.failsafe.debug=true`). This executes the assertion that checks the log lines, but since the server isn't fully started yet, it blows up.  A less ideal, but working solution, is to add a client wait time (see https://github.com/wildfly-swarm/wildfly-swarm-examples/pull/102) which fixes the problem. 

Anyway, if you agree, I would suggest we merge this and look into a proper fix for camel-mail in the next sprint.

bobmcwhirter:
Okay. That's an area I want to improve this week with arq anyhow. +1.

heiko-braun:
@bobmcwhirter All squashed into one.

bobmcwhirter (approved):

PR #197, opened 2016-10-25
SWARM-778 - Continued simplification and cleaning of our ARQ tests.

bobmcwhirter:

---


arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/DefaultDeploymentScenarioGenerator.java
```diff
@@ -68,8 +68,8 @@
             try {
                 Class<?> cls = cl.loadClass(className);
                 codeSources.add(cls.getProtectionDomain().getCodeSource());
-            } catch (ClassNotFoundException e) {
-                e.printStackTrace();
+            } catch (ClassNotFoundException|NoClassDefFoundError e) {
+                //e.printStackTrace();
```
kenfinnigan, line 72: Should this log something instead?

kenfinnigan (approved):

arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/DefaultDeploymentScenarioGenerator.java
bobmcwhirter: Since this uses a pretty basic heuristic to find classes, whatever classes that might throw either of these ends up being a pointless (internal) class anyhow.  Mostly because we're testing our own stuff, and things like `org.wildfly.swarm.cdi.**` matches not just our tests and our test "app", but also the internal fraction.

This is one reason I'm inserting `test` sub-packages into some of the test trees, just to avoid the whole issue.

The heuristic is basically "add all classes from the classpath (sorta) from the package of the `testClass`, recursively".

So, `org.wildfly.swarm.cdi` attempts to pick up our fraction bits, which works, until we also try to `loadClass(..)` and it failing is pointless anyhow, since it wasn't an important class for the deployment itself.

But `org.wildfly.swarm.cdi.test` and below avoids the whole issue.

PR #199, opened 2016-10-25
SWARM-791: Add jboss-repository back.

gastaldi:

---
## Motivation

In SWARM-770, it was observed that removing this repository causes collateral damages
## Modifications

The commit was reverted and the code was slighly improved
## Result

The ArtifactResolutionException no longer happens

### DIFF
--- a/plugin/src/main/java/org/wildfly/swarm/plugin/maven/MavenArtifactResolvingHelper.java
+++ b/plugin/src/main/java/org/wildfly/swarm/plugin/maven/MavenArtifactResolvingHelper.java
@@ -17,6 +17,7 @@ package org.wildfly.swarm.plugin.maven;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/wildfly/swarm/plugin/maven/MavenArtifactResolvingHelper.java
```diff
@@ -54,15 +56,18 @@
...
 
     public MavenArtifactResolvingHelper(ArtifactResolver resolver,
-                                        RepositorySystem system,
-                                        RepositorySystemSession session) {
+            RepositorySystem system,
+            RepositorySystemSession session) {
         this.resolver = resolver;
         this.system = system;
         this.session = session;
+        this.remoteRepositories.add(buildRemoteRepository("jboss-public-repository-group",
```
bobmcwhirter, line 64: Seems like a lot of other changes in this commit also?

bobmcwhirter (approved):

plugin/src/main/java/org/wildfly/swarm/plugin/maven/MavenArtifactResolvingHelper.java
gastaldi: Here is a summary of the changes: 
- Refactored the method by removing the session parameter (since it's an attribute) 
- Removed the duplicate call to `resolveArtifacts` and used `resolveDependencies` instead of `collectDependencies`
- Used `Objects::nonNull` instead of `x->x!= null`

PR #216, opened 2016-11-09
SWARM-714: Hide `jaxrs-cdi` fraction from developer

kenfinnigan:

-----

Motivation
----------
Having to know to include a new dependency just because an application uses jaxrs and cdi is not a good developer experience.

We want such a combination to be implicitly included

Modifications
-------------
Due to its small size, added jaxrs-cdi fraction into jaxrs by default but modified the module.xml rules to exclude weld modules as they will be brought in if cdi dependency is chosen.

Result
------
No longer necessary to select jaxrs-cdi dependency when already including jaxrs and cdi.


gastaldi:
Should the `jaxrs-cdi` be marked as `internal` to hide from the generator too?

kenfinnigan:
Hadn't thought of that, might not be a bad idea

gastaldi (approved):

PR #217, opened 2016-11-10
SWARM-610: The wildfly-self-contained.d directory lingers

heiko-braun:

-----
Motivation
Some files lingers in the tmp directory when the process stops

Modifications

Change the WF Core StandaloneContainer to manage the tmp directory explicitly and provide a stop() to shutdown the container.

Result

All allocated tmp files will be properly removed when Swarm stops.

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java
@@ -209,6 +209,7 @@ public class RuntimeServer implements Server {
     public void stop() throws Exception {
 
         this.container.stop();
+        this.container=null;
         this.client = null;
         this.deployer = null;
     }
--- a/pom.xml
+++ b/pom.xml
@@ -43,6 +43,7 @@
     <version.org.jboss.msc.jboss-msc>1.2.6.Final</version.org.jboss.msc.jboss-msc>
     <version.jboss-logmanager-ext>1.0.0.Alpha3</version.jboss-logmanager-ext>
 
+
     <version.org.arquillian>1.1.10.Final</version.org.arquillian>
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -43,6 +43,7 @@
     <version.org.jboss.msc.jboss-msc>1.2.6.Final</version.org.jboss.msc.jboss-msc>
     <version.jboss-logmanager-ext>1.0.0.Alpha3</version.jboss-logmanager-ext>
 
+
```
heiko-braun, line 46: This shouldn't be here, but since I am travelling this week, maybe we can just it ignore it?

core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java
```diff
@@ -181,13 +181,10 @@ private void configureUserSpaceExtensions() {
     }
 
     public void stop() throws Exception {
-        final CountDownLatch latch = new CountDownLatch(1);
-        this.serviceContainer.addTerminateListener(info -> latch.countDown());
-        this.serviceContainer.shutdown();
-
-        latch.await();
```
kenfinnigan: As we're not using this.serviceContainer as part of the stop() anymore, do we even need to hang onto the object?

Looks like, from a quick glance, that we can remove the instance from the class and use locally on start

heiko-braun: yes, good idea

heiko-braun:
i've included @kenfinnigan suggestions. retest this please

heiko-braun:
Ah damn, i've pushed the changes to #218. So if you merged this one before #218, you get the changes included.

PR #220, opened 2016-11-11
(Hold) SWARM-739 (and others) Improve configuration.

bobmcwhirter:

-----

Motivation
----------

Configuration is a mish-mash of values from fractions, properties
and YAML files.  Defaults are scattered to the winds. This is sub-optimal.

Modifications
-------------

Replace many fraction instance attributes with newly-created
Configurable<T>, which respect the idea of YAML wins over
user-set-on-Fraction wins over defaults.

Lessens the need to do specific value inspection and logic
within Customizers and such.

Configurable<T> can currently be used (and discovered) in:

  - Fractions
  - Customizers
  - ArchivePreparers
  - ArchiveMetadataProcessors

Usually they will live within a Fraction but that is not always
appropriately.

Result
------

Cleaner and more logical configuration, simpler customizers. More
information logged when a process starts.


bobmcwhirter:
Please review, but also please hold for the time-being. 

I just wanted to give it to CI and provide an opportunity for inspection by others on the team.

core/container/src/main/java/org/wildfly/swarm/container/runtime/cdi/ConfigurationClassExtension.java
```diff
@@ -0,0 +1,159 @@
+/**
+ * Copyright 2015-2016 Red Hat, Inc, and individual contributors.
```
bobmcwhirter, line 2: Note to self: delete this class

core/container/src/main/java/org/wildfly/swarm/container/runtime/ConfigurableManager.java
```diff
@@ -0,0 +1,153 @@
...
+        if (subresources) {
+            scanSubresources(instance);
+        }
+    }
+
+    protected void scan(Object instance, Class<?> curClass) throws IllegalAccessException {
+        if (curClass == null) {
+            return;
+        }
+        Field[] fields = curClass.getDeclaredFields();
```
gastaldi, line 64: What about fields from superclasses?

core/container/src/main/java/org/wildfly/swarm/container/runtime/LogSilencer.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public class LogSilencer {
+
+    private Logger logger;
+    private Level originalLevel;
+
+    public LogSilencer(String logger) {
+        try {
+            java.util.logging.LogManager lm = LogManager.getLogManager();
+            this.logger = lm.getLogger( "org.jboss.weld" );
```
gastaldi, line 19: logger should be used instead

gastaldi (approved):
Looks good to me

core/container/src/main/java/org/wildfly/swarm/container/runtime/ConfigurableManager.java
bobmcwhirter: 15 lines down we recourse with the instance and it's superclass.

gastaldi: Hah, you're right. Ignore that comment

core/container/src/main/java/org/wildfly/swarm/container/runtime/ConfigurableManager.java
```diff
@@ -0,0 +1,153 @@
...
+
+    protected void scan(Object instance, boolean subresources) throws IllegalAccessException, InvocationTargetException {
+        scan(instance, instance.getClass());
+        if (subresources) {
+            scanSubresources(instance);
+        }
+    }
+
+    protected void scan(Object instance, Class<?> curClass) throws IllegalAccessException {
+        if (curClass == null) {
```
gastaldi, line 61: curClass == null || curClass == Object.class

bobmcwhirter: Good point. And I can remove the recursion altogether with a plain loop.

core/container/src/main/java/org/wildfly/swarm/container/runtime/LogSilencer.java
gastaldi: I mean, use the logger parameter instead of the fixed value org.jboss.weld

PR #221, opened 2016-11-14
SWARM-723: Fraction for FluentD to integrate with EFK stack

heiko-braun:

-----

# Motivation

Allow logging to fluentd daemon

# Modification

Added new fraction `"fluentd"` that tweaks logging to add a custom fluentd log handler

# Result

With the fraction enabled, logs go to system out and fluentd

### DIFF
--- a/fractions/fluentd/src/main/java/org/wildfly/swarm/fluentd/FluentdFraction.java
+++ b/fractions/fluentd/src/main/java/org/wildfly/swarm/fluentd/FluentdFraction.java
@@ -1,0 +1,75 @@
+/**
+ * Copyright 2015-2016 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.fluentd;
+
+import org.wildfly.swarm.config.logging.Level;
+import org.wildfly.swarm.spi.api.Fraction;
+
+/**
+ * @author Heiko Braun
+ */
+public class FluentdFraction implements Fraction<FluentdFraction> {
+
+    public FluentdFraction() {
+        this.hostname = "localhost";
+        this.port = 24224;
+        this.level = Level.INFO;
+    }
+
+    public static Fraction createDefaultFluentdFraction() {
+        return new FluentdFraction();
+    }
+
+    public FluentdFraction level(Level level) {
+        this.level = level;
+        return this;
+    }
+
+    public Level level() {
+        return this.level;
+    }
+
+    public FluentdFraction hostname(String hostname) {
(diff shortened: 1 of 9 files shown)


fractions/fluentd/.gitignore
```diff
@@ -0,0 +1,58 @@
...
+local-repo-eap
+# Dependency reduced pom
+*/dependency-reduced-pom.xml
+#OS X index files
+.DS_Store
+testsuite-core/standalone/cli_output
+tmp
+log
+*.log
+embedded-cassandra.log*
```
kenfinnigan: Is the root .gitignore not sufficient?

fractions/fluentd/src/main/java/org/wildfly/swarm/fluentd/FluentdFraction.java
```diff
@@ -0,0 +1,75 @@
...
+import org.wildfly.swarm.config.logging.Level;
+import org.wildfly.swarm.spi.api.Fraction;
+
+/**
+ * @author Heiko Braun
+ */
+public class FluentdFraction implements Fraction<FluentdFraction> {
+
+    public FluentdFraction() {
+        this.hostname = "localhost";
```
kenfinnigan, line 27: Do we need host/port?

I was under the assumption that FluentD collected logs by default, we didn't have to publish them somewhere?

heiko-braun: AFAIK, it runs as a sidekick process in the docker container. You can pass it log events either through http or tcp.

heiko-braun: See also https://docs.openshift.com/enterprise/3.1/install_config/aggregate_logging.html

kenfinnigan: I'd had a look at those, but my impression was that FluentD was scaled to the exact number of nodes in the cluster because it essentially scrapes the console output to pass to ElasticSearch, but maybe I'm wrong

heiko-braun: Could be yes. I'll reach out to the openshift guys to see how it's intended to be used. Unfortunately the OS3 docs don't mention the client side at all.

But if your assumption is right, what would the purpose of a fluentd fraction then?

kenfinnigan: If it does suck up the logs from console, maybe we can put what you've added behind an `Environment.openshift()` check to only push to host/port when not running inside OpenShift so FluentD can be used outside.

In addition we may want to format log output as JSON, can't recall if you're already doing that, as I think that's the favored approach with FluentD

fractions/fluentd/src/main/java/org/wildfly/swarm/fluentd/runtime/FluentdCustomizer.java
```diff
@@ -0,0 +1,87 @@
...
+    @Any
+    private FluentdFraction fluentdFraction;
+
+    @Inject
+    @Any
+    private LoggingFraction loggingFraction;
+
+    @Inject
+    @ConfigurationValue(FluentdProperties.HOSTNAME)
+    private String hostname;
```
gastaldi: You can use 

```
@Inject 
@ConfigurationValue(FluentdProperties.HOSTNAME)
private Optional<String> hostName;
```

to avoid null checks

gastaldi (approved):

PR #222, opened 2016-11-15
SWARM-739 (and others) Improve configuration.

bobmcwhirter:

-----

Motivation
----------

Configuration is a mish-mash of values from fractions, properties
and YAML files.  Defaults are scattered to the winds. This is sub-optimal.

Modifications
-------------

Introduce a Defaultable<T> for providing fancier and trackable
defaults. Not used terribly many places, but in a few Customizers
and enhanced versions of fraction resources.  Currently used
for indicating that one default value is dependent upon some
other value, such as jndi-name=java:/jms/%{name} or such.

Additionally, provide a @Configurable annotation, to denote
that some arbitrary field should be configurable through the
project-stages/system-properties method, and to specify the
name to bind to that field.

@Configurable will be detected upon the following classes

  - Fractions
  - Customizers
  - ArchivePreparers
  - ArchiveMetadataProcessors

Additionally, any Fraction implementation is deemed to be
*implicitly* configurable, and the root Fraction object is
walked through its entire subresources to discover and name
configurable items.

In addition to allowing overriding values set by default
or through a main() using the project-stages.yml file,
completely new items may be created simply through the
specification of the appropriate configuration values.

e.g.

    swarm:
      datasources:
        data-sources:
          MyDS:
            driver-name: myh2
            connection-url: jdbc:h2:mem:test;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE
            user-name: sa
            password: sa
        jdbc-drivers:
          myh2:
            driver-class-name: org.h2.Driver
            xa-datasource-name: org.h2.jdbcx.JdbcDataSource
            driver-module-name: com.h2database.h2

Is exactly the equivalent of:

    datasourcesFraction.dataSource("MyDS");
    datasourcesFraction.jdbcDriver("myh2");

followed by the abovementioned setting of properties within
the fraction tree as though the user had invoked the above
two lines before applying configuration values.  In the
event that .jdbcDriver(key, consumer) exists and applies
extra logic (in the case of drivers, for instance, we set
a default driver-name to be equivalent to the key for the
resource via DatasourceFraction), that logic still applies,
hopefully resulting in unsurprising behavior.

Result
------

Cleaner, more logical and more consistent configuration, simpler customizers.
More information logged when a process starts.


bobmcwhirter:
Currently within `ConfigurableManager.log()` a boolean flag is set constantly to `true`, which causes all (including `unset`) configurables to be logged.  This is purely a temporary measure so that we can all look at what's set, and think about them, in terms of doing future simplification and winnowing, perhaps.

Basically, I wanted to surface this functionality loudly for a bit.  We can put it behind a property or a flag (defaulting to `verbose=false`) in time for the next release.

heiko-braun:
> any Fraction implementation is deemed to be
> implicitly configurable, and the root Fraction object is
> walked through its entire subresources to discover and name
> configurable items.

You mean the full class tree? I.e. `DataSource` and anything beyond in it's `config-api` Did you do a performance check to depict the impact of this?

heiko-braun:
> completely new items may be created simply through the
> specification of the appropriate configuration values
- I assume this still leads to a DMR representation internally? 
- What if several method to create the DMR model come together? i.e. xml , java and now the yaml way? Any precedence rules given? How are conflicts detected and resolved?

heiko-braun:
> new items may be created

Maybe you can elaborate on this a bit? Technically speaking how is the configuration resources tree created from a yaml representation? I.e how do you make a difference between resources,  resource keys and attributes when translating from yaml to java?

heiko-braun:
As a side note:

When running the example using this branch I get a lot of `swarm.null`s when the `ConfigurableManager` dumps the options:

```
2016-11-16 10:50:01,080 INFO  [org.wildfly.swarm] (main) WFSWARM0037: Configuration:
  swarm.ajp.enable                                                       = false
  swarm.ajp.port                                                         = 8009
  swarm.http.port                                                        = 8080
  swarm.https.port                                                       = 8443
  swarm.null.active-requests                                             = (unset)
  swarm.null.add-logging-api-dependencies                                = (unset)
  swarm.null.alias                                                       = (unset)
  swarm.null.annotation-property-replacement                             = (unset)
  swarm.null.archive-validation.enabled                                  = true
  swarm.null.archive-validation.fail-on-error                            = true
  swarm.null.archive-validation.fail-on-warn                             = true
  swarm.null.bean-validation.enabled                                     = true
  swarm.null.cached-connection-manager.debug                             = (unset)
  swarm.null.cached-connection-manager.error                             = (unset)
  swarm.null.cached-connection-manager.ignore-unknown-connections        = (unset)
  swarm.null.cached-connection-manager.install                           = true
  swarm.null.deep-copy-subject-mode                                      = (unset)
  swarm.null.default-bindings-service.context-service                    = java:jboss/ee/concurrency/context/default
  swarm.null.default-bindings-service.datasource                         = (unset)
  swarm.null.default-bindings-service.jms-connection-factory             = (unset)
  swarm.null.default-bindings-service.managed-executor-service           = java:jboss/ee/concurrency/executor/default
  swarm.null.default-bindings-service.managed-scheduled-executor-service = java:jboss/ee/concurrency/scheduler/default
  swarm.null.default-bindings-service.managed-thread-factory             = java:jboss/ee/concurrency/factory/default
```

heiko-braun:
I was looking at a simple logging use case, trying to change the log level of the root logger, but apparently it's not there?

```
 swarm.logging.add-logging-api-dependencies                                                      = (unset)
  swarm.logging.console-handlers.CONSOLE.autoflush                                                = (unset)
  swarm.logging.console-handlers.CONSOLE.enabled                                                  = (unset)
  swarm.logging.console-handlers.CONSOLE.encoding                                                 = (unset)
  swarm.logging.console-handlers.CONSOLE.filter-spec                                              = (unset)
  swarm.logging.console-handlers.CONSOLE.formatter                                                = (unset)
  swarm.logging.console-handlers.CONSOLE.named-formatter                                          = COLOR_PATTERN
  swarm.logging.pattern-formatters.COLOR_PATTERN.color-map                                        = (unset)
  swarm.logging.pattern-formatters.COLOR_PATTERN.pattern                                          = %K{level}%d{yyyy-MM-dd HH:mm:ss,SSS} %-5p [%c] (%t) %s%e%n
  swarm.logging.root-logger.filter-spec                                                           = (unset)
  swarm.logging.use-deployment-logging-config                   
```

It's seems in the current impl, `RootLogger.level` is not supported because it's type is `org.wildfly.swarm.config.logging.Level`, an enum.

bobmcwhirter:
It walks the full fraction tree of what's created by main() or defaultly.
It tends to be a rather narrow and shallow set of objects.

It supports all classes but only walks objects given to swarm.fraction()
ultimately.

bobmcwhirter:
It ultimately leads to augmenting the Fraction object tree. Which in turn
affects DMR.

For precedence, this occurs last, immediately within swarm.start() after
XML or user created Fractions or the default fractions are done.

heiko-braun:
Issues with default values in the DMR model:

Consider the case of `swarm.undertow.default-server`, which the `ConfigurableManager` consider to be `=(unset)`. In fact it's not, because when the DMR model is instantiated it applies it's own defaults values if needed. These can be seen when looking at `:read-resource-description` of  a resource, i.e.:

```
 "default-server" => {
                "type" => STRING,
                "description" => "The default server to use for deployments",
                "expressions-allowed" => false,
                "nillable" => true,
                "default" => "default-server",
                "min-length" => 1L,
                "max-length" => 2147483647L,
                "access-type" => "read-write",
                "storage" => "configuration",
                "restart-required" => "all-services"
            },
```

It  would be good to make this apparent through the `ConfigurableManager`

bobmcwhirter:
Ah. Definitely not good. I'll figure out why.

This is why I wanted to print them all for now.

bobmcwhirter:
Currently we support simple scalars. Numbers. Strings. Booleans. Enums and
such are not yet supported.

bobmcwhirter:
All of this occurs before DMR marshaling so we don't have the defaults
available to us. We could instill the defaults into the actual config-api.

I also plan to make verbose=false which then does not display any (unset)
configuration items.

heiko-braun:
> We could instill the defaults into the actual confit-api.

Yeah, I was thinking the same. We could add the meta data to the classes generated in the `config-api`. That way it could become available to `ConfigurableManager`

heiko-braun:
> I also plan to make verbose=false which then does not display any (unset)
> configuration items.

Yes, I am wondering how we could make this information available to users. The log works, but some offline docs would nice as well. Maybe we can  add the `ConfigurableManager` to the swarm plugin to you `wildfly-swarm:config` and it dumps the options for all fractions found in the project?

It could even be scoped to single fractions, i.e.:

```
mvn wildfly-swarm:config --fraction management
```

would lead to 

```
These are the available options for fraction 'org.wildfly.swarm:management:2016.12':

 swarm.management.http-interface-management-interface.console-enabled                            = false
  swarm.management.http-interface-management-interface.http-upgrade-enabled                       = true
  swarm.management.http-interface-management-interface.sasl-protocol                              = (unset)
  swarm.management.http-interface-management-interface.secure-socket-binding                      = (unset)
  swarm.management.http-interface-management-interface.security-realm                             = (unset)
  swarm.management.http-interface-management-interface.server-name                                = (unset)
  swarm.management.http-interface-management-interface.socket-binding                             = management-http
  swarm.management.http.port                                                                      = 9990
  swarm.management.https.port                       

```

Thinking further about this, if the scanning could be moved to plugin, we wouldn't have the performance impact at runtime, and the available options could even be included and used by forge or other tools.

bobmcwhirter:
Also, new things are created by simply see what keys exist (via StageConfig
keys with the prefix) and any missing keys are created using the fraction
subresource factory.

fract.someResource("missingKey")

Then we continue the configurable scan like normal with the newly created
item, so same rules apply as if the user or our defaults had created that
subresource.

The general format is

swarm.KEY.attribute

With appropriate recursion if the attribute is itself a list of keyed sub
resources.

bobmcwhirter:
Yah or I think we can wire it up to --help. That flag already prints any
keys found in a project-stages.yml if present.

bobmcwhirter:
And some fraction-plugin metadata could be used to generate docs at our
build time for all available keys.

bobmcwhirter:
HOLD until I can figure out the null in the names in the examples. @heiko-braun-- which example?

I can't apply label with the phone apparently.

heiko-braun:
I've put it on HOLD. 

Try building the examples with `-DskipITs=true` at the top level and you'll see. It's all over the place

bobmcwhirter:
Thanks! And thanks for the in-depth review!

bobmcwhirter:
As to coercion to Level for logging or other fancy stuff, this ultimately delegates to the StageConfig.Resolver so that would also benefit @Inject @ConfigurationValue. 

I think it could also be useful to round-trip INTO the stage config so that users can inject some of these deep values if they need? But I'm less sold on that idea.

bobmcwhirter:
swarm.null solved. Will push soon.

core/container/src/main/java/org/wildfly/swarm/container/runtime/ConfigurableManager.java
```diff
@@ -0,0 +1,456 @@
...
+        add(Float.class);
+        add(Float.TYPE);
+        add(String.class);
+
+        add(Defaultable.class);
+    }};
+
+    private static Logger LOG = Logger.getLogger("org.wildfly.swarm.config");
+
+    private final List<ConfigurableHandle> configurables = new ArrayList<>();
```
heiko-braun, line 62: @bobmcwhirter Do you think that the fact that `ConfigurableManager` holds on to `List<ConfigurableHandle>` can prevent garbage collection of all instances bound to `ConfigurableHandle`?

heiko-braun: I assume the list of `configurables` can become quiet big, so I am wondering if it needs to be released at some point during the server bootstrap, i.e. when when the list bootstrap DMR operations has been build?

bobmcwhirter:
Sure. We only hold them so we can log them.

We can add a clear()

fractions/javaee/remoting/src/main/java/org/wildfly/swarm/remoting/RemotingFraction.java
```diff
@@ -53,4 +54,14 @@ public RemotingFraction applyDefaults() {
...
+
+    public RemotingFraction port(int port) {
+        this.port.set( port );
+        return this;
+    }
+
+    public int port() {
+        return this.port.get();
+    }
+    private Defaultable<Integer> port = integer(4777);
```
Ladicek, line 66: Default value of the legacy remoting port is 4447... and I see that this is wrong in Swarm already. JIRA it is.

heiko-braun (approved):

PR #227, opened 2016-11-17
Using Object::notNull in lambda expressions

gastaldi:

-----

### DIFF
--- a/tools/src/main/java/org/wildfly/swarm/tools/DependencyManager.java
+++ b/tools/src/main/java/org/wildfly/swarm/tools/DependencyManager.java
@@ -26,6 +26,7 @@ import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Objects;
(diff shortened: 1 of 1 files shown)


kenfinnigan:
why no JIRA?

gastaldi:
Because it is too trivial. It was part of my investigation on https://issues.jboss.org/browse/SWARM-847 but does not fix it.

kenfinnigan:
Why is it too trivial?

There's clearly a problem with nulls being passed

gastaldi:
Nulls are already handled in the original code. This is just to use Objects::nonNull instead of e->e!=null

kenfinnigan:
Then why make the change?

To me this is refactoring for no real purpose.

If we want to go through the entire code based to do this kind of refactoring then its something we should discuss. I'd prefer we don't do it piecemeal

kenfinnigan:
Closing this as it's not related to a fix, or JIRA, and if we're going to look at "cleaning up" the code base we should do it on a larger scale than just a single class

PR #230, opened 2016-11-22
SWARM-693, SWARM-514: Health check updates

heiko-braun:
Motivation
----------

Simplify the health check procedurs and update to health check spec 1.0

Modifications
-------------

- common entry point /health (SWARM-514)
- aggregation of results (SWARM-693)
- simplified security

Result
------

User can now provide health checks as before, but an invocation to /health will gather the results from all checks. If security is applied, then it is enforced across all monitoring endpoints and jaxrs resources that act as health checks.

### DIFF
--- a/fractions/io/src/main/java/org/wildfly/swarm/io/IOFraction.java
+++ b/fractions/io/src/main/java/org/wildfly/swarm/io/IOFraction.java
@@ -30,7 +30,11 @@ import org.wildfly.swarm.spi.api.annotations.WildFlyExtension;
 public class IOFraction extends IO<IOFraction> implements Fraction<IOFraction> {
 
     public IOFraction applyDefaults() {
-        return worker(new Worker("default"))
+        return worker(
+                "default", w -> {
+                    w.ioThreads(100);
+                    w.taskMaxThreads(20);
+                })
                 .bufferPool(new BufferPool("default"));
     }
 }
--- a/fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/HttpContexts.java
+++ b/fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/HttpContexts.java
@@ -15,20 +15,44 @@
  */
 package org.wildfly.swarm.monitor.runtime;
 
+import java.io.IOException;
+import java.net.URL;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 
 import javax.enterprise.inject.Vetoed;
 import javax.naming.NamingException;
 
 import io.undertow.attribute.ReadOnlyAttributeException;
 import io.undertow.attribute.RelativePathAttribute;
+import io.undertow.client.ClientCallback;
+import io.undertow.client.ClientConnection;
+import io.undertow.client.ClientExchange;
+import io.undertow.client.ClientRequest;
+import io.undertow.client.ClientResponse;
+import io.undertow.client.UndertowClient;
+import io.undertow.server.DefaultByteBufferPool;
 import io.undertow.server.HttpHandler;
 import io.undertow.server.HttpServerExchange;
-import io.undertow.util.Headers;
+import io.undertow.util.AttachmentKey;
+import io.undertow.util.HttpString;
+import io.undertow.util.Protocols;
+import io.undertow.util.SameThreadExecutor;
 import io.undertow.util.StatusCodes;
+import io.undertow.util.StringReadChannelListener;
 import org.jboss.dmr.ModelNode;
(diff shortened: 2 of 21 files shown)


fractions/io/src/main/java/org/wildfly/swarm/io/IOFraction.java
```diff
@@ -30,7 +30,11 @@
 public class IOFraction extends IO<IOFraction> implements Fraction<IOFraction> {
 
     public IOFraction applyDefaults() {
-        return worker(new Worker("default"))
+        return worker(
+                "default", w -> {
+                    w.ioThreads(100);
+                    w.taskMaxThreads(20);
+                })
                 .bufferPool(new BufferPool("default"));
```
bobmcwhirter, line 38: Not worth editing, but `.bufferPool("default")` would also suffice and delegate to the factory internally.

fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/HttpContexts.java
```diff
@@ -61,70 +99,171 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
...
+            Runnable r = new Runnable() {
+                @Override
+                public void run() {
+
+                    try {
+
+                        final UndertowClient client = UndertowClient.getInstance();
+                        final List<ClientResponse> responses = new ArrayList();
+
+                        ClientConnection connection =
```
bobmcwhirter: Ah, so when a user hits `/health` or whatnot, we internally invoke additional HTTP requests back to ourself to aggregate the results?

fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/HttpContexts.java
```diff
@@ -61,70 +99,171 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
...
+                        for(HealthMetaData healthCheck : monitor.getHealthURIs()) {
+
+                            try {
+                                ClientRequest request = new ClientRequest();
+                                request.setProtocol(Protocols.HTTP_1_1);
+                                request.setPath(healthCheck.getWebContext());
+                                request.getRequestHeaders().add(X_SWARM_HEALTH_TOKEN, EPHEMERAL_TOKEN);
+                                final CountDownLatch latch = new CountDownLatch(1);
+                                connection.sendRequest(request, createClientCallback(responses, latch));
+                                latch.await(monitor.getProbeTimeoutSeconds(), TimeUnit.SECONDS);
```
bobmcwhirter: Should the `latch.await()` follow the loop of async requests, so we can wait in parallel instead of in sequence?

heiko-braun:
test this please

fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/HttpContexts.java
Ladicek: Very much valid concern, IMHO. I don't want to see how this performs when we have 100 custom health checks. Can't we just invoke the methods? (I know we can't. CDI and JAX-RS and whatnot. The question is more like: can we impose some constraints on the custom health checks so that we _could_ just invoke the methods?)

heiko-braun: Besides, even with 100 healthchecks, why do you think this automatically leads to a bottleneck?

Ladicek: Right, 100 health checks is a bit overblown (I got carried away and confused general monitoring metrics and health checks), but I'm fairly sure that noone expects that custom health check = additional network communication.

gastaldi (approved):

heiko-braun:
To please the sceptics on this thread i took a deep dive in undertow nio and replaced the probes with in-vm communication

heiko-braun:
The performance looks decent as well:

```
Concurrency Level:      20
Time taken for tests:   1.257 seconds
Complete requests:      1000
Failed requests:        486
   (Connect: 0, Receive: 0, Length: 486, Exceptions: 0)
Non-2xx responses:      486
Total transferred:      335634 bytes
HTML transferred:       231372 bytes
Requests per second:    795.48 [#/sec] (mean)
Time per request:       25.142 [ms] (mean)
Time per request:       1.257 [ms] (mean, across all concurrent requests)
Transfer rate:          260.73 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    8   5.1      8      41
Processing:     3   17   5.9     16      44
Waiting:        1   12   5.6     11      43
Total:          5   25   7.5     23      61

```

(please note that one check is setup to fail ~50% of the time, hence the failed requests)

PR #235, opened 2016-11-30
SWARM-858: additional fractions: unintuitive behavior if the fraction is not found

Ladicek:

-----

Motivation
----------
If the user sets an additional fraction in the Maven plugin
configuration but makes a mistake, two things can happen:

1. If the user used the format "artifactId", an unhelpful
   error message is shown.
2. If the user used the format "artifactId:version"
   or "groupId:artifactId:version", build passes just fine.

Both of these behaviors are unintuitive.

Modifications
-------------
The method FractionDescriptor.fromGav was changed to always
consult the FractionList. If the fraction isn't found
in the list, an appropriate error message is used.

Result
------
Wrong configuration of additional fractions will now always
result in build errors.

### DIFF
--- a/tools/src/main/java/org/wildfly/swarm/tools/FractionDescriptor.java
+++ b/tools/src/main/java/org/wildfly/swarm/tools/FractionDescriptor.java
@@ -39,6 +39,19 @@ public class FractionDescriptor {
         this(groupId, artifactId, version, null, null, null, false, FractionStability.UNSTABLE);
     }
 
+    /**
+     * Retrieves a {@link FractionDescriptor} from the {@code fractionList} based on the {@code gav} string.
+     *
+     * <p>The {@code gav} string contains colon-delimited Maven artifact coordinates. Supported formats are:</p>
+     *
+     * <ul>
+     *     <li>{@code artifactId}: the groupId {@code org.wildfly.swarm} is presumed</li>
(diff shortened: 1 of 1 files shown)


tools/src/main/java/org/wildfly/swarm/tools/FractionDescriptor.java
```diff
@@ -39,24 +39,50 @@ public FractionDescriptor(String groupId, String artifactId, String version) {
...
+                    throw new RuntimeException("Version mismatch: requested " + gav + ", found " + desc.av());
+                }
                 break;
             case 3:
-                desc = new FractionDescriptor(parts[0], parts[1], parts[2]);
+                desc = fractionList.getFractionDescriptor(parts[0], parts[1]);
+                if (desc == null) {
+                    throw new RuntimeException("Fraction not found: " + gav);
+                }
+                if (!desc.getVersion().equals(parts[1])) {
```
kenfinnigan: Should this be parts[2] ??

Ladicek: Oops! Absolutely.

Ladicek: Fixed, thanks.

kenfinnigan (approved):

Ladicek:
Note to self: I need to check if the old code didn't support some sort of wildcards, especially for the version field. That would be regressed by this commit.

PR #242, opened 2016-12-03
SWARM-879: Maven is missing auth. for mirrors

philschaller:
Motivation
----------
Maven builds with mirrors that need authentication are failing since
release 2016.11.0.

Modifications
-------------
Maven plugin mirror resolution, especially the authentication part.

Result
------
Maven builds with mirrors that need authentication are working again.


kenfinnigan:
ok to test

kenfinnigan (approved):

PR #247, opened 2016-12-07
SWARM-882: Generate keycloak.json from System properties

emag:

-----

Motivation
----------
It'd be useful that generating keycloak.json from System properties on
runtime. We don't have to re-build app when re-writing each property of
keycloak.json.

Modifications
-------------
Add KeycloakJsonGenerator and expose some depdencies(keycloak-core, jackson)

Result
------
You can generate keycloak.json on runtime.

### DIFF
--- a/fractions/keycloak/module.conf
+++ b/fractions/keycloak/module.conf
@@ -1,9 +1,12 @@
 org.wildfly.swarm.undertow
 io.undertow.servlet
 io.undertow.core
+org.keycloak.keycloak-core
 org.keycloak.keycloak-undertow-adapter
 org.keycloak.keycloak-adapter-subsystem
 org.keycloak.keycloak-jboss-adapter-core
 org.keycloak.keycloak-wildfly-adapter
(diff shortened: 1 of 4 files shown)


kenfinnigan:
Is generating the file from system properties the right solution, or should we simply sidecar the keycloak.json file so that it can be external to the uber jar?

The latter has been requested as well: https://issues.jboss.org/browse/SWARM-849

Adding the additional module dependencies to make the generator possible does add some extra weight to the uber jar for a use case that not everyone would use.

emag:
@kenfinnigan Sorry for not checking the exisisting JIRA enough.
Yes, I agree with just loading external kyecloak.json. I think there are no differences between generating from sys props and loading external keycloak.json effectively.
> if I could say a pros of sys props, we can aggregate all parameters in one place such as project-stages.yml.

emag:
Now I sent another p-r (#251) for SWARM-849 , I cancel this p-r.

PR #249, opened 2016-12-08
SWARM-891: Swarm's Arquillian container should support fraction autodetection

Ladicek:

-----

Motivation
----------
The Arquillian container invocation of `BuildTool` hardcodes
`BuildTool.FractionDetectionMode.never`. This means that
Swarm's Arquillian support can't be used in projects
that rely on fraction autodetection.

Modifications
-------------
Added a configuration option `fractionDetectMode`
to the Swarm Arquillian container.

Result
------
It is now possible to use fraction autodetection with Arquillian.


Ladicek:
I don't understand why the CI job could possibly have failed. I only changed stuff in `arquillian/adapter`, which didn't even get built.

kenfinnigan:
The reason it failed is because previously fraction-list wasn't built by Maven until just before the testsuites, and after every other fraction, as it needs all the fractions to be built first.

This change adds a dependency to fraction-list on arquillian test, causing fraction-list to be built before every fraction which means those fractions don't exist in the local M2 REPO yet

Ladicek:
~~Aha, OK, I see where the problem (probably) is. I added a dependency on `fraction-list` to `arquillian/adapter`, which means that the `fraction-list` module is now built much sooner than it used to. Hmm...~~

Thanks Ken. I'll have to think about how can I solve this.

Ladicek:
I've slightly reordered the modules in top-level POM, to move `fraction-list`, `arquillian` etc. _after_ all the fractions. It's not like they will actually get built after all of them, because apparently `fractions/camel/core/tests` already depends on `arquillian`, but it should be enough to fix the tests. And I'd guess that all tests that need `org.wildfly.swarm:arquillian` should be inside `testsuite`?

kenfinnigan:
in theory anything using arquillian should be in testsuite, so it may be worth fixing camel/core/tests

Ladicek:
I'll try to move the module, doesn't look too dificult.

kenfinnigan:
Wouldn't necessarily move the module, but move the tests to testsuite as there should be camel tests there already

Ladicek:
I'm thinking about only moving the `fractions/camel/core/tests` module. That should be inside `testsuite` IMHO. And while there's a bunch of `testsuite-camel-*` modules in `testsuite` already, there's no `testsuite-camel-core`. In any case, I'll have a look and see what's sensible.

Ladicek:
Not again! :-)

Ladicek:
I'm thinking if I possibly didn't introduce a runtime circular dependency. Looking at the build log https://projectodd.ci.cloudbees.com/job/wildfly-swarm-core-pr-validation/482/consoleFull the Fraction List module build needs to download `maven-metadata.xml` of the Arquillian modules. Which obviously come from previous builds, because in this build, the Arquillian modules weren't built yet. So maybe Arquillian shouldn't really depend on Fraction List, because Fraction List needs Arquillian to be built sooner. I don't know.

kenfinnigan:
Right, arquillian is part of the fractions that are added to fraction-list so will need to be built before fraction-list runs.

I'm thinking that it's not possible for arquillian to depend on fraction-list.

Maybe a more appropriate solution is for package to fraction mapping to be moved into BuildTool?

Ladicek:
Thanks Ken. I'll take a look.

kenfinnigan:
@Ladicek so what's the status of this PR? Does it need a "hold" for now?

Ladicek:
So I tried to merge `fraction-list` into `tools`, but the problem remains: `arquillian` is fraction(-ish), so it can't itself depend on anything that requires the functionality of `fraction-list` (no matter where it lives). I have no idea how I would break that dependency cycle, so I'm gonna close this PR and unassign myself from the JIRA.

But if you have any miraculous idea how the problem could be solved, I'm all ears :-)

PR #251, opened 2016-12-08
SWARM-849: Read external keycloak.json

emag:

-----

Motivation
----------
It'd be helpful reading external keycloak.json
because you don't have to re-build -swarm app
when re-writing keycloak.json.

Modifications
-------------
Add a process to read external keycloak.json
pointed from system property(swarm.keycloak.json.path).
If the system prop doesn't exist, search it from classpath as
usual.

Result
------
-swarm apps can read external keycloak.json.


fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/internal/SecuredImpl.java
```diff
@@ -65,28 +67,12 @@ public SecuredImpl(ArchiveBase<?> archive) {
...
-                    if (jsonNode != null && jsonNode.getAsset() != null) {
-                        keycloakJson = jsonNode.getAsset().openStream();
-                    }
-                } catch (IOException e) {
-                    // ignore
-                    // e.printStackTrace();
-                }
-            }
+        InputStream keycloakJson;
+        String keycloakJsonPath = System.getProperty("swarm.keycloak.json.path");
```
gastaldi: It would be nice to use @Configurable here. The information might come from `project-stages.yml` too

kenfinnigan:
Maybe instead of implementing it within SecuredImpl, as we're looking to move away from main() usage, it should be an ArchivePreparer instead?

Then it would be possible to @Inject Configurable into the ArchivePreparer?

PR #258, opened 2016-12-13
SWARM-916: default security subsystem configuration different from WildFly

Ladicek:

-----

Motivation
----------
The default configuration of the `security` subsystem as created
by `SecurityFraction.applyDefaults` only contains a subset
of WildFly's default. Most of WildFly's default configuration
also applies to Swarm, perhaps with the exception of the `Remoting`
login module.

Modifications
-------------
Modified `SecurityFraction.applyDefaults` to match the default
configuration of WildFly, with the exception of the `Remoting`
login module.

Result
------
Default configuration is closer to WildFly.

### DIFF
--- a/fractions/wildfly/security/src/main/java/org/wildfly/swarm/security/SecurityFraction.java
+++ b/fractions/wildfly/security/src/main/java/org/wildfly/swarm/security/SecurityFraction.java
@@ -23,7 +23,12 @@ import org.wildfly.swarm.config.Security;
 import org.wildfly.swarm.config.security.Flag;
 import org.wildfly.swarm.config.security.SecurityDomain;
 import org.wildfly.swarm.config.security.security_domain.ClassicAuthentication;
+import org.wildfly.swarm.config.security.security_domain.ClassicAuthorization;
+import org.wildfly.swarm.config.security.security_domain.JaspiAuthentication;
+import org.wildfly.swarm.config.security.security_domain.authentication.AuthModule;
 import org.wildfly.swarm.config.security.security_domain.authentication.LoginModule;
+import org.wildfly.swarm.config.security.security_domain.authentication.LoginModuleStack;
+import org.wildfly.swarm.config.security.security_domain.authorization.PolicyModule;
 import org.wildfly.swarm.spi.api.Fraction;
 import org.wildfly.swarm.spi.api.annotations.MarshalDMR;
 import org.wildfly.swarm.spi.api.annotations.WildFlyExtension;
@@ -45,7 +50,8 @@ public class SecurityFraction extends Security<SecurityFraction> implements Frac
     }
 
     public SecurityFraction applyDefaults() {
-        return securityDomain(new SecurityDomain("other")
+        securityDomain(new SecurityDomain("other")
+                .cacheType(SecurityDomain.CacheType.DEFAULT)
                 .classicAuthentication(new ClassicAuthentication()
                         .loginModule(new LoginModule("RealmDirect")
                                 .code("RealmDirect")
@@ -55,6 +61,23 @@ public class SecurityFraction extends Security<SecurityFraction> implements Frac
                                 }})
 
                         )));
+
+        securityDomain(new SecurityDomain("jaspitest")
+                .cacheType(SecurityDomain.CacheType.DEFAULT)
+                .jaspiAuthentication(new JaspiAuthentication()
+                        .loginModuleStack(new LoginModuleStack("dummy")
+                                .loginModule(new LoginModule("Dummy")
+                                        .code("Dummy")
+                                        .flag(Flag.OPTIONAL)
+                                )
+                        )
+                        .authModule(new AuthModule("Dummy")
+                                .code("Dummy")
+                        )
+                )
+        );
+
+        return this;
     }
 
 }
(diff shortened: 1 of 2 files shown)


fractions/wildfly/security/src/main/java/org/wildfly/swarm/security/SecurityFraction.java
```diff
@@ -55,6 +61,42 @@ public SecurityFraction applyDefaults() {
                                 }})
 
                         )));
+
+        securityDomain(new SecurityDomain("jboss-web-policy")
```
kenfinnigan: I think this should be a customizer in undertow, as to me it doesn't make sense to have a `jboss-web-policy` in place without undertow

fractions/wildfly/security/src/main/java/org/wildfly/swarm/security/SecurityFraction.java
```diff
@@ -55,6 +61,42 @@ public SecurityFraction applyDefaults() {
...
+                .cacheType(SecurityDomain.CacheType.DEFAULT)
+                .classicAuthorization(new ClassicAuthorization()
+                        .policyModule(new PolicyModule("Delegating")
+                                .code("Delegating")
+                                .flag(Flag.REQUIRED))
+                )
+        );
+
+        securityDomain(new SecurityDomain("jboss-ejb-policy")
```
kenfinnigan: Similar to above, I would only expect this to be created if `ejb` is present

Ladicek:
Hmm, right, that makes sense.

Ladicek:
Fixed, I believe. But I can't see those customizers being applied, so please __hold__ this, I need to check better.

bobmcwhirter:
Each `Customizer` must have one of `@Pre` or `@Post` annotation on the class to denote when it should run.

Without either, it never will fire.

Ladicek:
Thanks Bob, but the problem was elsewhere. The `Customizer`s live in the `security` fraction and they want to inject `EJBFraction` / `UndertowFraction`. While the `security` fraction depends on the `ejb` / `undertow` fraction at compile-time, it didn't depend on them at JBoss Modules runtime. So those classes weren't visible. I had to add the fraction modules to the `security` fraction's `module.conf` and it works fine now.

Ladicek:
OK, looks like removing the `security` fraction dependency from the `ejb` fraction wasn't the best idea.

fractions/wildfly/security/module.conf
```diff
@@ -1,2 +1,4 @@
 org.jboss.as.security
 org.jboss.security.negotiation
+org.wildfly.swarm.ejb
```
kenfinnigan: i think both of these need to be optional otherwise they will always be added to uber jar when `security` is included

kenfinnigan:
Hmm, not sure how the ejb/undertow dependencies on undertow should be, maybe provided?

Ladicek:
OK, let's pretend we didn't see the circular dependency thing I created in rush.

I see two options:

- The security `Customizer`s will live in the `security` fraction. This is technically a breaking change, as shown by https://projectodd.ci.cloudbees.com/job/wildfly-swarm-core-pr-validation/498/ It would require users to depend on the `security` fraction explicitly. I don't think we want that, because we mark the`security` fraction as _internal_.
- The security `Customizer`s will live in the `ejb` / `undertow` fractions.

@kenfinnigan WDYT?

kenfinnigan:
Shoot, forgot the security fraction was internal.

Yes, though it would be "nice", for some level of "nice", for all the security to be captured in one fraction, it makes sense to not require developers to explicitly add security unless they really need to do something custom.

Given all the above blurb, customizers in ejb/undertow is the way to go it seems

fractions/javaee/undertow/module.conf
```diff
@@ -3,3 +3,4 @@ org.wildfly.extension.undertow
 io.undertow.core
 io.undertow.servlet
 org.ow2.asm
+org.wildfly.swarm.security
```
kenfinnigan: is that needed? maven dep should be sufficient.

Ladicek: I'll try again, but my understanding is that without the module dependency, the `SecurityFraction` class won't be visible to the customizer.

bobmcwhirter: So, we have a trick, sorta, to avoid having to have mvn dependencies, but being able to have pseudo fraction optional dependencies.

if you inject

    Instance<Security> security

you can reference the superclass of the `SecurityFraction` and inject by it, without having to have enough Maven dependency to see `SecurityFraction`.  This works because the parent class of `SecurityFraction` is `Security` (I think) and exists in the config-api which is available across the board for all fractions.

So, while we actually will inject (if present) a `SecurityFraction`, you only see it by the one level up superclass of `Security` from config-api.

kenfinnigan:
forgot about that, good call

fractions/javaee/undertow/module.conf
bobmcwhirter: Example of this cross-fraction optionality:

https://github.com/wildfly-swarm/wildfly-swarm/blob/master/fractions/wildfly/infinispan/src/main/java/org/wildfly/swarm/infinispan/runtime/InfinispanCustomizer.java#L45-L55

Notice those injections are bare config-api classes, so require no mvn (or module) dependency across them.

Ladicek: OK, that makes sense, but the `module.conf` addition would still have to be there, correct?

kenfinnigan:
that's the beauty of what @bobmcwhirter proposed, it doesn't

PR #273, opened 2016-12-21
SWARM-917 - Be less eager about generating JAXRS Applications

bobmcwhirter:
Motivation
----------

In the event a user paints with a wide brush and includes
the entire JavaEE API spec jar, auto-detection will mix in
JAX-RS.  While this would still be sub-par, we should protect
against then assuming that the user has actually created a
JAX-RS application.

Modifications
-------------

We only cast/coerce an Archive to a JAXRSArchive (and thus
optionally creating the @ApplicationPath annotated generated
Application class) if there is indeed usage of JAXRS annotations
within the application (or its dependencies).

Result
------

Just because the JAXRS fraction is detected doesn't necessarily
mean we actually truly need it, and now we look deeper into our
souls before committing ourselves to presuming JAXRS is actually
involved.

-----


kenfinnigan (approved):

kenfinnigan:
retest this please

bobmcwhirter:
Weird. Will investigate

PR #276, opened 2016-12-23
SWARM-961: Place temporary files under build directory while running tests

gastaldi:

-----

Motivation
----------
WildFly Swarm creates a lot of temporary dirs/files and in some cases they are not deleted, causing the CI agent to crash with Insufficient Disk space errors after some time.

Modifications
-------------
java.io.tmpdir is set to the ${project.build.directory}.

Result
------
Temporary files created during a test execution are placed under target/ and deleted when the clean command is issued, leaving temporary dirs contained within the project

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,6 +130,7 @@
           <runOrder>alphabetical</runOrder>
           <failIfNoTests>false</failIfNoTests>
           <systemPropertyVariables>
+            <java.io.tmpdir>${project.build.directory}</java.io.tmpdir>
             <swarm.bind.address>127.0.0.1</swarm.bind.address>
             <project.version>${project.version}</project.version>
           </systemPropertyVariables>
@@ -164,6 +165,9 @@
             </goals>
             <configuration>
               <testSourceDirectory>src/it/java</testSourceDirectory>
+              <systemPropertyVariables>
+                <java.io.tmpdir>${project.build.directory}</java.io.tmpdir>
+              </systemPropertyVariables>
             </configuration>
           </execution>
         </executions>
--- a/tools/src/main/java/org/wildfly/swarm/tools/exec/SwarmExecutor.java
+++ b/tools/src/main/java/org/wildfly/swarm/tools/exec/SwarmExecutor.java
@@ -75,7 +75,8 @@ public class SwarmExecutor {
         Set<String> names = System.getProperties().stringPropertyNames();
 
         for (String name : names) {
-            if (name.startsWith("jboss") || name.startsWith("swarm") || name.startsWith("wildfly") || name.startsWith("maven")) {
+            if (name.startsWith("jboss") || name.startsWith("swarm") || name.startsWith("wildfly") || name.startsWith("maven")
+                    || name.equals("java.io.tmpdir")) {
                 String value = System.getProperty(name);
                 this.properties.put(name, value);
             }


gastaldi:
I still see 2 `wildfly-self-contained*.d` directories under /tmp after the build (could not find where that is being created) but this PR succeeds in placing the temporary files inside target/

gastaldi:
Found the reason why these directories are being created. Will push the fix to this PR soon

bobmcwhirter:
Can we not just pass this sys prop via the job?

gastaldi:
Probably, but do you see any issues with this strategy? At least your local
/tmp will remain clean ;)

Em 22 de dez de 2016 23:47, "Bob McWhirter" <notifications@github.com>
escreveu:

gastaldi:
Ah no, sorry that's impossible because UberJarContainer forks the process
and ignores my java.io.tmpdir override

Em 23 de dez de 2016 00:06, "George Gastaldi" <ggastald@redhat.com>
escreveu:

kenfinnigan:
Could we not do a .startsWith("java") in that check, rather than the full property?

Then it would be passed through right?

I'd also prefer to not make all builds run this way

bobmcwhirter:
Unfortunately that'd allow java.class.path and others.

I'd be okay whitelisting exactly java.io.tmpdir.

kenfinnigan:
ah ok, but do we want it part of the regular build?

bobmcwhirter:
I meant whitelist it in general and then add to the CI job.

gastaldi:
Sorry, but what do you suggest to add to the CI job? 

@kenfinnigan why don't you prefer to run builds this way but is ok having tons of temp files in your /tmp that never gets deleted?

FYI this setting has existed for some time in the jboss-parent-pom: https://github.com/jboss/jboss-parent-pom/blob/master/pom.xml#L592

bobmcwhirter:
So, I think prior to this (or possibly obviating it) would be to ensure that -Djava.io.tmpdir set on a root mvn gets communicated to any forked uberjar thingy.

When I say "whitelisted" I mean copied, since we currently support -Dswarm.* properties. 

The one concern is that we tend to take these properties and jam them into swarm.properties within the uberjar, which we definitely don't want to do with java.io.tmpdir.

But, assuming we can communicate, safely, from the root mvn invocation, I'd rather not edit our POMs to explicitly say where to put TMP.

And speaking of which, maybe $TMP or the windows equiv is sufficient to set on CI.  java.io.tempdir should respect that if not otherwise set.

The problem with manually managing java.io.tmpdir is that the OS does a much better job of house-keeping than a user, along with concerns of disk allocation, etc.  I'd personally hate someone leaving tmp in my dir tree.

bobmcwhirter:
So, I'd say given current, unmodified HEAD, let's try setting TMP=$WORKSPACE/tmp or whatnot and see if that suffices to accomplish what we want.

bobmcwhirter:
Meanwhile, I'll try to figure out why we still have crap lingering.

bobmcwhirter:
And leave an un-ending stream of comments on this PR.

bobmcwhirter:
Just because it's fun.

bobmcwhirter:
I failed to get setting TMP to work.
But..

    export _JAVA_OPTIONS=-Djava.io.tempdir=/path/to/whatever

Seems to work and percolate down through to sub-processes.

That being said, even on OSX I'm seeing lingering files, so working towards that.

gastaldi:
That seems like an ugly hack :)

gastaldi:
This may be worth a shot: https://wiki.jenkins-ci.org/display/JENKINS/Tmp+Cleaner+Plugin

bobmcwhirter:
So, I've got a fix for this underlying issue, at least as exhibited on OSX.  Will boot into Windows and ensure I didn't break anything there.

Basically, our SelfContainerContainer (in wildfly-core repo) has `stop()`, which places a task on the executor to clean up the `wildfly-selfcontained.d/`, but then nothing actually waits for the executor to shutdown cleanly, so we end up blowing it out the water before it's accomplished its goal of removing the `-selfcontained.d/`

bobmcwhirter:
@gastaldi wrt an ugly hack... I still think tmpdir management is a local environment thing and instituting a policy via pom.xml seems strange to me. 

But hopefully with my soon-to-be-inbound PR, the root cause will be less of an issue.  

Once I boot into Windows. *sigh*

gastaldi:
Another solution: http://stackoverflow.com/questions/15675783/how-to-change-the-jenkins-tmp-dir-location

gastaldi:
@bobmcwhirter I agree with you, but at least you can know what files/dirs are created and left in a module level

bobmcwhirter:
Okay, while I've created this PR:

https://github.com/wildfly-swarm/wildfly-swarm/pull/278

It doesn't actually solve the problem, because, as noted in the commit, our Arquillian does a process.destroy(), which on OSX is a SIGINT, but on Windows is basically a SIGKILL, which prevents any shutdown hooks from running.  So we never actually get a chance to clean up after ourselves.

A better solution, but not today, would be to communicate with our Arq Daemon to trigger a clean shutdown from within the running container, before we attempt to just `process.destroy()` it.

So, that being said, we still need some solution on CI to clean up the lingering TMP dir on a regular basis.  Either putting it into a `$WORKSPACE` that we routinely blow away, or otherwise scrub out the dir.  

I leave that choice up to the person who deals with CI the most.  @gastaldi

bobmcwhirter (approved):

PR #288, opened 2017-01-06
SWARM-982: Honor the existing root-logger level and handlers if enabled Logstash Fraction

emag:

-----

Motivation
----------
The existing root-logger level and handlers are overridden by
LogstashFraction level and a single logstash-handler.

Modifications
-------------
* Set LogstashFraction level as logstash-handler one
* Add logstash-handler to the existing root-logger handlers.

Result
------
It never overrides the existing root-logger level and handlers.

### DIFF
--- a/fractions/logstash/src/main/java/org/wildfly/swarm/logstash/runtime/LogstashCustomizer.java
+++ b/fractions/logstash/src/main/java/org/wildfly/swarm/logstash/runtime/LogstashCustomizer.java
@@ -52,13 +52,16 @@ public class LogstashCustomizer implements Customizer {
                     .module("org.jboss.logmanager.ext")
                     .attributeClass("org.jboss.logmanager.ext.handlers.SocketHandler")
                     .namedFormatter("logstash")
-                    .properties(handlerProps);
+                    .properties(handlerProps)
+                    .level(this.logstash.level());
 
             this.logging
                     .customFormatter("logstash", "org.jboss.logmanager.ext", "org.jboss.logmanager.ext.formatters.LogstashFormatter",
                                      this.logstash.formatterProperties())
                     .customHandler(logstashHandler)
-                    .rootLogger(this.logstash.level(), logstashHandler.getKey());
+                    .subresources()
+                    .rootLogger()
+                    .handler(logstashHandler.getKey());
         }
     }
 }
--- a/fractions/logstash/src/test/java/org/wildfly/swarm/logstash/runtime/LogstashCustomizerTest.java
+++ b/fractions/logstash/src/test/java/org/wildfly/swarm/logstash/runtime/LogstashCustomizerTest.java
@@ -6,6 +6,7 @@ import org.junit.Before;
 import org.junit.Test;
 import org.wildfly.swarm.config.logging.CustomFormatter;
 import org.wildfly.swarm.config.logging.CustomHandler;
+import org.wildfly.swarm.config.logging.Level;
 import org.wildfly.swarm.logging.LoggingFraction;
 import org.wildfly.swarm.logstash.LogstashFraction;
 
@@ -23,7 +24,7 @@ public class LogstashCustomizerTest {
     @Before
     public void setUp() {
         this.customizer = new LogstashCustomizer();
-        this.customizer.logging = new LoggingFraction();
+        this.customizer.logging = new LoggingFraction().rootLogger(Level.INFO, "HANDLER");
         this.customizer.logstash = new LogstashFraction();
     }
 
@@ -37,7 +38,6 @@ public class LogstashCustomizerTest {
 
     @Test
     public void testExplicitlyEnabledWithDefaults() {
-
         this.customizer.logstash.enabled(true);
         this.customizer.customize();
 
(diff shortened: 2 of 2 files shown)


kenfinnigan:
retest this please

kenfinnigan:
retest this please

fractions/logstash/src/main/java/org/wildfly/swarm/logstash/runtime/LogstashCustomizer.java
```diff
@@ -54,11 +57,19 @@ public void customize() {
...
+                rootLoggerHandlers.addAll(existingRootLogger.handlers());
+            }
+            rootLoggerHandlers.add(logstashHandler.getKey());
+
             this.logging
                     .customFormatter("logstash", "org.jboss.logmanager.ext", "org.jboss.logmanager.ext.formatters.LogstashFormatter",
                                      this.logstash.formatterProperties())
                     .customHandler(logstashHandler)
-                    .rootLogger(this.logstash.level(), logstashHandler.getKey());
+                    .rootLogger(this.logstash.level(), rootLoggerHandlers.toArray(new String[rootLoggerHandlers.size()]));
```
kenfinnigan: Wondering whether it makes sense for LogStash to have its own level and root logger?

Or instead if we do something like `this.logging.rootLogger().handler(logstashHandler.getKey())`.

Just got me thinking about the relationship between the two logging fractions.

emag: Ah, I should have mentioned Level as well.

Anyway, regarding whether Logstash should have its own level and root logger, I think Level is yes, but root logger is no.

## Level
Handler can have its own level, so we should set `swarm.logstash.level` in the definition of logstashHandler like:

``` java
final CustomHandler<?> logstashHandler = new CustomHandler<>("logstash-handler")
    [...]
    .level(logstash.level()); // added
```

## root logger

At least about the level, `swarm.logging.root-logger.level`  or LoggingFraction root-logger level (set by applyDefaults) should be honored rather than Logstash Fraction Level.

``` java
this.logging
   [...]
   .rootLogger(existingRootLogger.level(), rootLoggerHandlers.toArray(new String[rootLoggerHandlers.size()]));
```

kenfinnigan: Would you be ok making these changes?

To use the logstash level in the handler only, and then maybe using code from my comment to add handler key to root logger?

emag: AS you said on IRC, I was trying to simplify this p-r  like this:

``` java
this.logging
   [...]
  .subresources()
  .rootLogger()
  .handler(logstashHandler.getKey());
```

but, org.wildfly.swarm.config.logging.RootLogger.handlers could be `java.util.Arrays$ArrayList` which cannot be added element(thrown UnsupportedOperationException).

``` java
/**
 * The handlers associated with the root logger.
 */
@SuppressWarnings("unchecked")
public T handler(String value) {
  if (this.handlers == null) {
    this.handlers = new java.util.ArrayList<>();
  }
  this.handlers.add(value);
  return (T) this;
}

/**
 * The handlers associated with the root logger.
 */
@SuppressWarnings("unchecked")
public T handlers(String... args) {
  handlers(Arrays.asList(args));
  return (T) this;
}
```
> If that helps, when specified `swarm.logging.root-logger.handlers` with yml, handlers will be overwritten with java.util.ArrayList.

It seems that we need to re-create it to set up root-logger with LoggingFraction.rootLoggerLevel level, String... handlers) like my above comment for now.

Should it be fixed in config-api?
> I guess this point is here: 
> https://github.com/wildfly-swarm/wildfly-config-api/blob/master/generator/src/main/java/org/wildfly/swarm/config/generator/generator/ResourceFactory.java#L293-L300

bobmcwhirter:
I may not be completely following, but for config-api, typically:

* `foos(listOfFoos)` is meant to completely replace the entire set of `foos`
* `foo(singleFoo)` is meant to add a `foo` to the set

bobmcwhirter:
Ah, so `foos(foos)` effectively locks it with an unmodifiable collection which prevents subsequent singular `foo(someFoo)` addition.  That should be fixed, yes.

kenfinnigan:
so we need a config-api fix to make this work?

emag:
Thank you, @bobmcwhirter ! I'm not sure config-api should be fixed really.

instead, how about changing `LoggingFraction#rootLogger(Level level, String... handlers)`  ?

```
public LoggingFraction rootLogger(Level level, String... handlers) { // plan A: change parameter `handlers` from varargs to List
    rootLogger(new RootLogger().level(level)
                       .handlers(handlers)); // plan B, change variables `handlers` from array to Modifiable List
    return this;
}
```

https://github.com/wildfly-swarm/wildfly-swarm/blob/master/fractions/wildfly/logging/src/main/java/org/wildfly/swarm/logging/LoggingFraction.java#L330

emag:
@bobmcwhirter @kenfinnigan I rethink about it, the config-api should be fixed.my above proposal is half measures.
I'll open an issue of about it and send p-r later.

emag:
Now my p-r is included the root-logger level issue. I've edited the contents of the commit. Please review again. Thanks!

kenfinnigan (approved):

PR #292, opened 2017-01-09
Classifier may be null, improved checking to avoid NPE.

nelsongraca:
Made what I believe to be a small improvement, hope it helps.

### DIFF
--- a/plugin/src/main/java/org/wildfly/swarm/plugin/maven/MultiStartMojo.java
+++ b/plugin/src/main/java/org/wildfly/swarm/plugin/maven/MultiStartMojo.java
@@ -188,7 +188,7 @@ public class MultiStartMojo extends AbstractSwarmMojo {
                 .stream()
                 .filter(e -> (e.getGroupId().equals(groupId) && e.getArtifactId().equals(artifactId) && ((e.getClassifier() == null && classifier == null) || (e.getClassifier() != null && e.getClassifier().equals(classifier)))))
                 .findFirst()
-                .orElseGet(null);
+                .orElse(null);
     }
 
     protected Xpp3Dom getGlobalConfig() {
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/wildfly/swarm/plugin/maven/MultiStartMojo.java
```diff
@@ -186,7 +186,10 @@ protected MavenProject findProject(String groupId, String artifactId) {
     protected Artifact findArtifact(String groupId, String artifactId, String classifier) {
         return this.project.getArtifacts()
                 .stream()
-                .filter(e -> (e.getGroupId().equals(groupId) && e.getArtifactId().equals(artifactId) && e.getClassifier().equals(classifier)))
+                .filter(e -> (e.getGroupId().equals(groupId) && e.getArtifactId().equals(artifactId)
+                        && ((e.getClassifier() != null e.getClassifier().equals(classifier)) || e.getClassifier() == null && classifier == null)
```
gastaldi: This doesn't seem to compile

nelsongraca: Will double-check.

nelsongraca: Seems to be checkstyle, can you help with that?

nelsongraca:
Now it compiles.

nelsongraca:
Another fix for a possible NPE.

kenfinnigan:
ok to test

kenfinnigan:
Thanks for this fix!

In the future, would be appreciated if you could also create an issue in [JIRA](https://issues.jboss.org/browse/SWARM)

nelsongraca:
Will do in the future.

PR #303, opened 2017-01-17
SWARM-966: Resource root in module.xml not found running in IDE

kenfinnigan:

-----

Motivation
----------
When providing a custom module.xml, the jar within a <resource-root> element was not being found such that classes could be loaded from it.

Modifications
-------------
Create a ResourceLoader for Jars when we're dealing with a .jar or .war resource.

Result
------
<resource-root> jars are correctly loaded in a custom module.xml

### DIFF
--- a/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/NestedJarResourceLoader.java
+++ b/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/NestedJarResourceLoader.java
@@ -76,6 +76,11 @@ public class NestedJarResourceLoader {
                 }
             }
         } else if (urlString.startsWith("file:")) {
+            if (loaderName.endsWith(".jar") || loaderName.endsWith(".war")) {
+                return ResourceLoaders.createJarResourceLoader(
+                        loaderName,
+                        new JarFile(new File(new File(urlString.substring(5, urlString.length())), loaderPath)));
+            }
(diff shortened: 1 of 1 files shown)


core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/NestedJarResourceLoader.java
```diff
@@ -76,6 +76,11 @@ public static ResourceLoader loaderFor(URL base, String rootPath, String loaderP
                 }
             }
         } else if (urlString.startsWith("file:")) {
+            if (loaderName.endsWith(".jar") || loaderName.endsWith(".war")) {
+                return ResourceLoaders.createJarResourceLoader(
+                        loaderName,
+                        new JarFile(new File(new File(urlString.substring(5, urlString.length())), loaderPath)));
```
gastaldi, line 82: Not a blocker, but the nested `File` object is not needed, as you can use the `File(String, String)` constructor

gastaldi (approved):

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/NestedJarResourceLoader.java
gastaldi: also, `urlString.substring(5)` is cleaner than `urlString.substring(5,urlString.length())` 👍

PR #304, opened 2017-01-18
SWARM-958 - Improve hierarchy of configuration items.

bobmcwhirter:
Motivation
----------

We lose the hiearchy of swarm.logging.com.foo.bar.level=TRACE
when we flatten everything down to properties, even when it was
originally represented in the YAML.

This makes it impossible to adjust logging levels through the YAML
or properties.

Modifications
-------------

Radically adjust how we handle all configuration inputs (Yaml and Properties)
to build a tree (of ConfigNodes) where each node can have a value and/or
children.

Provide a resolution strategy that can load properties or several YAMLs
and provide override capabilities.

BREAKING CHANGE: project-stages.yml is deprecated and removed from the CLI,
but if found it is still supported.  For now.

Property representation of configuration item names that include dots
(such as logger category names) are now bracketed with square brackets:

e.g.

  -Dswarm.logging.loggers.[com.foo.bar].level=TRACE

Regardless of the source of the configuration value, appropriate properties
will be set for each item, in the event someone wants to not @Inject or
otherwise deal with our fanciness, and only wish to use System.getProperty(...).

Additionally, by using -PconfigName, possibly many times, you can activate,
in-order, project-${configName}.yml as sources of inputs.  This is the
replacement to the monolithic project-stages.yml file.

These new YAML files may be found relative to $PWD or from within your
application's classpath. (not in META-INF, just at the root).

Result
------

You can now configure logging via yaml. Basically.

-----


testsuite/testsuite-container/src/test/java/org/wildfly/swarm/container/test/ProjectStagesTest.java
```diff
@@ -15,26 +15,13 @@
...
-import java.util.List;
-import java.util.Map;
-
-import static org.junit.Assert.*;
-
 /**
  * @author Heiko Braun
  * @since 07/04/16
  */
 public class ProjectStagesTest {
```
gastaldi: Remove this class

core/container/src/main/java/org/wildfly/swarm/container/runtime/cdi/ConfigViewProductingExtension.java
```diff
@@ -15,52 +15,39 @@
...
 /**
- * Produces an explicitly set project-stage.
+ * Produces an explicitly set {@link ConfigView}
  *
  * @author Bob McWhirter
  */
-public class ProjectStageProducingExtension implements Extension {
-
-    private final ProjectStage projectStage;
+public class ConfigViewProductingExtension implements Extension {
```
kenfinnigan: Was this meant to be ConfigViewProducingExtension? Or was the 't' intentional?

kenfinnigan (approved):

gastaldi (approved):

core/container/src/main/java/org/wildfly/swarm/cli/CommandLine.java
```diff
@@ -106,33 +106,24 @@
...
+            .withDescription("URL to configuration YAML to use")
             .then((cmd, opt, value) -> cmd.put(opt, Option.toURL(value)));
 
-
-    public static final Option<String> ACTIVE_STAGE = new Option<String>()
-            .withShort('S')
-            .withLong("stage")
-            .hasValue("<active-stage>")
+    public static final Option<String> PROFILES = new Option<String>()
+            .withShort('P')
```
emag, line 122: `-P` is already used for properties file.

core/container/src/main/java/org/wildfly/swarm/cli/CommandLine.java
```diff
@@ -155,8 +146,7 @@ public static Options defaultOptions() {
                 PROPERTY,
                 PROPERTIES_URL,
                 SERVER_CONFIG,
-                STAGE_CONFIG,
-                ACTIVE_STAGE,
+                CONFIG,
```
emag, line 149: We need to add `PROFILES` ?

bobmcwhirter:
Thanks! Great catch! I'll fix Monday.

PR #308, opened 2017-01-20
SWARM-672 - Multiple sources of configuration

bobmcwhirter:
Motivation
----------

We may need to load configuration from various places.  While
we don't currently, we should have an architecture in place to
support it when/if required.  Plus we *do* currently load several
types of configuration (.yml and .properties) from several places
(filesystem, system classpath, swarm.application module).  So
we should clean up that code.

Modifications
-------------

Internal architectural changes to support strategies for locating
and loading .yml or .properties from various places.

Result
------

Softer skin.

-----


kenfinnigan (approved):
Couple of System.err's in there that should be log statements, but not a major issue

PR #311, opened 2017-01-24
SWARM-839: Do not override user configured context root

harrol:
(Possible) fix for https://issues.jboss.org/browse/SWARM-839

The problem seems to be that SwaggerArchivePreparer always overrides the user configured context root (Swagger base path) with the context root from the deployment. 
This prevents the user from configuring their own basePath using SwaggerArchive.setContextRoot()

The example application at https://github.com/harrol/wildfly-swarm-swagger-test can be used to reproduce the bug in Wildfly Swarm 2017.1.1. The basePath is not set to '/context' as expected. When using the proposed fix the basePath is set.


kenfinnigan:
ok to test

bobmcwhirter (approved):

kenfinnigan:
Thanks @harrol for your contribution!

PR #314, opened 2017-01-25
SWARM-891: Fraction autodetection with Arquillian tests

kenfinnigan:

-----

Motivation
----------
Currently there is no way for Arquillian tests to autodetect fractions when none are specified as Maven dependencies.

Modifications
-------------
Add fraction-list to arquillian and set autodetection to `when_missing` on BuildTool.

Modify Shrinkwrap resolving to add dependencies directly as opposed to passing a list of gavs into resolve() to ensure that the scope of each dependency is set to its actual value, as Shrinkwrap defaults to compile.

Modify the ordering of explicit deps from `DependencyTree` so that compile and provided dependencies come first. This ensures that when iterating to retrieve transitives for each of them, we're retrieving transitive compile dependencies first. Otherwise a test scoped dependency could take precedence over a compile one for the same gav.

Result
------
Arquillian tests are now able to utilize autodetection of fractions.

### DIFF
--- a/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
+++ b/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
@@ -51,6 +51,7 @@ import org.wildfly.swarm.arquillian.CreateSwarm;
 import org.wildfly.swarm.arquillian.adapter.resources.ContextRoot;
 import org.wildfly.swarm.arquillian.resolver.ShrinkwrapArtifactResolvingHelper;
 import org.wildfly.swarm.bootstrap.util.BootstrapProperties;
+import org.wildfly.swarm.fractionlist.FractionList;
 import org.wildfly.swarm.internal.FileSystemLayout;
 import org.wildfly.swarm.spi.api.DependenciesContainer;
 import org.wildfly.swarm.spi.api.JARArchive;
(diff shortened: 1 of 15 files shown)


bobmcwhirter (approved):

arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
```diff
@@ -152,7 +153,8 @@ public void start(Archive<?> archive) throws Exception {
         final ShrinkwrapArtifactResolvingHelper resolvingHelper = ShrinkwrapArtifactResolvingHelper.defaultInstance();
 
         BuildTool tool = new BuildTool(resolvingHelper)
-                .fractionDetectionMode(BuildTool.FractionDetectionMode.never)
+                .fractionDetectionMode(BuildTool.FractionDetectionMode.when_missing)
```
Ladicek, line 156: I'd really like to have this configurable.

kenfinnigan: Where and why?

Ladicek: For running Java EE 7 Samples with Swarm, I sometimes had to add some fractions manually, because they couldn't be autodetected (see SWARM-974), but I only wanted to do minimal changes, so I used fraction autodetection mode of `force`.

It can easily be made configurable in `arquillian.xml`, see https://github.com/Ladicek/wildfly-swarm/commits/arquillian-fraction-autodetection and https://github.com/Ladicek/javaee7-samples/tree/wildfly-swarm

PR #346, opened 2017-02-14
SWARM-1009: Duplication of WEB-INF/lib artifacts in /m2repo of uber jar

heiko-braun:

-----

- Use synthetic module for app dependencies
- Derive dependencies from DependencyTree, rather then DependencyManager
- Leverage synthetic module for Arq tests as well


kenfinnigan:
I've been thinking on this some, and I feel that a synthetic module may not be the best approach as it takes us further away from "regular WAR" behavior.

I have some ideas around how we could do that, but haven't played with it yet

heiko-braun:
Let me first bring this too a point where we can decide on it. I think that helps to see the full picture and makes it easier to compare with an alternative.

kenfinnigan:
Earlier today I said we should postpone any further work on this until March sprint

heiko-braun:
right but that doesnt mean i abandon work in progress that minute you take the decision. I want to finish it properly. Dont worry i am not going to drag it into the next sprint, which iirc starts in two days, doesnt it?

kenfinnigan:
why not? Right now you don't have a timeline for how long it might take, and i'm not convinced this is the right approach, so why spend more time on it?

Next sprint starts 15th, tomorrow, though in JIRA its active now

PR #352, opened 2017-02-16
SWARM-542: Gradle download missing dependency

frami:
Motivation
----------
When developing with gradle, it should be possible to work without the local maven repository. This works after the pull request https://github.com/wildfly-swarm/wildfly-swarm/pull/342, but doesn't download it, if the artifact is missing in the local gradle cache. With this pull request, it will download the missing dependency.
[SWARM-542](https://issues.jboss.org/browse/SWARM-542)

-----


core/bootstrap/src/main/java/org/jboss/modules/maven/MavenArtifactUtil.java
```diff
@@ -18,6 +18,15 @@
```
kenfinnigan, line 18: I don't think the changes in this file are necessary, right?

Can we get this change removed from the PR?

frami: The change from private to public is necessary, but the change in the import not. If I shouldn't change the visibility, I need to copy the two methods in the GradleResolver.

kenfinnigan: Maybe an alternative approach is to create a parent class, or interface, with those methods that the Maven and Gradle helpers then use.

That way it's clear there's common functionality being reused vs Gradle using something from Maven, which isn't the case but could look that way from the calls.

If you could undo the import changes that would be great also.

Thanks

frami: I reverted the changed imports. I agree, it would be better to have a super class with the shared functionality. For that we have to rename the MavenResolver interface to ArtifactResovler and move it to an other package. 
From my perspective, this will be a big refactoring and I would prefer to do it later in a separate pull request.

kenfinnigan: no problem, we can tackle that another time

kenfinnigan (approved):

kenfinnigan:
ok to test

kenfinnigan:
@frami could you take a look at the test failure in the GradleResolverTest?

Thanks

PR #358, opened 2017-02-17
SWARM-1065 - Better logging granularity.

bobmcwhirter:
Motivation
----------

We need better control of what gets logged at which levels, and
be able to increase/decrease the output based on use-case.

Modifications
-------------

Split some SwarmMessages into Config and Metrics-related
messages.

Result
------

Somewhat less output by default.

-----


kenfinnigan (approved):
Do we need to renumber SwarmMessages to remove gaps?

bobmcwhirter:
I think that is a discussion we need to have.  This purely moved them to different categories, while retaining the original numbering.  So WFSWARM-00034 is still WFSWARM-0034, but moved from `org.wildfly.swarm` to `org.wildfly.swarm.config`.

I think numbering needs to be unique for the entire prefix (WFSWARM) irrespective of the category used.

I'm okay with "breaking" the current numbering to clean stuff up and institute a better numbering scheme (groups of 100 or 1000 or such by class)

kenfinnigan:
Good point, forgot about the prefixing thing.

Right, so maybe a group of 1000 per class?

bobmcwhirter:
retest this please

PR #384, opened 2017-02-26
SWARM-1131: Pass user defined settings.xml to arquillian tests

famod:
Motivation
----------
When building wildfly-swarm with an alternate settings.xml via mvn -s ..., arquillian tests still use the default one.

Modifications
-------------
Pass built-in Maven property userSettingsFile to forked surefire test processes in form of a ShrinkWrap Resolvers built-in system property which is automatically picked up.
See also: https://developer.jboss.org/message/969270#969270

Result
------
Arquillian test execution is now consistent with Maven execution/session.

-----

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -139,6 +139,7 @@
             <java.io.tmpdir>${project.build.directory}</java.io.tmpdir>
             <swarm.bind.address>127.0.0.1</swarm.bind.address>
             <project.version>${project.version}</project.version>
+            <org.apache.maven.user-settings>${session.request.userSettingsFile.path}</org.apache.maven.user-settings>
           </systemPropertyVariables>
         </configuration>
       </plugin>
@@ -173,6 +174,7 @@
               <testSourceDirectory>src/it/java</testSourceDirectory>
               <systemPropertyVariables>
                 <java.io.tmpdir>${project.build.directory}</java.io.tmpdir>
+                <org.apache.maven.user-settings>${session.request.userSettingsFile.path}</org.apache.maven.user-settings>
               </systemPropertyVariables>
             </configuration>
           </execution>
--- a/pom.xml
+++ b/pom.xml
@@ -178,6 +178,13 @@
               </systemPropertyVariables>
             </configuration>
           </execution>
+          <execution>
+            <id>verify-integration-test</id>
+            <phase>verify</phase>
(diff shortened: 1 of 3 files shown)


bobmcwhirter:
Ok to test

famod:
Please hold this PR:
```VerifyBuildArtefactIT``` in the ```multi-module-view``` testuite does not seem to pick up the specified settings.xml. Need to investigate first.

**Comment by famod** [comment]
I found the reason but this is very strange:
In root ```pom.xml``` there is a pre-defined execution for failsafe-plugin which binds to phase ```integration-test```.
There I added the new configuration property which means it will only be active for that very execution.

What puzzles me is that there are a couple of modules that explicitly (re-)define failsafe-plugin execution to ```integration-test``` **and** ```verify``` which results in **duplicate test executions**!
```
[someuser@localhost wildfly-swarm]$ grep -R --include=pom.xml 'goal>verify<' .
./standalone-servers/pom.xml:                  <goal>verify</goal>
./testsuite/testsuite-buildtool/multi-module/view/pom.xml:              <goal>verify</goal>
./testsuite/testsuite-maven-plugin/src/test/resources/testing-project/pom.xml:              <goal>verify</goal>
./testsuite/testsuite-maven-plugin/target/test-classes/testing-project/pom.xml:              <goal>verify</goal>
```
In fact the first execution of ```VerifyBuildArtefactIT``` is fine (phase ```integration-test```) but the second one fails (phase ```verify```) because the property is missing.

@bobmcwhirter: Shall I remove the superflous bindings to ```verify``` or shall I just move the property from the execution specific configuration to the general plugin configuration?

famod:
After checking http://maven.apache.org/surefire/maven-failsafe-plugin/usage.html I now do understand that you _need_ to define ```integration-test``` **and** ```verify```.
But you should only do this once, _either_ in root pom ```pluginManagement``` _or_ in each module individually.

famod:
I removed the executions without execution ids (which caused a second failsafe-plugin run) from the respective modules. Root pom.xml now contains everything that is needed.

I did see ```testsuite/testsuite-maven-plugin/src/test/resources/testing-project/pom.xml``` but that seems something different altogether so I left this file untouched.

kenfinnigan (approved):

PR #394, opened 2017-03-02
SWARM-900: Resource security protection outside main()

heiko-braun:

---

Motivation
----------

Support HTTP security constraints to be configured outside main()

Modifications
-------------

Move parsing to undertow archive preparer and consolidate keycloak and standard auth method config layout

Result
------

Basic, Digest, Form and Keycloak constraints can be configured in project-defaults.yml

### DIFF
--- a/fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HttpSecurityPreparer.java
+++ b/fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HttpSecurityPreparer.java
@@ -1,0 +1,149 @@
+/**
+ * Copyright 2015-2017 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.undertow.runtime;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.jboss.logging.Logger;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.Node;
+import org.jboss.shrinkwrap.api.asset.Asset;
+import org.wildfly.swarm.spi.api.ArchivePreparer;
+import org.wildfly.swarm.spi.api.JARArchive;
+import org.wildfly.swarm.spi.api.annotations.Configurable;
+import org.wildfly.swarm.undertow.descriptors.JBossWebAsset;
+import org.wildfly.swarm.undertow.descriptors.SecurityConstraint;
+import org.wildfly.swarm.undertow.descriptors.WebXmlAsset;
+
+/**
+ * @author Heiko Braun
+ */
+@ApplicationScoped
+public class HttpSecurityPreparer implements ArchivePreparer {
+
+    private static final Logger LOG = Logger.getLogger(HttpSecurityPreparer.class);
+
+    private final String[] SUPPORTED_AUTH_METHODS = new String[] {"BASIC", "DIGEST", "FORM", "KEYCLOAK"};
+
+    @Override
+    public void prepareArchive(Archive<?> archive) {
+
+        if (httpConfig == null || httpConfig.isEmpty()) {
+            return;
+        }
+
+        if (!httpConfig.containsKey("login-config")) {
+            throw new RuntimeException("Syntax error: HTTP security constraints requires a login-config declaration");
+        }
+
+        // unsupported auth method
+        Map<String, Object> loginConfig = (Map<String, Object>)httpConfig.get("login-config");
+        String authMethod = (String)loginConfig.getOrDefault("auth-method", "NONE");
+        boolean isSupported = false;
+        for (String supported : SUPPORTED_AUTH_METHODS) {
+            if (authMethod.equals(supported)) {
+                isSupported = true;
+                break;
+            }
+        }
+
+        if (!isSupported) {
+            LOG.warn("Ignoring unsupported auth-method: " + authMethod);
+            return;
+        }
+
+        WebXmlAsset webXml = findWebXml(archive);
+        JBossWebAsset jbossWeb = findJBossWeb(archive);
+
+        // Setup web.xml
+        webXml.setContextParam("resteasy.scan", "true");
+        webXml.setLoginConfig(authMethod, "ignored");
+
+        // security domain
+        if (loginConfig.containsKey("security-domain")) {
+            jbossWeb.setSecurityDomain((String)loginConfig.get("security-domain"));
+        }
+
+        // form login
+        if (loginConfig.containsKey("form-login-config")) {
+            Map<String, Object> formLoginConfig = (Map<String, Object>) loginConfig.get("form-login-config");
+            webXml.setFormLoginConfig(
+                    "Security Realm",
+                    (String)formLoginConfig.get("form-login-page"),
+                    (String)formLoginConfig.get("form-error-page")
+                    );
+        }
+
+        // security constraints
+        List<Map<String, Object>> securityConstraints =
+                (List<Map<String, Object>>) httpConfig.getOrDefault("security-constraints", Collections.EMPTY_LIST);
+
+        for (Map<String, Object> sc : securityConstraints) {
+            SecurityConstraint securityConstraint = webXml
+                    .protect((String) sc.getOrDefault("url-pattern", ""));
+
+            ((List<String>) sc.getOrDefault("methods", Collections.emptyList()))
+                    .forEach(securityConstraint::withMethod);
+
+            ((List<String>) sc.getOrDefault("roles", Collections.emptyList()))
+                    .forEach(securityConstraint::withRole);
+        }
+    }
+
+    private WebXmlAsset findWebXml(Archive<?> archive) {
+        WebXmlAsset webXml;
+        Node node = archive.as(JARArchive.class).get("WEB-INF/web.xml");
+        if (node == null) {
+            webXml = new WebXmlAsset();
+            archive.as(JARArchive.class).add(webXml);
+        } else {
+            Asset asset = node.getAsset();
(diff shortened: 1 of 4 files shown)


heiko-braun:
https://github.com/wildfly-swarm/wildfly-swarm-examples/pull/135 contains examples of the security constraints in project-defaults.yml. 

I.e. https://github.com/heiko-braun/wildfly-swarm-examples/blob/d954ad4f8645a6269f5c1aca9f5c5dd1e049ae06/security/jaas/basic/src/main/resources/project-defaults.yml

heiko-braun:
@kenfinnigan some feedback on the yml layout would be good, before we nail it down.

https://github.com/heiko-braun/wildfly-swarm-examples/blob/d954ad4f8645a6269f5c1aca9f5c5dd1e049ae06/security/jaas/basic/src/main/resources/project-defaults.yml

fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HttpSecurityPreparer.java
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+
+        if (!isSupported) {
+            LOG.warn("Ignoring unsupported auth-method: " + authMethod);
+            return;
+        }
+
+        WebXmlAsset webXml = findWebXml(archive);
+        JBossWebAsset jbossWeb = findJBossWeb(archive);
```
kenfinnigan: These shouldn't be needed.

It should be possible to:

  ```
WARArchive war = archive.as(WARArchive.war);
  war.addContextParam("resteasy.scan", "true");
```

Anything in web.xml or jboss-web.xml should be accessible from the WARArchive, making the find methods redundant.

File: fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HttpSecurityPreparer.java [hunk 2]
```diff
@@ -0,0 +1,149 @@
...
+
+        // Setup web.xml
+        webXml.setContextParam("resteasy.scan", "true");
```
kenfinnigan: I'm a bit wary of setting a RESTEasy web.xml property from within undertow.

Though it probably won't cause an issue, it does look a bit odd.

fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HttpSecurityPreparer.java
```diff
@@ -0,0 +1,149 @@
...
+public class HttpSecurityPreparer implements ArchivePreparer {
+
+    private static final Logger LOG = Logger.getLogger(HttpSecurityPreparer.class);
+
+    private final String[] SUPPORTED_AUTH_METHODS = new String[] {"BASIC", "DIGEST", "FORM", "KEYCLOAK"};
+
+    @Override
+    public void prepareArchive(Archive<?> archive) {
+
+        if (httpConfig == null || httpConfig.isEmpty()) {
```
kenfinnigan: Did you decide not to put the yaml config under a "deployment" key in the yaml?

If it's under "swarm.http" it will be triggered if "swarm.http.port" is set and cause a runtime failure when no security settings were required.

kenfinnigan (changes requested):

fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HttpSecurityPreparer.java
heiko-braun: good point

fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HttpSecurityPreparer.java
heiko-braun: Does as `archive.as(WARArchive.war)` support existing web.xml files?

fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HttpSecurityPreparer.java
heiko-braun: I am not sure it's even needed. It was present in the keycloak fraction and has been simply copied over here. But it should probably be removed.

fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HttpSecurityPreparer.java
kenfinnigan: It should do.

If it doesn't work, there's a bug in our WebXmlContainer

heiko-braun: I see, it's using the same fin methods

heiko-braun:
Some yml examples for security constraints:

Basic/Digest:
```
swarm:
  http:
    login-config:
      auth-method: BASIC
      security-domain: my-domain
    security-constraints:
      - url-pattern: /*
        methods: [GET]
        roles: [admin]
```

Form based:

```
swarm:
  http:
    login-config:
      auth-method: FORM
      form-login-config:
        form-login-page: /login.jsp
        form-error-page: /login_error.jsp
      security-domain: my-domain
    security-constraints:
      - url-pattern: /*
        methods: [GET]
        roles: [admin]
```

Keycloak:

```
swarm:
  http:
    login-config:
      auth-method: KEYCLOAK
    security-constraints:
      - url-pattern: /*
        methods: [GET]
        roles: [admin]
```

As you can see, the former `web.xml` and `jboss-web.xml` converge. Do you think the contents and structure is appropriate? Should it stay that way or does anything jump to your eyes?

@kenfinnigan @bobmcwhirter Any thoughts or suggestions?

kenfinnigan:
I recall the other day you mentioning that you wanted to add these under a deployment type key.

Maybe it makes sense to group these in a format such as:

```
swarm:
  deployment:
    login-config:
      auth-method: KEYCLOAK
    security-constraints:
      - url-pattern: /*
        methods: [GET]
        roles: [admin]
```

or even:

```
swarm:
  deployment:
    web:
      login-config:
        auth-method: KEYCLOAK
      security-constraints:
        - url-pattern: /*
          methods: [GET]
          roles: [admin]
```

heiko-braun:
> deployment type keys

What I suggested was to use the deployment names as keys to the security constraints:

```
swarm:
  deployment:
     my.war:
       login-config:
          [...]
```
But I dropped the idea, because I believe the url patterns already provide the level of distinction needed and thus switched to `swarm.http` as the main key

kenfinnigan:
Couple of comments:

- Proposal works fine for a single deployment only. I don't know whether we need to care about multiple deployments too much, but worth mentioning.
- There was a concern a while back that this type of usage mixes server and application config settings. Would it make sense to split the deployment specific bits out into a separate key under 'swarm' so that it's clearly separate? For me, 'http' doesn't jump out as being related to web urls of deployment, it seems more related to something configuring http for undertow.

heiko-braun:
But if we are unsure about future use cases we can design it with deployment keys with and support wildcards for the common case:

```
swarm:
  deployment:
     *:
       login-config:
          [...]
```

heiko-braun:
Yes, your arguments make sense. If we want to be on the safe side, this would be the best I think:

```
swarm:
  deployment:
     *:
       login-config:
          [...]
  http:
     port: 8080
```

It separates `deployment` from `http` and uses deployments names (with wildcard fallback) in case we ever need to support multiple deployment configurations

bobmcwhirter:
I'd suggest another level:

    swarm:
      deployment:
        *:
          undertow:
            login-config: ...
          keycloak:
            realm: ...

emag:
I think the last @bobmcwhirter style is good. The Keycloak Adapter Config is also related to per-deployment thing.

kenfinnigan:
I'd prefer not to reference specific fraction names as another level, favoring `web` or the like as another level.

Also, i'd be loathe to add Keycloak Client API pieces into anything under `swarm.deployment.*` as we're then having to explicitly map from our yaml back into WF subsystem, vs it happening by default if `swarm.keycloak.*` is used instead.

I'd be ok with `swarm.deployment.*.keycloak` providing a way to populate a `keycloak.json` for the deployment.

emag (changes requested):

heiko-braun:
The latest commit now relies on the following structure:

```
swarm:
  deployment:                                     (1)
    app.war:                                         (2)
      web:                                             (3)
        login-config:
          auth-method: BASIC
        security-constraints:
          - url-pattern: /protected
            methods: [GET, POST]
            roles: [admin]
```

This follows @kenfinnigan's suggestions and seems to be most distinctive with little overlapping concerns.

1.  `deployment` subsection of configuration to address ArchivePreparers
2. specific deployment names to support multiple deployments if needed
3. `web` subsection to clearly address `web.xml` and `jboss-web.xml` settings

fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HttpSecurityPreparer.java
```diff
@@ -0,0 +1,107 @@
...
+                    );
+        }
+
+        // security constraints
+        List<Map<String, Object>> securityConstraints =
+                (List<Map<String, Object>>) httpConfig.getOrDefault("security-constraints", Collections.EMPTY_LIST);
+
+        for (Map<String, Object> sc : securityConstraints) {
+            SecurityConstraint securityConstraint = webXml
+                    .protect((String) sc.getOrDefault("url-pattern", ""));
```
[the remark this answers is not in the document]
heiko-braun: thanks @emag, good catch. it got accidentally removed

emag (approved):

fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/HttpSecurityPreparer.java
```diff
@@ -0,0 +1,124 @@
...
+        if (deploymentConfigs == null || deploymentConfigs.isEmpty()) {
+            return;
+        }
+
+        // find a matching archive declaration
+        Optional<String> match = deploymentConfigs.keySet()
+                .stream()
+                .filter(c -> archive.getName().equals(c))
+                .findFirst();
+        if (!match.isPresent()) {
```
emag, line 56: Does wild card(`*`) not supported?

heiko-braun: The yaml parser chokes on it.

emag: Ah, OK.

kenfinnigan (approved):

PR #397, opened 2017-03-02
SWARM-1115 - @DefaultDeployment should ensure persistence.xml is loca…

bobmcwhirter:
…ted right.

Motivation
----------

For .wars, src/main/resources/META-INF/persistence.xml is weird and should
end up in WEB-INF/classes/META-INF/persistence.xml.

Modifications
-------------

Change DefaultDeployment to handle the special-case of persistence.xml.

Result
------

@DefaultDeployment is slightly more useful.

-----

### DIFF
--- a/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/DefaultDeploymentScenarioGenerator.java
+++ b/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/DefaultDeploymentScenarioGenerator.java
@@ -109,8 +109,8 @@ public class DefaultDeploymentScenarioGenerator extends AnnotationDeploymentScen
                             Files.walkFileTree(e, new SimpleFileVisitor<Path>() {
                                 public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
                                     if (!file.toString().endsWith(".class")) {
-                                        Path location = e.relativize(file);
-                                        archive.add(new FileAsset(file.toFile()), javaSlashize(location));
+                                        String location = javaSlashize(handleExceptionalCases(archive, e.relativize(file)));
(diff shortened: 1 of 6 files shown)


emag:
This commit covers only persistence.xml? Sorry for not referring to it and the bad title in SWARM-1115, some JPA examples point a sql file for auto-loading(like import.sql).
I think if the app packaging is war, the all assets in src/main/resources should be put under WEB-INF/classes.

bobmcwhirter:
okay, I can do that. I wasn't sure.

arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/DefaultDeploymentScenarioGenerator.java
```diff
@@ -154,6 +154,16 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
         return Collections.singletonList(description);
     }
 
+    protected String handleExceptionalCases(Archive archive, String input) {
+        if (archive.getName().endsWith(".war")) {
+            if (input.startsWith("META-INF/")) {
```
emag: I think not only META-INF dir, but also all 'src/main/resources' things should be put in WEB-INF/classes.

emag: Ah, but maybe need to exclude project-\<name\>.yml ...

bobmcwhirter: Or maybe we put it all under WEB-INF/classes, and make our yaml-finding bits smart enough to look there.

kenfinnigan (approved):

PR #398, opened 2017-03-03
SWARM-1116: Monitor fraction doesn't work with @ConfigurationValue anymore

heiko-braun:

-----

Motivation
----------

Classloading issues when using monitor and cdi-config together in one project

Modifications
-------------

Separate health api in library and link it to deployments

Result
------

Classloading issues gone.


fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/InstallMonitorFilter.java
```diff
@@ -47,10 +42,11 @@
 
     @Override
     public void prepareArchive(Archive<?> archive) {
-        JARArchive jarArchive = archive.as(JARArchive.class);
-
-        Asset resource = new ClassAsset(HealthResponseFilter.class);
-        ArchivePath location = new BasicPath(PATH_CLASSES, AssetUtil.getFullPathForClassResource(HealthResponseFilter.class));
-        jarArchive.add(resource, location);
+        try {
```
kenfinnigan, line 33: I think this can be removed by creating a module.xml for health-api.jar and then adding that module to module.conf with export=true, similar to the jaxrs client api cdi extension we have.

pom.xml
```diff
@@ -1188,6 +1188,8 @@
 
     <module>arquillian</module>
 
+    <module>fractions/health-api</module>
```
kenfinnigan: As this isn't a fraction, I'm thinking we need to add a top level directory `apis`, or the like, to add this there, and also move `jaxrs-client-api` there as well.

heiko-braun: Yeah, it should be moved and I will also restore the previous package location so it remains compatible. 

I am fine with a top level `apis` module, but would suggest to move the `jaxrs-client-api` separately (i.e. create a new Jira for it.)

kenfinnigan: i don't see the problem with also moving the other module, but if you think its a bad idea that's fine

heiko-braun: ok, i'll cover it in one sweep

fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/InstallMonitorFilter.java
kenfinnigan: ignore this comment, just realized that monitor doesn't have a "deployment" module that things could be exposed from.

However, is there a method for archives where we can add a GA dependency, or is that only inside a main()?

heiko-braun: Sorry, I don't really understand the question. Can you elaborate?

heiko-braun:
On hold until an issue with the examples is resolved

fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/InstallMonitorFilter.java
kenfinnigan: So, did a quick check, and you're already calling the method I was thinking of `WARArchive.addDependency()`.

Did you try it with just GA? According to method that should work, and saves relying on version, etc.

heiko-braun: yes, but for some reason it didn't resolve the artefact without a specific version

kenfinnigan: Weird, no worries

heiko-braun:
From my POV it's good to merge

kenfinnigan (approved):

PR #401, opened 2017-03-03
SWARM-1035: Swarm Jar executable doesn't start in background

nelsongraca:
Motivation
----------
Fixing 2 problems with launch.sh, that could cause an executable jar not to start in the background

Modifications
-------------
-The syntax {{variable:default_value}} does not work, should be ${variable:-default_value}
-start-stop-daemon may not support the --no-close flag, checking for its availability was added


bobmcwhirter:
okay to test

kenfinnigan:
ok to test

kenfinnigan (approved):

PR #405, opened 2017-03-06
SWARM-1163: Provide interface for version, etc

kenfinnigan:

-----

Motivation
----------
Currently the WF Swarm version is only available as a system property. We want to make it, and other details, available an interface.

Modifications
-------------
Created SwarmInfo that is filtered during build to add the version and groupId.

Result
------
No impact to product functionality, but version is available through SwarmInfo.

### DIFF
--- a/core/container/src/main/java-templates/org/wildfly/swarm/SwarmInfo.java
+++ b/core/container/src/main/java-templates/org/wildfly/swarm/SwarmInfo.java
@@ -9,7 +9,7 @@ public interface SwarmInfo {
 
     final String GROUP_ID = "${project.groupId}";
 
-    default boolean isProduct() {
+    static boolean isProduct() {
         return VERSION.contains("-redhat-");
     }
 }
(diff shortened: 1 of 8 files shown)


core/container/src/main/java-templates/org/wildfly/swarm/SwarmInfo.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * @author Ken Finnigan
+ */
+public interface SwarmInfo {
+
+    final String VERSION = "${project.version}";
+
+    final String GROUP_ID = "${project.groupId}";
+
+    default boolean isProduct() {
```
gastaldi: This can be a static method (replace `default` with `static`)

gastaldi: Otherwise you won't be able to call it without implementing this interface somehow

core/container/src/main/java/org/wildfly/swarm/Swarm.java
```diff
@@ -123,8 +122,6 @@
 @Vetoed
 public class Swarm {
```
gastaldi, line 123: Another possibility is to make Swarm implements SwarmInfo, then it will not break code that use `Swarm.VERSION`

gastaldi (approved):
LGTM. Added some comments but that should not prevent merging of this PR

PR #411, opened 2017-03-07
SWARM-1087: Adjust to handle product config-api

kenfinnigan:

-----

Motivation
----------
Our productized wildfly-config-api doesn't include a means to generate a self signed certificate so we need to handle cases where that method isn't present.

Modifications
-------------
Throw an error when generate self signed cert requested and it's not available. Switch to calling the method through reflection to handle when it isn't there.

Result
------
No direct impact.

### DIFF
--- a/fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/CertInfoProducer.java
+++ b/fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/CertInfoProducer.java
@@ -17,7 +17,9 @@ import javax.inject.Singleton;
 
 import org.jboss.modules.Module;
 import org.jboss.modules.ModuleIdentifier;
+import org.wildfly.swarm.SwarmInfo;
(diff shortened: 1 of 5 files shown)


fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/runtime/CertInfoProducer.java
```diff
@@ -47,6 +49,9 @@
     @Singleton
     public CertInfo produceCertInfo() {
         if (generateSelfCertificate.get()) {
+            if (SwarmInfo.isProduct()) {
```
bobmcwhirter, line 52: Instead of the explicit check, can we just try to call it, and log the happy message if it fails or we can't find it?

That way, when/if WF/EAP supports it, we have zero code-change to make, aside from removing the now-dead code.

With this explicit check, once it's supported upstream, we have to remember to change this.

Else, looks great!

bobmcwhirter (approved):

PR #417, opened 2017-03-14
SWARM-1139 / SWARM-1168 / SWARM-1156 - Logging/stderr/printStackTrace.

bobmcwhirter:
Motivation
----------

Our logging is sub-par, many times just println()'ing.  We also are not
handling exceptions in a decent manner, just printing stack traces.

Modifications
-------------

Split WFSWARM into more logging prefixes and categories.  Renumber
(BREAKING CHANGE if you grep logs), avoid simple printStackTrace().

Modify a few SPI interfaces to allow "throws Exception" so that we
can actually percolate errors up instead of blindly trundling forward
when a customizer has an issue.

Result
------

Cleaning logs, better logging, better error-handling.

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
@@ -37,7 +37,6 @@ import org.jboss.as.controller.client.ModelControllerClient;
 import org.jboss.dmr.ModelNode;
 import org.jboss.jandex.Index;
 import org.jboss.jandex.Indexer;
-import org.jboss.logging.Logger;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.ArchivePath;
 import org.jboss.shrinkwrap.api.Filters;
@@ -59,6 +58,7 @@ import org.wildfly.swarm.container.DeploymentException;
 import org.wildfly.swarm.container.internal.Deployer;
 import org.wildfly.swarm.container.runtime.deployments.DefaultDeploymentCreator;
 import org.wildfly.swarm.container.runtime.wildfly.SimpleContentProvider;
+import org.wildfly.swarm.internal.DeployerMessages;
 import org.wildfly.swarm.internal.FileSystemLayout;
 import org.wildfly.swarm.internal.SwarmMessages;
 import org.wildfly.swarm.spi.api.ArchiveMetadataProcessor;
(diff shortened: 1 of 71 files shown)


core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
```diff
@@ -240,16 +230,17 @@ public void deploy(Archive<?> deployment) throws DeploymentException {
                 preparer.prepareArchive(deployment);
             }
 
-            if (this.debug) {
+            if (DeployerMessages.MESSAGES.isDebugEnabled()) {
+                DeployerMessages.MESSAGES.deploying(deployment.getName());
                 Map<ArchivePath, Node> ctx = deployment.getContent();
                 for (Map.Entry<ArchivePath, Node> each : ctx.entrySet()) {
-                    System.err.println(each.getKey() + " // " + each.getValue());
+                    DeployerMessages.MESSAGES.deploymentContent(each.getKey().toString());
```
kenfinnigan, line 237: Does this need to output key and value or just key now?

core/spi/src/main/java/org/wildfly/swarm/spi/api/ArtifactLookup.java
```diff
@@ -38,12 +38,8 @@ static ArtifactLookup get() {
...
             try {
                 return (ArtifactLookup) Class.forName("org.wildfly.swarm.internal.ArtifactManager").newInstance();
-            } catch (InstantiationException e1) {
-                e1.printStackTrace();
-            } catch (IllegalAccessException e1) {
-                e1.printStackTrace();
-            } catch (ClassNotFoundException e1) {
-                e1.printStackTrace();
+            } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e1) {
+                // TODO handle error
```
kenfinnigan, line 42: Guessing this was missed?

kenfinnigan (approved):
Few comments i've added, but awesome stuff!

core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
bobmcwhirter: The value was always crap like "InputStreamAsset" or "ByteArrayAsset"

bobmcwhirter: or actually just NodeImpl.

PR #423, opened 2017-03-16
SWARM-1210: gradle related elements pulled out from the prod build

michalszynkiewicz:

-----


kenfinnigan:
ok to test

kenfinnigan:
@michalszynkiewicz Could you rebase your PR as it looks like there are conflicting changes from a PR I just merged?

Also, as the Gradle pieces were put in their own Arquillian Adapter, how is that then used by someone with Gradle?

michalszynkiewicz:
rebase done. 

It seems I forgot to add the gradle arquillian adapter to the arquillian module, sorry.
Then it should work by simply adding the org.wildfly.swarm:arquillian dependency in gradle.
I will update the PR when I verify it works.

michalszynkiewicz:
The import that a gradle user would use is: `org.wildfly.swarm:gradle-arquillian-adapter`

arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
```diff
@@ -195,7 +195,7 @@ public void start(Archive<?> archive) throws Exception {
         // project dependencies
         FileSystemLayout fsLayout = FileSystemLayout.create();
         DeclaredDependencies declaredDependencies =
-                DependencyDeclarationFactory.newInstance(fsLayout).create(resolvingHelper);
+                DependencyDeclarationFactory.newInstance(fsLayout).create(fsLayout, resolvingHelper);
```
michalszynkiewicz, line 198: This is a bit ugly. It's because, having injected `DependencyDeclarationFactory` via SPI, I can no longer pass fsLayout as a constructor parameter.
An alternative I could think of is to introduce another layer of factory/provider but I think it would make the code even worse.

kenfinnigan (approved):

michalszynkiewicz:
thanks!

PR #429, opened 2017-03-18
SWARM-1212: Introduce flag to skip execution of PackageMojo.

famod:
Motivation
----------
Packaging is useless and just wastes time when using wildfly-swarm:run (useUberJar=false).

Modifications
-------------
Add "skip" parameter to PackageMojo. Many Maven plugins have this parameter.

Result
------
Packaging can be skipped when you don't need it.

-----

### DIFF
--- a/plugin/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
+++ b/plugin/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
@@ -69,6 +69,12 @@ public class PackageMojo extends AbstractSwarmMojo {
     @Parameter(alias = "hollow", defaultValue = "false", property = "swarm.hollow")
     protected boolean hollow;
 
+    /**
+     * Flag to skip packaging entirely.
+     */
+    @Parameter(alias = "skip", defaultValue = "false", property = "swarm.package.skip")
+    protected boolean skip;
+
     protected File divineFile() {
         if (this.project.getArtifact().getFile() != null) {
             return this.project.getArtifact().getFile();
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
```diff
@@ -69,6 +69,12 @@
     @Parameter(alias = "hollow", defaultValue = "false", property = "swarm.hollow")
     protected boolean hollow;
 
+    /**
+     * Flag to skip packaging entirely.
+     */
+    @Parameter(alias = "skip", defaultValue = "false", property = "swarm.package.skip")
```
gastaldi, line 75: This alias means that -Dskip is also valid?

gastaldi (approved):

gastaldi:
Ok to test

plugin/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
kenfinnigan: Isn't "skip" a fairly generic name? Could cause confusing overlap

bobmcwhirter:
It's swarm.package.skip on the CLI

plugin/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
famod: Actually I just blindly copied this from `hollow`. Does this really influence the CLI?

PR #430, opened 2017-03-18
SWARM-1218: Add RunNoForkMojo.

famod:
Motivation
----------
RunMojo always forks the maven lifecycle which is annoying/wasting time in some cases.

Modifications
-------------
As there is no way to toggle lifecycle forking, introduce new RunNoForkMojo.
maven-source-plugin, maven-javadoc-plugin and others have done the same in the past.

Result
------
Swarm application can be started faster.

-----

### DIFF
--- a/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/RunNoForkMojo.java
+++ b/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/RunNoForkMojo.java
@@ -1,0 +1,31 @@
+/**
+ * Copyright 2017 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.plugin.maven;
+
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+
+/**
+ * Non-forking version of {@link RunMojo}.
+ */
+@Mojo(name = "run-no-fork",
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/wildfly/swarm/plugin/maven/RunNoForkMojo.java
```diff
@@ -0,0 +1,31 @@
+/**
+ * Copyright 2015-20176 Red Hat, Inc, and individual contributors.
```
famod: Adjusting 2016 to 2017 went slightly wrong. Do I need to fix this?

kenfinnigan: Technically it probably needs to be 2017 on its own.

kenfinnigan:
I'd like to understand whether we need a separate mojo or we modify the existing one?

famod:
Removing `@Execute(phase = LifecyclePhase.PACKAGE)` from the existing Mojo might break backward compatibility for some folks so I did not touch it.
In fact, forking _can_ come in handy in certain cases.

I also don't see how we can "toggle" `@Execute`. That would be the best solution, though.

plugin/src/main/java/org/wildfly/swarm/plugin/maven/RunNoForkMojo.java
famod: Ok, it's now just 2017.

bobmcwhirter:
I guess I'm ignorant.  Removing this:

    @Execute(phase = LifecyclePhase.PACKAGE)

Doesn't actually avoid the fork (external Runtime.process) of the execution, it just avoids the :package lifecycle leading up to it?

famod:
Yes, exactly.

The only problem I see with removing the annotation from the _existing_ Mojo is that people are used to it.
People would need to call "mvn package wildfly-swarm:run" instead of just "mvn wildfly-swarm:run" to have the old behaviour.
Then there might also be documentation bits that would need adjustment.

PS: I'm fine with changing the existing Mojo, personally.

famod:
Oops, closing was no intended, reopened.

famod:
Rebased.

famod:
@kenfinnigan & @bobmcwhirter Is there anything blocking this from being merged?
Do you prefer to remove `@Execute` from the existing `RunMojo` instead?

bobmcwhirter:
I think (but not certain) some of the work @kenfinnigan is doing might help reduce the requirement to package prior to wildfly-swarm:run anyhow, I think probably best to hold off on this for a bit.  It might resolve itself shortly.

heiko-braun:
Closing due to inactivity

PR #444, opened 2017-03-26
[SWARM-1236] multistart goal: make the startup timeout configurable

psiroky:
Motivation
----------
30s as a startup timeout is not always enough. Big applications may
take longer to deploy.

Modifications
-------------
The <process> configuration now accepts <start.timeout.seconds>
parameter. The default value is still 30s.

Result
------
Startup timeout for individual processes is configurable.

-----

### DIFF
--- a/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/MultiStartMojo.java
+++ b/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/MultiStartMojo.java
@@ -162,10 +162,16 @@ public class MultiStartMojo extends AbstractSwarmMojo {
         for (PlexusConfiguration each : env.getChildren()) {
             executor.withEnvironment(each.getName(), each.getValue());
         }
+        int startTimeoutSeconds;
+        try {
+            startTimeoutSeconds = Integer.valueOf(props.getChild("start.timeout.seconds").getValue("30"));
+        } catch (NumberFormatException nfe) {
+            throw new IllegalArgumentException("Wrong format of the start timeout for " + artifact + "!. Integer expected.", nfe);
+        }
 
         try {
             SwarmProcess launched = executor.execute();
-            launched.awaitReadiness(30, TimeUnit.SECONDS);
+            launched.awaitReadiness(startTimeoutSeconds, TimeUnit.SECONDS);
             procs.add(launched);
         } catch (IOException | InterruptedException e) {
             throw new MojoFailureException("Unable to execute: " + artifact, e);
--- a/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StartMojo.java
+++ b/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StartMojo.java
@@ -132,8 +132,14 @@ public class StartMojo extends AbstractSwarmMojo {
                 } catch (InterruptedException e) {
                 }
             }));
+            int startTimeoutSeconds;
+            try {
+                startTimeoutSeconds = Integer.valueOf(this.properties.getProperty("start.timeout.seconds", "120"));
+            } catch (NumberFormatException nfe) {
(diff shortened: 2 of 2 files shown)


kenfinnigan:
ok to test

famod:
Shouldn't there be a regular `@Parameter` field for this or am I missing something?

kenfinnigan:
Good point, should use a Plugin parameter for setting

psiroky:
I've just followed what the plugin does when retrieving info about the configured processes. I guess there could be a `@Parameter`, but then it would be a global timeout for all the processes and there would no way to configure this per-process. Or am I missing something?

kenfinnigan:
That would be correct.

However, if we want a per process startup timeout we probably should define an `@Parameter` similar to https://github.com/wildfly-swarm/wildfly-swarm/blob/master/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/AbstractSwarmMojo.java#L69 as that is where some of the properties retrieved via child process come from.

Doing so could then also allow for setting https://github.com/wildfly-swarm/wildfly-swarm/blob/master/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StartMojo.java#L136 via the same property.

I think it makes to add a property to StartMojo that can be retrieved via the child process from the MultiStartMojo

psiroky:
Thanks for the additional details! I am still not sure I understand your proposal correctly though. I agree the param should be added to the `StartMojo` to make the timeout configurable also there. However, I am not following how would that work for the `MultistartMojo`. Could you please paste a sample XML configuration of the plugin which do that for the multistart mojo?

psiroky:
In case we don't need per-process timeout config, this is not really an issue I guess. It just seemed like something that could be handy.

kenfinnigan:
I'd imagine something like:

          <execution>
            <id>multistart</id>
            <phase>pre-integration-test</phase>
            <goals>
              <goal>multistart</goal>
            </goals>
            <configuration>
              <processes>
                <process>
                  <artifactId>example-ribbon-frontend</artifactId>
                </process>
                <process>
                  <artifactId>example-ribbon-time</artifactId>
                  <properties>
                    <swarm.http.port>8081</swarm.http.port>
                    <start.timeout.seconds>50</start.timeout.seconds>
                  </properties>
                </process>
                <process>
                  <artifactId>example-ribbon-time</artifactId>
                  <properties>
                    <swarm.http.port>8082</swarm.http.port>
                  </properties>
                </process>
                <process>
                  <artifactId>example-ribbon-events</artifactId>
                  <properties>
                    <swarm.http.port>8083</swarm.http.port>
                    <start.timeout.seconds>120</start.timeout.seconds>
                  </properties>
                </process>
              </processes>
            </configuration>
          </execution>

psiroky:
@kenfinnigan thanks! Now I see what you meant. Can this be done with the @Parameter? The whole `<process>` is being parsed manually, so I am not sure how to plug-in the @Parameter for the properties.

kenfinnigan:
Ok. Maybe @Parameter wouldn't work for that then

psiroky:
I am now looking at http://maven.apache.org/guides/mini/guide-configuring-plugins.html#Mapping_Complex_Objects which describes how to map complex objects into/from the plugin config. It seems this is the preferable solution. We would have 
```
@Parameter(alias = "processes")
protected List<Process> processes;
```
in the Multistart mojo class and then new `Process` class which would include all the necessary config options:
```
public class Process {
  @Parameter
  private String groupId;
  @Parameter
  private String artifactdId;
  @Parameter
  private Properties properties;
}
```

WDYT?

kenfinnigan:
Sounds good

psiroky:
@kenfinnigan for now I've just added the simplest change -- support for the configurable timeout via the current mechanism. I am still not sure about the `start.timeout.seconds` property name. Any other suggestions?

I will take a look at using the class (instead of just parsing the XML) later. It seems there is some merging going on, which makes this a bit harder.

kenfinnigan (approved):

PR #445, opened 2017-03-26
SWARM-1219: Avoid losing original exception when stop() fails

famod:
Motivation
----------
SWARM-1219 introduced an automatic Swarm termination via swarm.stop() in case .start() or .deploy() fails.
As stop() itself can throw an exception (especially if start() fails early, see SWARM-1234) additional effort is needed to prevent
that such an exception overlays/shadows the original exception (which would make error analysis much harder or even impossible).

Modifications
-------------
swarm.stop() is now wrapped in try/catch and it is made sure that the original exceptions survives and is printed out.

Additionally, RuntimeServer does not try to call SelfContainedContainer.stop() anymore when this container hasn't been started at all
due to an "early" error (see also SWARM-1234). Before, this call caused an NPE on SelfContainedContainer:160.

Result
------
On termination due to a start/deploy error, the original exception is retained and printed out (no information is lost).
RuntimeServer does not cause a misleading NPE on JVM shutdown anymore.

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/Swarm.java
+++ b/core/container/src/main/java/org/wildfly/swarm/Swarm.java
@@ -609,9 +609,30 @@ public class Swarm {
         Swarm swarm = new Swarm(args);
         try {
             swarm.start().deploy();
-        } catch (Throwable t) {
-            swarm.stop();
+        } catch (final VirtualMachineError vme) {
+            // Don't even try to swarm.stop() in case of OOM etc.
+            vme.printStackTrace();
+            System.exit(1);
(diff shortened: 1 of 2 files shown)


famod:
@kenfinnigan & @bobmcwhirter I'd like to point out that 2017.4.0 (which seems right around the corner) should _not_ be published without this change, IMHO.

kenfinnigan:
ok to test

kenfinnigan (approved):

famod:
Thanks, @kenfinnigan!

PR #451, opened 2017-04-03
SWARM-1252 - Don't treat on/off/yes/no as Yaml booleans.

bobmcwhirter:
Motivation
----------
We like our booleans to be true or false, not on or off.

Modifications
-------------
Override the Yaml parser to not treat on/off/yes/no as
magical boolean-like creatures.

Result
------
John Ament is happier.

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/config/ConfigViewFactory.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/config/ConfigViewFactory.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2015-2016 Red Hat, Inc, and individual contributors.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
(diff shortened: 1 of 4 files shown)


core/container/src/main/java/org/wildfly/swarm/container/config/ConfigViewFactory.java
```diff
@@ -166,14 +172,39 @@ private void loadYamlProjectConfig(String name, URL url) throws IOException {
...
+    }
+
+    private static Yaml newYaml() {
+        return new Yaml(new Constructor(),
+                        new Representer(),
+                        new DumperOptions(),
+                        new Resolver() {
+                            @Override
+                            public Tag resolve(NodeId kind, String value, boolean implicit) {
+                                if (value != null) {
```
gastaldi, line 194: I think it could use: 
``` if (BOOL.matches(value)) return Tag.STR; ```

bobmcwhirter: We only want a partial special-case, not a complete retargetting of everything to STR.

gastaldi (dismissed):

heiko-braun:
@gastaldi Approved although tests fail?

gastaldi:
I forgot to add a comment: Merge it only if tests pass ;)

kenfinnigan (approved):

PR #459, opened 2017-04-13
SWARM-1001: Add dependencies to WAR default deployments.

mkouba:
Motivation
----------
A default deployment of type WAR should also include the non-fraction project dependencies which
would be normally added to WEB-INF/lib.

Modifications
-------------
For a default deployment of type WAR add the DependenciesContainer.ALL_DEPENDENCIES_MARKER.
This marker is later detected in UberjarSimpleContainer.
Explicit non-fraction dependencies are added as a libraries to a ShrinkWrap test archive.

Result
------
WAR default deployments also contain non-fraction project dependencies in WEB-INF/lib.

-----

### DIFF
--- a/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
+++ b/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
@@ -42,6 +42,8 @@ import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.Node;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
 import org.jboss.shrinkwrap.api.container.ClassContainer;
+import org.jboss.shrinkwrap.api.container.LibraryContainer;
+import org.jboss.shrinkwrap.api.container.ManifestContainer;
 import org.jboss.shrinkwrap.api.exporter.ZipExporter;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
@@ -54,6 +56,7 @@ import org.wildfly.swarm.bootstrap.util.BootstrapProperties;
 import org.wildfly.swarm.internal.FileSystemLayout;
 import org.wildfly.swarm.spi.api.DependenciesContainer;
 import org.wildfly.swarm.spi.api.JARArchive;
+import org.wildfly.swarm.spi.api.MarkerContainer;
 import org.wildfly.swarm.spi.api.SwarmProperties;
 import org.wildfly.swarm.spi.api.internal.SwarmInternalProperties;
 import org.wildfly.swarm.tools.ArtifactSpec;
@@ -201,9 +204,15 @@ public class UberjarSimpleContainer implements SimpleContainer {
         // check for "org.wildfly.swarm.allDependencies" flag
         // see DependenciesContainer#addAllDependencies()
         if (archive instanceof DependenciesContainer) {
-            DependenciesContainer depContainer = (DependenciesContainer) archive;
-            if (depContainer.hasMarker("org.wildfly.swarm.allDependencies")) {
+            DependenciesContainer<?> depContainer =  (DependenciesContainer<?>) archive;
+            if (depContainer.hasMarker(DependenciesContainer.ALL_DEPENDENCIES_MARKER)) {
                 munge(depContainer, declaredDependencies);
+            }
+        } else if (archive instanceof WebArchive) {
+            // Handle the default deployment of type WAR
+            WebArchive webArchive = (WebArchive) archive;
+            if (MarkerContainer.hasMarker(webArchive, DependenciesContainer.ALL_DEPENDENCIES_MARKER)) {
+                munge(webArchive, declaredDependencies);
             }
         }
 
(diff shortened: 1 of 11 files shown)


arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
```diff
@@ -201,10 +204,16 @@ public void start(Archive<?> archive) throws Exception {
         // check for "org.wildfly.swarm.allDependencies" flag
         // see DependenciesContainer#addAllDependencies()
         if (archive instanceof DependenciesContainer) {
-            DependenciesContainer depContainer = (DependenciesContainer) archive;
-            if (depContainer.hasMarker("org.wildfly.swarm.allDependencies")) {
+            DependenciesContainer<?> depContainer =  (DependenciesContainer<?>) archive;
+            if (depContainer.hasMarker(DependenciesContainer.ALL_DEPENDENCIES_MARKER)) {
                 munge(depContainer, declaredDependencies);
             }
+        } else if (archive instanceof WebArchive) {
```
kenfinnigan, line 211: Is there a situation where a WebArchive wouldn't implement DependenciesContainer?

core/spi/src/main/java/org/wildfly/swarm/spi/api/MarkerContainer.java
```diff
@@ -32,12 +32,19 @@
...
+        addMarker(this, markerName);
         return (T) this;
     }
 
     default boolean hasMarker(String markerName) {
-        return contains(ArchivePaths.create("META-INF/markers", markerName));
+        return hasMarker(this, markerName);
+    }
+
+    static void addMarker(ManifestContainer<?> manifestContainer, String markerName) {
```
kenfinnigan, line 43: What's the benefit of the static methods?

Also, should the marker methods both use `Archive` and not `ManifestContainer`?

kenfinnigan:
Add to whitelist

arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
mkouba: Yep, for instance:
https://github.com/wildfly-swarm/wildfly-swarm/blob/master/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/DefaultDeploymentScenarioGenerator.java#L61

kenfinnigan: But you've modified that above to set the marker if it's of type WAR

core/spi/src/main/java/org/wildfly/swarm/spi/api/MarkerContainer.java
mkouba: The benefit is we can use these methods for archives which do not implement `MarkerContainer` but fulfill the requirements.

> Also, should the marker methods both use Archive and not ManifestContainer?

I'm not sure. We could change the `PATH_MARKERS` to `META-INF/markers` but then we wouldn't allow the `ManifestContainer` to adjust the final path (e.g. WAR might use `WEB-INF/classes/META-INF/markers` instead). But the truth is that `addMarker()` and `hasMarker()` are not symmetric now... Unfortunately, `ManifestContainer` does not declare anything like `getManifestPath()`.

kenfinnigan: What kind of archives would not implement MarkerContainer but still fulfill requirements?

My concern is we're opening the door to non standard usage doing that

arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
mkouba: We're still using `WebArchive` which does not implement `DependenciesContainer`. `WARArchive` is defined in undertow fraction and arquillian-adapter does not have this fraction as a dependency. We can add this dep and the code would be clearer, but I'm not sure it's acceptable.

core/spi/src/main/java/org/wildfly/swarm/spi/api/MarkerContainer.java
mkouba: E.g. `WebArchive` mentioned in https://github.com/wildfly-swarm/wildfly-swarm/pull/459#discussion_r111376689.

arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
kenfinnigan: Nuts, missed that aspect.

I hate how there are essentially two WAR Archive types, one we've done and one in SW.

Someday we should fix that

mkouba: Yep, that's not very nice...

mkouba:
`org.wildfly.swarm.jaxrs.ZipkinJAXRSTest.testSpanLogging()` fails with:
> RESTEASY003190: Could not find constructor for class: org.jboss.resteasy.core.AsynchronousDispatcher

Investigating...

mkouba:
`org.jboss.resteasy:resteasy-client` brings in `org.jboss.resteasy:resteasy-jaxrs`. I tried to add an exclusion in the pom.xml but it does not seem to help... I would expect the exclusions to be honored. Will try to get more info.

mkouba:
Ok, so it really seems the `DeclaredDependencies.getTransientDependencies()` method returns also dependencies which are excluded in pom.xml. I could revert the `UberjarSimpleContainer.munge()` logic to only include the explicit dependencies -> the original behavior of `DependenciesContainer.addAllDependencies()` (sounds a little bit odd from user POV) or we can try to fix `DeclaredDependencies` somehow. @kenfinnigan WDYT?

mkouba:
So after several hours of debugging I was not able to find out how exactly the resolution works together with exclusions defined in pom.xml. 

My findings:
1. For `DefaultDeploymentProjectDependencyTest` (added in this PR) the `<exclusion>` defined in pom.xml **does work**. However, when looking at the source code I would suppose it should not work (e.g. the resolver used in `DependencyManager.resolveDependencies()` and `DependencyManager.analyzeRemovableDependencies()` does not seem to have any link to the pom.xml file).
2. For `ZipkinJAXRSTest` the `<exclusion>` defined in pom.xml **does not work** - `org.jboss.resteasy:resteasy-jaxrs` is added to the deployment and the test fails for obvious reasons.

So for now I would like to revert the `UberjarSimpleContainer.munge()` logic to only include the explicit dependencies -> the original behavior. I think adding explicit dependencies is better than nothing. @kenfinnigan? We can fix the transient dependencies later.

kenfinnigan:
Agreed

mkouba:
The builds seem to be stuck and ci.wildfly-swarm.io does not respond.

bobmcwhirter:
retest this please

kenfinnigan (approved):

PR #460, opened 2017-04-13
SWARM-1217: Do not add generated Application if servlet mapping exists.

mkouba:
Motivation
----------
According to the JAX-RS spec if no Application subclass is present and there is web.xml that specifies a
servlet mapping for a servlet with name javax.ws.rs.core.Application, the servlet is added automatically.
In this case, Swarm must not add a generated Application class.

Modifications
-------------
JAXRSArchiveImpl now attempts to parse the existing web.xml and does not add/removes the generated Application class.

Result
------
jboss-eap-quickstarts/helloworld-html5 should pass.

-----

### DIFF
--- a/fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/JAXRSMessages.java
+++ b/fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/JAXRSMessages.java
@@ -1,0 +1,40 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2017 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.swarm.jaxrs;
+
+import org.jboss.logging.BasicLogger;
+import org.jboss.logging.Logger;
+import org.jboss.logging.annotations.LogMessage;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageLogger;
+
+/**
+ * @author Martin Kouba
+ */
+@MessageLogger(projectCode = " WFSJAXRS", length = 4)
+public interface JAXRSMessages extends BasicLogger {
+
+    JAXRSMessages MESSAGES = Logger.getMessageLogger(JAXRSMessages.class, "org.wildfly.swarm.jaxrs");
+
+    @LogMessage(level = Logger.Level.INFO)
+    @Message(id = 1, value = "Unable to parse web.xml while searching for javax.ws.rs.core.Application servlet mapping:\n  %s")
+    void unableToParseWebXml(Object info);
+
+}
--- a/fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/internal/JAXRSArchiveImpl.java
+++ b/fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/internal/JAXRSArchiveImpl.java
@@ -20,6 +20,8 @@ import java.io.InputStream;
 import java.util.Map;
 import java.util.logging.Logger;
 
+import javax.ws.rs.core.Application;
+
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.ArchiveEvent;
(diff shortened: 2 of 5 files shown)


bobmcwhirter:
okay to test

bobmcwhirter:
add to whitelist

fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/JAXRSMessages.java
```diff
@@ -0,0 +1,42 @@
...
+import org.jboss.logging.BasicLogger;
+import org.jboss.logging.Logger;
+import org.jboss.logging.annotations.LogMessage;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageLogger;
+
+/**
+ * @author Martin Kouba
+ */
+@MessageLogger(projectCode = "WFSWARM", length = 4)
```
bobmcwhirter: I've actually been going for WFSSOMETHING for each new @MessageLogger, to avoid having to deal with offsets across the entire project.

We don't need to stick to just on WFSWARM.  WFSJAXRS works for me.

fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/internal/JAXRSArchiveImpl.java
```diff
@@ -99,8 +112,62 @@ private static boolean hasApplicationPathAnnotation(Archive<?> archive) {
...
 
+    private static boolean hasApplicationServletMapping(Archive<?> archive) {
+        Node webXmlNode = archive.get(PATH_WEB_XML);
+        if (webXmlNode != null) {
+            return hasApplicationServletMapping(webXmlNode.getAsset());
+        }
+        return false;
+    }
+
+    private static boolean hasApplicationServletMapping(Asset webXmlAsset) {
```
bobmcwhirter: Instead of this XML parsing and XPath, can you use the WebXmlAsset directly?

https://github.com/wildfly-swarm/wildfly-swarm/blob/master/fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/descriptors/WebXmlAsset.java#L80

Or add more accessors if we you need direct access to the underlying `WebAppDescriptor` which should already fully model the `web.xml`?

bobmcwhirter (changes requested):

Ladicek:
@mkouba asked me to test this PR, so I tried it with the test in SWARM-1217 and it passes. Yay :-)

fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/JAXRSMessages.java
mkouba: Ok. So we should also probably change the project code for `org.wildfly.swarm.datasources.runtime.DatasourcesMessages`.

fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/internal/JAXRSArchiveImpl.java
mkouba: > Instead of this XML parsing and XPath,...

Well, I think that parsing XML is more robust - in theory, you don't have to use `WebXmlAsset` but e.g. a `FileAsset`.

> Or add more accessors if we you need...

I don't understand. Could you explain in more details?

bobmcwhirter: The WebXmlAsset has an accessor for testing if a servlet exists.  If we need to add another accessor to test if a servlet-mapping exists, that cool.

I still prefer using the shrinkwrap descriptors, though, instead of XML/XPath to poke around, if you don't object.

fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/JAXRSMessages.java
bobmcwhirter: Yah, there are probably others that I've not successfully converted.

fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/internal/JAXRSArchiveImpl.java
mkouba: Well, my point is that you don't have to use `WebXmlAsset` to specify the `web.xml` in a test (which is by the way the case of the EAP quickstart mentioned in the issue description.. if I'm not wrong). You could use e.g. a `FileAsset` or `StringAsset`. And in that case the only reliable way to detect the servlet mapping is to parse the "real" `web.xml` contents.

A reasonable compromise might be to check `asset instanceof WebXmlAsset || asset instanceof WebAppDescriptor` and use a new accessor or `WebAppDescriptor.getAllServletMapping()` respectively but fallback to XML parsing otherwise.

mkouba:
@bobmcwhirter Ok, I finally understand ;-) and will modify the PR to use `WebXmlAsset(InputStream)`...

bobmcwhirter:
I think .as(WARArchive.class).getWebXmlAsset() will handle it all for you behind the scenes. Then you can scrub around in its contents.

bobmcwhirter (approved):

PR #462, opened 2017-04-14
SWARM-1051: Fix Undertow custom jboss-deployment-structure.xml

famod:
Motivation
----------
Currently, a custom jboss-deployment-structure.xml in webapp/WEB-INF is simply ignored.

Modifications
-------------
Instead of adding the "org.jboss.modules" module in WebXmlContainer, use the default approach via @DeploymentModule.
See also:
https://issues.jboss.org/browse/SWARM-1051?focusedCommentId=13392917&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13392917

Result
------
A custom jboss-deployment-structure.xml is now picked up correctly and is properly extended with various entries from the installed Swarm fractions.

-----

### DIFF
--- a/fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/UndertowFraction.java
+++ b/fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/UndertowFraction.java
@@ -28,6 +28,7 @@ import org.wildfly.swarm.spi.api.Defaultable;
 import org.wildfly.swarm.spi.api.Fraction;
 import org.wildfly.swarm.spi.api.annotations.Configurable;
 import org.wildfly.swarm.spi.api.annotations.ConfigurableAlias;
+import org.wildfly.swarm.spi.api.annotations.DeploymentModule;
 import org.wildfly.swarm.spi.api.annotations.MarshalDMR;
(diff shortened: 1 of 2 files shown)


kenfinnigan:
Not sure if this solves the jboss-deployment-structure.xml for all situations.

In the end I think `@DeploymentModule` references get added to that file anyway.

famod:
> Not sure if this solves the jboss-deployment-structure.xml for all situations.

Me neither.

But as this change fixes the problem that was originally reported in SWARM-1051 (I just checked that) and also fixes the problems in my current project I suggest to merge this.
Additionally, many users seem to be affected by this bug.

People can always file new bug tickets later in case a special `jboss-deployment-structure.xml` setup doesn't work for them.

Things that should be checked though:
- Why doesn't `ClassLoaderTest` run into the same problem? Must have something to do with the arquillian setup.
- Should there be a warning in the JavaDoc of `JBossDeploymentStructureContainer.addModule()` that no fraction should use this? Or can/should the method even be "fixed"?
- Is the "org.jboss.modules" module dependency really necessary at all?

bobmcwhirter (approved):

bobmcwhirter (approved):

PR #463, opened 2017-04-17
SWARM-1271 SWARM-1273: Activate project-whatever.yml better (and .yaml)

bobmcwhirter:
Motivation
----------
People like to be able to activate a project-*.yml based upon the
swarm.project.stage property, passed either directly to the JVM
or through the swarm -D-alike processing.  That didn't work.

Also, people, such as myself, can never remember .yml or .yaml as
the preferred extension, so hey presto, let's support both.

Modifications
-------------
Better scrubbing around and loading of stages mentioned in the
swarm.project.stage property, no matter where it appears.  Also,
scrub around from .yaml in addition to .yml.

Result
------
Ken's happier because his project-openshift.yaml is found.

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/config/ConfigViewFactory.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/config/ConfigViewFactory.java
@@ -104,7 +104,7 @@ public class ConfigViewFactory {
     public void load(String profileName, URL url) throws IOException {
         if (url.getPath().endsWith(".properties")) {
             loadProperties(profileName, url);
-        } else if (url.getPath().endsWith(".yml")) {
+        } else if (url.getPath().endsWith(".yml") || url.getPath().endsWith(".yaml")) {
             loadYaml(profileName, url);
         }
     }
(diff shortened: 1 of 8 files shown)


core/container/src/main/java/org/wildfly/swarm/container/config/ConfigViewFactory.java
```diff
@@ -104,7 +104,7 @@ public ConfigViewFactory load(String profileName) {
     public void load(String profileName, URL url) throws IOException {
         if (url.getPath().endsWith(".properties")) {
             loadProperties(profileName, url);
-        } else if (url.getPath().endsWith(".yml")) {
+        } else if (url.getPath().endsWith(".yml") || url.getPath().endsWith(".yaml")) {
             loadYaml(profileName, url);
         }
     }
```
famod, line 110: Just a thought: Shouldn't there at least be a warning log entry here in case neither .properties nor .yml/.yaml has been passed in?

PS: Might also apply to other methods.

bobmcwhirter: I don't think here, but a warning somewhere would be useful.

Problem is `-Pfoo` can be valid with `project-stages.yaml` defining a `foo` stage, so we can't explicitly care about the file names.

I'll follow with another PR if you don't mind adding a JIRA about issuing a warning for `an activated profile without a matching profile defined`.

kenfinnigan (approved):

PR #466, opened 2017-04-19
SWARM-1255: HealthCheck returns 500 - Internal Server Error due to a NPE

heiko-braun:

-----

Motivation
----------
Health endpoints yield an NPE when used with keycloak security

Modifications
-------------
Handle exception on InVm invocations gracefully and provide an InetSocketAddress to the internal invocations so that security contexts can be properly resolved
 
Result
------
The given reproducer works and so to the other health and security tests

### DIFF
--- a/fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/HttpContexts.java
+++ b/fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/HttpContexts.java
@@ -16,6 +16,7 @@
 package org.wildfly.swarm.monitor.runtime;
 
 import java.io.IOException;
+import java.net.InetSocketAddress;
 import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/HttpContexts.java
```diff
@@ -150,14 +151,13 @@ private void proxyRequests(HttpServerExchange exchange) {
...
-                        } else if (503 == resp.getStatus()) {
-                            sb.append(resp.getPayload());
-                            failed = true;
-                        } else {
-                            throw new RuntimeException("Unexpected status code: " + resp.getStatus());
+
+                        sb.append(resp.getPayload());
+
+                        if (!failed) {
+                            failed = resp.getStatus() != 200;
```
bobmcwhirter, line 158: for what it's worth, I think any response between 200 and 299 is considered non-failed.  Maybe irrelevant given what this bit does, but a strict 200 check might be too strict.

bobmcwhirter (approved):

fractions/monitor/src/main/java/org/wildfly/swarm/monitor/runtime/HttpContexts.java
heiko-braun: I was considering that, but than we again we control the response here https://github.com/wildfly-swarm/wildfly-swarm/blob/master/client-apis/health-api/src/main/java/org/wildfly/swarm/health/HealthResponseFilter.java

PR #475, opened 2017-04-25
SWARM-1282: Support a bom-certified.

bobmcwhirter:
Motivation
----------
We want, in the product build, another artifact :bom-certified
pointing to certified community bits that play friendly with
the productized bits, assuming there is a product.

Modifications
-------------
New project, boms/bom-certified, only enabled if -Dswarm.product.build.
Produces :bom-certified referencing bits from `certified.conf` near that
pom.xml using the fraction-list configured as a plugin dependency.

The dependency ultimately uses ${version.certified-community} defined
near the top of the root pom.xml, and in the community release is irrelevant
but currently matches itself.

Additionally, the plugin also avoids adding :arquillian from the productized
build (assuming there is a product) if -Dswarm.product.build, since :arquillian
would be purely and only a certified community fraction, if there is a product.

Result
------
When -Dswarm.product.build, a :bom-certified.

-----

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,9 @@
   </scm>
 
   <properties>
-    <version.wildfly.swarm.fraction.plugin>53</version.wildfly.swarm.fraction.plugin>
+    <version.certified-community>2017.5.0-SNAPSHOT</version.certified-community>
+
+    <version.wildfly.swarm.fraction.plugin>54</version.wildfly.swarm.fraction.plugin>
     <version.wildfly.swarm.checkstyle>3</version.wildfly.swarm.checkstyle>
 
     <version.org.snakeyaml>1.17</version.org.snakeyaml>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -1288,6 +1290,7 @@
     <module>plugins/maven</module>
     <module>boms</module>
     <module>boms/bom</module>
+    <module>boms/bom-certified</module>
```
bobmcwhirter: @kenfinnigan from IRC... apparently I need to adjust this some.

While we have a way to do community-only builds (`-Dswarm.product.build`) I think I need to add a profile to also only include stuff if a product build.

I think a profile activated by swarm.product.build (as opposed activated by it's lack) to include exactly bom-certified.

I think preventing bom-certified from being released during community releases would help prevent confusion in the community, since we're dealing with cross-versioning of things.

kenfinnigan (approved):

PR #478, opened 2017-04-26
SWARM-1133: Add test for SWARM-1133.

juangon:
Motivation
----------
Add tests to verify SWARM-1133 fix is working in future.

Modifications
-------------
Following a similar approach as in fraction-metadata, FractionUsageAnalyzerTest class, added a simple JUnit test with an MDB to check if right messaging fraction was actually added

Result
------
After these tests, they would only fail if SWARM-1133 is removed some day in future, keeping that feature safe.

-----

### DIFF
--- a/fractions/javaee/messaging/src/test/java/org/wildfly/swarm/messaging/detector/MessageDetectorTest.java
+++ b/fractions/javaee/messaging/src/test/java/org/wildfly/swarm/messaging/detector/MessageDetectorTest.java
@@ -1,0 +1,32 @@
+package org.wildfly.swarm.messaging.detector;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+import java.io.File;
+import java.nio.file.Files;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.exporter.ZipExporter;
+import org.junit.Test;
+import org.wildfly.swarm.fractions.FractionUsageAnalyzer;
+import org.wildfly.swarm.spi.api.JARArchive;
+
+public class MessageDetectorTest {
+
+	@Test
+	public void testFractionMatching() throws Exception {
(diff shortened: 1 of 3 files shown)


fractions/javaee/messaging/src/test/java/org/wildfly/swarm/messaging/detector/MessageDetectorTest.java
```diff
@@ -0,0 +1,32 @@
...
+	    FractionUsageAnalyzer analyzer = new FractionUsageAnalyzer();
+	
+	    final File out = Files.createTempFile(archive.getName(), ".war").toFile();
+	    archive.as(ZipExporter.class).exportTo(out, true);
+	
+	    analyzer.source(out);
+	    assertThat(analyzer.detectNeededFractions()
+	                       .stream()
+	                       .filter(fd -> fd.getArtifactId().equals("messaging"))
+	                       .count())
```
Ladicek, line 29: I'd probably use `stream.findAny().isPresent()`, but that's just nitpicking :-)

Ladicek: OK, I see that asserting count is equal to 1 is in the existing tests, fine with me :-)

Ladicek:
I'm not sure in which module this should be. The existing tests similar to this are in the `fraction-metadata` module, which then depends on the `jaxrs` fraction; here, the test is in the fraction module depending on `fraction-metadata`, so I'm slightly worried if there's a dependency cycle somewhere?

juangon:
Ok @Ladicek I finally thought this would be the best place, but that you said makes sense. I am going to try adding this into test-messaging meanwhile...

bobmcwhirter:
Labelling as HOLD until you give the all-clear.

juangon:
Moved to test-messaging here: https://github.com/wildfly-swarm/wildfly-swarm/pull/489

Thanks!

PR #489, opened 2017-05-03
SWARM-1133 Add tests

juangon:
Motivation
----------
Add tests to verify SWARM-1133 fix is working in future.

Modifications
-------------
Following a similar approach as in fraction-metadata, FractionUsageAnalyzerTest class, added a simple JUnit test with an MDB to check if right messaging fraction was actually added

Result
------
After these tests, they would only fail if SWARM-1133 is removed some day in future, keeping that feature safe.

-----

Hi @bobmcwhirter @Ladicek ! So after many problems I can't make these kind of tests working with Arquilian.

First there were some problems for adding some POM dependencies as libraries in Arquilian generated WAR file. Fortunately seems that was fixed after SWARM-1001 was committed.  Then I was stuck with some error due to different location for JARArchive and JARArchiveImpl (Impl could not be assigned as JARArchive), probably due to different classloader. JARArchive is located in fraction-metadata (bundled with Arquilian WAR) and JARArchiveImpl is located in Wildfl Swarm core library.

Said this, seems that there aren't any test using own Shrinkwrap archives and fraction-metadata using Arquilian, so I ended just adding other simple Junit test class.

Hope it makes sense.

P.D: this replaces previous pull: https://github.com/wildfly-swarm/wildfly-swarm/pull/478

### DIFF
--- a/testsuite/testsuite-messaging/src/test/java/org/wildfly/swarm/messaging/test/MessagingTest.java
+++ b/testsuite/testsuite-messaging/src/test/java/org/wildfly/swarm/messaging/test/MessagingTest.java
@@ -1,0 +1,31 @@
+package org.wildfly.swarm.messaging.test;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+import java.io.File;
+import java.nio.file.Files;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.exporter.ZipExporter;
+import org.junit.Test;
+import org.wildfly.swarm.fractions.FractionUsageAnalyzer;
+import org.wildfly.swarm.spi.api.JARArchive;
+
+public class MessagingTest {
+
+    @Test
+    public void testFractionMatching() throws Exception {
+	    JARArchive archive = ShrinkWrap.create(JARArchive.class);
+	    archive.addClass(MyTopicMDB.class);
+	    FractionUsageAnalyzer analyzer = new FractionUsageAnalyzer();
+
+	    final File out = Files.createTempFile(archive.getName(), ".war").toFile();
+	    archive.as(ZipExporter.class).exportTo(out, true);
+
+	    analyzer.source(out);
+	    assertThat(analyzer.detectNeededFractions()
+	                       .stream()
+	                       .filter(fd -> fd.getArtifactId().equals("messaging"))
+	                       .count()).isEqualTo(1);
+    }
+}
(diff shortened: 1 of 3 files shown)


Ladicek:
@juangon This is exactly how I'd write the test if I were to use `fraction-metadata`. I'd only use Arquillian if I used a more end2end approach. Looks like you learned a lot writing this simple test -- good :-)

+1 from me.

Ladicek (approved):

testsuite/testsuite-messaging/src/test/java/org/wildfly/swarm/messaging/test/MessagingTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class MessagingTest {
+
+    @Test
+    public void testFractionMatching() throws Exception {
+	    JARArchive archive = ShrinkWrap.create(JARArchive.class);
+	    archive.addClass(MyTopicMDB.class);
+	    FractionUsageAnalyzer analyzer = new FractionUsageAnalyzer();
+
+	    final File out = Files.createTempFile(archive.getName(), ".war").toFile();
```
Ladicek, line 22: Actually, one comment: this file should be deleted at the end of the test. `try/finally` looks like the best choice.

juangon:
Thanks @Ladicek. As this test is heavily based on this one: https://github.com/wildfly-swarm/wildfly-swarm/blob/master/meta/fraction-metadata/src/test/java/org/wildfly/swarm/fractions/FractionUsageAnalyzerTest.java#L37 does it mean the latter should be changed to remove the file too?

testsuite/testsuite-messaging/src/test/java/org/wildfly/swarm/messaging/test/MessagingTest.java
juangon: Thanks @Ladicek. As this test is heavily based on this one: https://github.com/wildfly-swarm/wildfly-swarm/blob/master/meta/fraction-metadata/src/test/java/org/wildfly/swarm/fractions/FractionUsageAnalyzerTest.java#L37 does it mean the latter should be changed to remove the file too?

Ladicek: Hmm, I think the temp file gets created somewhere in `target`, in which case deleting it isn't strictly necessary...

bobmcwhirter:
I think it creates them in $TMP, which ultimately should get cleaned up by the OS, maybe, at some point.  Unless Windows.

famod:
There is also [File.deleteOnExit()](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#deleteOnExit--).

bobmcwhirter:
okay to test

bobmcwhirter:
add to whitelist

bobmcwhirter (approved):

PR #492, opened 2017-05-04
SWARM-1322: Swarm's Arquillian container doesn't enable the @EJB enricher

Ladicek:
Motivation
----------
Arquillian ships with a bunch of test instance "enrichers"
for injecting objects to annotated fields. Among them are
enrichers for Java EE annotations `@Inject`, `@Resource`
and `@EJB`. These enrichers are enabled only if the respective
annotation (or some other identifying class) is "present
on the classpath". Which, for in-container tests, means
_visible from the classloader that loaded Arquillian_.
In Swarm's case, that's the module classloader for module
`org.jboss.arquillian`. This module is defined by Swarm itself.

This module currently exports `javax.annotation.api`, which
contains `@Resource`, and `javax.enterprise.api`, which
contains `@Inject` (and the `BeanManager`). For the `@EJB`
enricher to work, an export of `javax.ejb.api` is needed.

Modifications
-------------
Added an exported dependency on the `javax.ejb.api` module
to the `org.jboss.arquillian` module.

Result
------
`@EJB` enricher can actually work with Swarm Arquillian.

-----

### DIFF
--- a/arquillian/arquillian/src/main/resources/modules/org/jboss/arquillian/main/module.xml
+++ b/arquillian/arquillian/src/main/resources/modules/org/jboss/arquillian/main/module.xml
@@ -17,6 +17,7 @@
 
   <dependencies>
     <module name="javax.annotation.api" export="true"/>
+    <module name="javax.ejb.api" export="true"/>
     <module name="javax.enterprise.api" export="true"/>
     <module name="org.jboss.shrinkwrap" export="true"/>
     <module name="javax.api"/>


bobmcwhirter (approved):

Ladicek:
FTR, I do have a test for this, but can't currently attach it because of [ARQ-2109](https://issues.jboss.org/browse/ARQ-2109). By the time that is fixed and released, [SWARM-1326](https://issues.jboss.org/browse/SWARM-1326) will hopefully be fixed as well.

PR #501, opened 2017-05-16
SWARM-1330: Swarm file detection not detecting file path.

juangon:
Motivation
----------
Improve Swarm fraction through file existence, by determine the file path instead of just file name.

Modifications
-------------
Added new "FileSource" type so it can carry on needed file path. Made required changes in order to apply that.

Result
------
A new "FileSource" class is available for file detectors, so they can use it to improve the file detection by also checking file path.

-----


meta/fraction-metadata/src/main/java/org/wildfly/swarm/fractions/scanner/FilePresenceScanner.java
```diff
@@ -22,12 +22,13 @@
...
 import org.wildfly.swarm.spi.meta.FractionDetector;
 
 /**
  * @author Heiko Braun
  */
-public class FilePresenceScanner implements Scanner<String> {
+public class FilePresenceScanner implements Scanner<FileSource> {
     @Override
     public String extension() {
         return XML;
```
Ladicek, line 34: I know you didn't touch this line and I don't want you to do anything about it, but I was wondering some time before if this is possibly wrong? I mean, we currently only detect XML files, so it doesn't restrict anything, but maybe it could in the future? WDYT? Just a comment, nothing more.

juangon: Hmmm I don't quite understand your question. You mean extension() shouldn't be there? FileSource?

Ladicek: No no no, this question has nothing to do with your changes. I'm just wondering what's your opinion about `FilePresenceScanner.extension` always returning `"xml"`. I was thinking it might be overly restrictive if we ever detect on something else than just XML files, but it's probably a premature generalization on my side :-)

juangon:
Closing, will resend with the windows fix.

meta/fraction-metadata/src/main/java/org/wildfly/swarm/fractions/scanner/FilePresenceScanner.java
juangon: I think it's fine to have this kind of mechanism just in case someday file extension won't be XML. But maybe we could get rid of it as having the file name should be enough and extension is contained there too.

juangon:
Resent here: https://github.com/wildfly-swarm/wildfly-swarm/pull/502

PR #502, opened 2017-05-16
SWARM-1330: Swarm file detection not detecting file path.

juangon:
Motivation
----------
Improve Swarm fraction through file existence, by determine the file path instead of just file name.

Modifications
-------------
Added new "FileSource" type so it can carry on needed file path. Made required changes in order to apply that.

Result
------
A new "FileSource" class is available for file detectors, so they can use it to improve the file detection by also checking file path.

-----


bobmcwhirter:
While this looks good, I do wonder about the `File.separator` bits.

I think this is scrubbing around within .jar and .war archives, which always have `/`, while File.separator will be `\` on Windows systems?

juangon:
Hi @bobmcwhirter. I noticed this issue when CI windows tests weren't passing. It's clear that for exploded archives should have a File.separator to support all OSes. Seems that ZipEntry needs the system path separator too. I added an "exploded" test too, so this won't fail again no matter using exploded/unexploded archive and Linux, Windows, etc.

bobmcwhirter (approved):

PR #506, opened 2017-05-19
SWARM-1349 && SWARM-1313 - Deployment-level configurables.

bobmcwhirter:
To support SWARM-1313 (YAML configuration of Swagger deployments),
large-scale changes have occurred, supporting @DeploymentScoped
components.

The @DeploymentScoped components are:

* DeploymentProcessor, which has completely replaced both ArchivePreparer
  and ArchiveMetadataProcessor

* Archive and IndexView, which are @DeploymentScoped and can be @Injected
  into a DeploymentProcessor.

Additionally, within a @DeploymentScoped component, @Configurables
with a name pattern of `swarm.deployment.*.whatever` may be used, and
maps to YAML similar to:

    swarm:
      deployment:
        myapp.war:
          whatever: value

Current short-comings:

* Only Swagger-related things have deployment-specific configuration
* If your .war has a version encoded in it's name, your life will be more difficult

-----


fractions/cdi-extensions/cdi-jaxrsapi/src/main/java/org/wildfly/swarm/cdi/jaxrsapi/runtime/ClientServiceFactory.java
```diff
@@ -0,0 +1,166 @@
+package org.wildfly.swarm.cdi.jaxrsapi.runtime;
```
kenfinnigan, line 1: Was there a reason this needed to be pulled out?

Or just cleaner?

kenfinnigan (approved):

fractions/cdi-extensions/cdi-jaxrsapi/src/main/java/org/wildfly/swarm/cdi/jaxrsapi/runtime/ClientServiceFactory.java
bobmcwhirter: Ah. Just a side effect of me tracking down a bug that was ultimately in fraction plugin.

kenfinnigan: not a problem, more curiousity

bobmcwhirter: fwiw, the changed I slipped into the fraction-plugin (v55, released, used here) works a lot harder to ensure that when we add :deployment modules, and they add :main, that other things don't slip into the actual deployment module, making CDI grumpy.

I had an issue where jaxrsapi's :deployment module was mixed into the user's .war, which then depended on the jaxrsapi:main module (which is currect), but then Weld saw the `META-INF/beans.xml` in that module, found the jar, and then CDI'd the whole dang jar, including both the `*Fraction` and everything in its `runtime.*` package.  And then complained. 

I think it was complaining more quietly before, but this PR's changes resulted in more than a slight whinge, and caused full-blown errors.

So, now, :deployment modules depend on :main, as before, but explicitly exclude `META-INF` in that dependency, to avoid Weld going far afield to find components.

kenfinnigan: So if anything, fixing these things should reduce hidden errors. Nice

PR #507, opened 2017-05-19
SWARM-1351 - @Configurable and @ConfigurableAlias changes

bobmcwhirter:
Motivation
----------
Given that we support multiple configurable names (A name an alias
previous to this commit), we should be sure to have a preference,
most likely first one wins.

Given that they are just names, and JDK8 supports multiple repeated
annotations and they don't annoy Weld, let's replace @ConfigurableAlias
with simply multiple usages of @Configurable.  And once again, first one
wins.

Modifications
-------------
Added @Configurables to accomodate the structure of multiple @Configurable
on a field.  The plural is not required to be used directly. The JDK will
glom up the singulars to the plural if required.

Then, using the list of names we derive from the one-or-more Configurables
(or zero, really), attempt to apply them in-order until one succeeds.

Result
------
Less annotations, used more often.

-----


bobmcwhirter:
retest this please.

kenfinnigan (approved):

PR #508, opened 2017-05-23
SWARM-1354 Add FractionDetectMode to Gradle and fix the moduleDirs problem

SetoKaiba:
SWARM-1354 Add FractionDetectMode to Gradle and fix the moduleDirs problem

### DIFF
--- a/plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/PackageTask.java
+++ b/plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/PackageTask.java
@@ -17,6 +17,7 @@ package org.wildfly.swarm.plugin.gradle;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -32,6 +33,7 @@ import org.gradle.api.Task;
 import org.gradle.api.artifacts.ResolvedConfiguration;
 import org.gradle.api.artifacts.ResolvedDependency;
 import org.gradle.api.plugins.ApplicationPluginConvention;
+import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.InputFile;
 import org.gradle.api.tasks.InputFiles;
@@ -39,6 +41,7 @@ import org.gradle.api.tasks.Optional;
 import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.TaskAction;
 import org.gradle.api.tasks.bundling.Jar;
+import org.gradle.api.tasks.SourceSet;
 import org.wildfly.swarm.fractions.PropertiesUtil;
 import org.wildfly.swarm.spi.meta.SimpleLogger;
 import org.wildfly.swarm.tools.ArtifactSpec;
@@ -49,6 +52,8 @@ import org.wildfly.swarm.tools.DeclaredDependencies;
  * @author Bob McWhirter
  */
 public class PackageTask extends DefaultTask {
+
+    private static final String MODULE_DIR_NAME = "modules";
 
     private BuildTool tool;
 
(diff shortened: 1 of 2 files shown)


Ladicek:
ok to test

plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/SwarmExtension.java
```diff
@@ -53,6 +54,8 @@
 
     private Jar archiveTask;
 
+    private BuildTool.FractionDetectionMode fractionDetectMode;
```
Ladicek: Oh, this should definitely have a default value of `when_missing`.

SetoKaiba: Ok.

SetoKaiba:
A problem of Gradle plugin as well. 

https://github.com/wildfly-swarm/wildfly-swarm/blob/master/plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/PackagePlugin.java#L38
The SwarmExtension is created in when the apply plugin.

https://github.com/wildfly-swarm/wildfly-swarm/blob/master/plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/SwarmExtension.java#L58-L70
Additional modules are initialized here.

But the resourceOutputDir will only come existing after the archive(jar/war) task.
Then the default modules won't be added.
So I have to package twice to package the modules. Or I have to run archive(jar/war) separately before package.

I made a fix for it as well. Any better suggestion?

SetoKaiba:
@Ladicek Please review. Thank you.

plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/PackageTask.java
```diff
@@ -66,6 +71,23 @@ public void packageForSwarm() throws Exception {
 
         GradleArtifactResolvingHelper resolvingHelper = new GradleArtifactResolvingHelper(project);
         Properties propertiesFromExtension = getPropertiesFromExtension();
+        List<File> moduleDirs = getModuleDirs();
+        if (moduleDirs.isEmpty()) {
```
Ladicek, line 75: Looking at the deleted code, it seems the original indention was to _add_ the `modules` dir, if it exists, to the configured ones. But here you use it only if there's no configured module directory. Is that ok?

SetoKaiba: You can check the original code.
Before
Case modulesDir is not set.
When SwarmExtension is created, it will added the default from resourceOutputDir.
It will use the default. But the situation only works war is run before.
Case modulesDir is set
Clear the list which is initialized when SwarmExtension is created, add the new value to the list.
If will use the new value from it.

After
Case modulesDir is not set.
modulesDir is empty. it will added the default from resourceDir when the task run. The only change is that it's in the task instead of creation of SwarmExtension.
Case modulesDIr is set
modulesDIr is set to new value. The same as it previously does. The only difference is when the new value is empty list. It will use default. But it's meaningless setting an empty list, right?

SetoKaiba: https://github.com/SetoKaiba/wildfly-swarm/blob/ef7b403f6a930f9102b823eb80575b0ff3cbb471/plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/SwarmExtension.java#L114-L117

Maybe you forget to check the unmodified code. The setter of moduleDirs.

SetoKaiba: The manually setting of moduleDirs will clear the list first. That means the added moduleDirs during creation is cleared before the new value is added.

Ladicek: OK, I'm not a skilled Gradle user, but I think that setting `modulesDir` from `build.gradle` doesn't necessarily overwrite the existing value -- it can, but it can also just append to it. Right?

Ladicek: Ah, OK, you're right, indeed I didn't check that method.

Ladicek (approved):

Ladicek:
retest this please

Ladicek:
... looks like I'm not authorized to trigger a CI build.

bobmcwhirter:
retest this please

PR #509, opened 2017-05-23
SWARM-1222: Fix security web xml detector and add security fraction detection test.

juangon:
-----
Hi @bobmcwhirter @Ladicek. Although code could be a bit simpler in WebXmlDescriptorScanner.java I preferred adding those checks just in case the inputstream passed to method is changed at some time , to avoid coupling between method parameters and what we are expecting. That's why I added checks for any inputstream. Hope it makes sense.


Ladicek:
I don't understand what does this do, and I don't really understand the uncloseable stream stuff. If that's because we need to read from the stream multiple times, then maybe we can switch from using an input stream to a string?

juangon:
Hi @Ladicek. There is already a consistent API across all detectors that is using inputstream etc. Additionally, changing API would require much more changes than these ones AFAIK.

juangon:
BTW @Ladicek , about the uncloseable stuff I added an explanation in commit, "Modifications" section:

"Modifications
-------------
After fixing security detector, other bug arised, because shrinkwrap utility class for parsing web.xml (Descriptors) was closing the stream, which caused error when a second webxml fraction detector was executed (for example, JSFServletDetector and SecurityWebXmlDetector). In order to fix that, it was need to use inputstream that needed 2 requirements:
 1-Not closing stream just in case any used tool was doing it without asking for it (like Descriptors class).
 2-Enabling support for marking inputStream, as we need to "rewind" stream to the initial point so next webxml fraction detectors is chain could do their job.

Although the new inputstream with previous requirements (UncloseableBufferedInputStream) was added at the generic scanning method (we need the same inputstream so we can reset to the beginning of the stream), it shouldn't affect other detectors that doesn't use mark and reset. These features are used in SecurityWebXmlDetector, though, where we mark the inputstream each time, so we can reset the stream if it reached its end (stream.available ==0)"

juangon:
Said that, @bobmcwhirter @Ladicek it's true that changing the utility that parses web.xml from Shrinkwrap's Descriptors to other (custom or available third party library) would avoid the need of adding that UncloseableBufferedInputstream . If both of you want me to implement such a thing, just let me know. Just wanted to keep the same existing code if possible.

Ladicek:
Ah, sorry @juangon, I didn't read the commit message. Thanks.

Frankly, I still think if we need to read the content multiple times, then we shouldn't be using `InputStream` but something like `String`, `byte[]` or `Supplier<InputStream>`. But I just tried to understand the fraction autodetection code and failed (again), so I'm not trying to guess how much work that would be.

I don't think we should avoid `Descriptors`, no.

Ladicek:
So I took the time to go through the fraction detection code once more and this PR started to make sense to me. But with the uncloseable stuff -- we'll be leaking resources, no?

Here's a branch https://github.com/Ladicek/wildfly-swarm/commits/stop-passing-around-inputstreams where I replaced passing around `InputStream`s by passing around `InputStreamSupplier`s. Just for `Scanner`s -- the `FractionDetector`s don't need that. @juangon What do you think, would something like this work?

juangon:
Thanks @Ladicek ! I am checking it

bobmcwhirter:
@Ladicek -- added you as the reviewer.  Once you approve, you can merge (or I will).

juangon:
@bobmcwhirter I was able to check @Ladicek pull request with some slightly modification and it works, but I have the feeling that API is "asking us to move to other side". I am trying moving some parts to Scanner as I think it makes more sense. I did that yesterday but it isn't working properly, I am fixing that now. After fix I think each part will do the job it should be. Hope to finish it today or tomorrow...Sorry!

kenfinnigan:
Should this be closed in favor of the other PR?

juangon:
Yes @kenfinnigan. Closing this one, let's focus on https://github.com/wildfly-swarm/wildfly-swarm/pull/514

PR #519, opened 2017-05-30
SWARM-1369: User projects with different dependencies

kenfinnigan:

-----

Motivation
----------
User projects that depend on WF Swarm fractions can bring in different versions of dependent artifacts if WF has adjusted those versions.

Modifications
-------------
Add WF Core and WF parent POMs into WF Swarm BOMs to ensure correct versions.

Result
------
User projects now have correct dependent artifact versions, and removes duplicate versions in m2repo of uber jar where this occured.

### DIFF
--- a/boms/src/main/resources/bom-template.xml
+++ b/boms/src/main/resources/bom-template.xml
@@ -55,6 +55,20 @@
     <dependencies>
 #{dependencies}
       <dependency>
+        <groupId>org.wildfly.core</groupId>
+        <artifactId>wildfly-core-parent</artifactId>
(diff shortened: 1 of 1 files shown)


bobmcwhirter (approved):

Ladicek:
I was briefly thinking if we should possibly use the WildFly BOM instead of the parent POMs, but probably not. Those would only bring a subset of what we need.

kenfinnigan:
I had a look at the WF BOM before pushing, and there's very little of what we need in it unfortunately

PR #528, opened 2017-06-06
SWARM-1373: Print warning when mainClass property set

heiko-braun:
Motivation
----------
We are moving away from custom main()

Modifications
-------------
Print warning when property is still used during builds.

Result
------
No impact on functionality

### DIFF
--- a/plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/PackageTask.java
+++ b/plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/PackageTask.java
@@ -200,6 +200,15 @@ public class PackageTask extends DefaultTask {
             mainClassName = app.getMainClassName();
         }
 
+        if (mainClassName != null && !mainClassName.equals("")) {
+            getLogger().warn(
+                    "\n------\n" +
+                    "Custom main() usage is intended to be deprecated in a future release and is no longer supported, \n" +
+                    "please refer to http://reference.wildfly-swarm.io for YAML configuration that replaces it." +
+                    "\n------"
+            );
+        }
+
         return mainClassName;
     }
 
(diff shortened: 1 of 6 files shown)


Ladicek (changes requested):
`WildFly Swarm reference` instead of `WildflySwarm reference`, and it should probably print the URL: https://reference.wildfly-swarm.io/

heiko-braun:
@Ladicek The URL doesn't work in product builds.

Ladicek:
Hmm, not sure if we will have a different URL for product, but it's OK.

Ladicek (dismissed):

Ladicek:
Actually the typo is still there in the Gradle plugin, but that doesn't concern me _that_ much :-)

Ladicek (dismissed):

kenfinnigan:
As @Ladicek mentioned, this needs a link to the YAML reference guide, as described in the JIRA.

At this stage we don't know if there will be a product specific version of the reference guide. If there does end up being one, we can sort that out later.

Not sure it's necessary to put the calls into each Mojo, why not add the `deprecationWarnings` logic into constructor of `AbstractSwarmMojo`?

kenfinnigan:
Forgot to add, looks like the Maven Plugin testsuite needs to be modified to support the WARNING log now

Ladicek:
I'm fine with the WARNING log test being commented out for now -- I'd then think about what to do with the test. Maybe we don't have to test using `main` if we're deprecating it.

heiko-braun:
@kenfinnigan seems the field values are injected after the ctor, hence the separation.

kenfinnigan:
Thanks for checking

heiko-braun:
@Ladicek Do you suggest to remove the WITH_MAIN tests? Or simply remove the assertion on WARNING log messages?

Ladicek:
@heiko-braun I'd suggest commenting out the assertion on WARNING log messages for now. And I will file a JIRA for myself to get back to it -- maybe removing custom main tests would be best, but I'd rather think about it more.

heiko-braun:
Incorporated all changes and squashed into one

PR #529, opened 2017-06-06
SWARM-1327: Fix file leaks in tests and add new maven enforcer rule to check if a file exist by pattern to avoid same problem in future

juangon:
Motivation
----------

Modifications
-------------

Result
------

-----
Hi @bobmcwhirter @Ladicek. After adding these temporal files check, other unexpected errors appeared, like the one that needed a fix in RuntimeDeployer in order to remove all tmp files that were created in SwarmRepository.
Those projects that uses wildfly-swarm-plugin needs to change the file checking to a different phase ("stop" instead of "verify") so the Swarm plugin stops Swarm properly (and hence all resources are removed).

Hope it makes sense.

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
@@ -284,6 +284,10 @@ public class RuntimeDeployer implements Deployer {
     void stop() {
     }
 
+    public void removeAllContent() throws IOException {
+        this.contentRepository.removeAllContent();
+    }
+
     @Inject
     DeploymentContext deploymentContext;
 
--- a/core/container/src/main/java/org/wildfly/swarm/container/runtime/cdi/TempFileProviderProducer.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/runtime/cdi/TempFileProviderProducer.java
@@ -21,8 +21,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 
 import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
 import javax.enterprise.context.ApplicationScoped;
-import javax.enterprise.inject.Disposes;
 import javax.enterprise.inject.Produces;
 import javax.inject.Singleton;
 
@@ -61,14 +61,12 @@ public class TempFileProviderProducer {
         return this.tempFileProvider;
     }
 
-    void dispose(@Disposes TempFileProvider provider) {
-        // To ensure we only close the one we produce
-        if (this.tempFileProvider == provider) {
-            try {
-                this.tempFileProvider.close();
-            } catch (IOException e) {
-                SwarmMessages.MESSAGES.errorCleaningUpTempFileProvider(e);
-            }
+    @PreDestroy
+    void close() {
+        try {
+            this.tempFileProvider.close();
+        } catch (IOException e) {
+            SwarmMessages.MESSAGES.errorCleaningUpTempFileProvider(e);
         }
     }
 
--- a/plugins/maven-enforcer-pattern-size/src/main/java/org/wildfly/swarm/plugin/enforcer/patternsize/RequireFilePatternSize.java
+++ b/plugins/maven-enforcer-pattern-size/src/main/java/org/wildfly/swarm/plugin/enforcer/patternsize/RequireFilePatternSize.java
@@ -1,0 +1,161 @@
+/**
+ * Copyright 2015-2017 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.plugin.enforcer.patternsize;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.FileVisitOption;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.maven.enforcer.rule.api.EnforcerRule;
+import org.apache.maven.enforcer.rule.api.EnforcerRuleException;
+import org.apache.maven.enforcer.rule.api.EnforcerRuleHelper;
+
+/**
+ *
+ * @author Juan Gonzalez
+ *
+ */
+public class RequireFilePatternSize implements EnforcerRule {
+
+    private RequiredFilePattern[] filePatterns;
+
+    public RequiredFilePattern[] getRequiredFilePatterns() {
+        return filePatterns;
+    }
+
+    public void setRequiredFilePatterns(RequiredFilePattern[] filePatterns) {
+        this.filePatterns = filePatterns;
+    }
+
+    List<RequiredFilePatternFailure> checkFilePattern(final RequiredFilePattern requiredFilePattern) throws IOException {
+        if (requiredFilePattern == null || requiredFilePattern.getDirectory() == null) {
+            return null;
+        }
+
+        if (requiredFilePattern.getPattern() == null) {
+            return null;
+        }
+
+        final Pattern pattern = Pattern.compile(requiredFilePattern.getPattern());
+
+        String directory = requiredFilePattern.getDirectory();
+        File directoryFile = new File(directory);
+
+        final List<File> matchedFiles = new ArrayList<File>();
+
+        if (Files.isDirectory(directoryFile.toPath())) {
+            Files.walkFileTree(directoryFile.toPath(), EnumSet.of(FileVisitOption.FOLLOW_LINKS),
+                requiredFilePattern.isRecursive() ? Integer.MAX_VALUE : 1,
+                        new SimpleFileVisitor<Path>() {
+                        @Override
+                        public FileVisitResult visitFile(Path filePath, BasicFileAttributes attrs) throws IOException {
+                             Matcher matcher = pattern.matcher(filePath.toAbsolutePath().toString());
+                             if (matcher.matches()) {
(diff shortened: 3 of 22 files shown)


plugins/maven-enforcer-pattern-size/src/main/java/org/wildfly/swarm/plugin/enforcer/patternsize/RequireFilePatternSize.java
```diff
@@ -0,0 +1,160 @@
...
+import java.nio.file.FileVisitOption;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+import java.util.EnumSet;
+/*import java.util.ArrayList;
+import java.util.List;*/
```
Ladicek: Not really :-)

juangon: Ohhh sh*t. Thanks!

plugins/maven-enforcer-pattern-size/src/main/java/org/wildfly/swarm/plugin/enforcer/patternsize/RequiredFilePatternFailure.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+package org.wildfly.swarm.plugin.enforcer.patternsize;
+
+/**
+ *
+ * @author Juan Gonzalez
+ *
+ */
+public class RequiredFilePatternFailure extends RequiredFilePattern {
```
Ladicek, line 23: Should it really be a subclass? Maybe it should just contain a reference to `RequiredFilePattern`?

Ladicek:
Please avoid `e.printStackTrace();`. Also, what's the `stop` phase in the Maven lifecycle? Never heard of that.

plugins/maven-enforcer-pattern-size/src/main/java/org/wildfly/swarm/plugin/enforcer/patternsize/RequiredFilePatternFailure.java
juangon: Hmmm what advantages would be having a reference instead of subclassing in this case?

juangon:
Ok, as there is a list of files and I wouldn't like to throw for a single file before all files are being removed, do you suggest to just swallow it? 
"Stop" belongs to our own maven swarm plugin, when swarm is stopped, etc.

Ladicek:
There's `e.printStackTrace();` on multiple places. In the one situation where you delete files in the loop, well, maybe store the first exception thrown, swallow the rest, and throw the stored exception at the end (if there's one).

Re `stop` phase: I don't think Maven plugins can add phases to the lifecycle. The `stop` goal of our plugin is typically bound to `post-integration-test` phase. Which is _before_ `verify`, so I don't see why the temp file checking can't be done in the `verify` phase.

plugins/maven-enforcer-pattern-size/src/main/java/org/wildfly/swarm/plugin/enforcer/patternsize/RequiredFilePatternFailure.java
Ladicek, line 23: I'd personally ask the exact opposite question, but that's just my anti-inheritance nature :-) But maybe it's common for the Maven Enforcer plugin rules to do this, I don't know. So it was just a question.

juangon:
About the stop phase, I indeed made a stupid thing there and seems that it fails silently even providing a wrong phase. I am checking it. Thanks as always.

juangon:
I could fix all cases (so no more maven plugin in child pom.xml) except the case of testsuite-container. I am reviewing wildfly-core and I think the failure is there, as it doesn't remove a temporal directory. Still debugging...

Ladicek:
Could we add an exception? Fixing WF Core is a long run.

juangon:
Sure, I just want to ensure failure is there first. Fortunately we are creating our own SelfContainedContainer (the part I think the leak comes from), so we could override some methods and access the private File attribute by Reflection (same way as it's done for execute attribute too).

juangon:
So, seems the culprit is TempFileProvider when shutting down WeldContainer. I am investigating that and how to fix it.

juangon:
@bobmcwhirter @Ladicek I've updated the PR.

testsuite/testsuite-container/src/test/java/org/wildfly/swarm/container/test/DeploymentFailureTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.wildfly.swarm.spi.api.JARArchive;
+
+import static org.fest.assertions.Assertions.assertThat;
+import static org.fest.assertions.Fail.fail;
+
+public class DeploymentFailureTest {
+
+  @Test
+  public void testDeploymentFailure() throws Exception {
+      Swarm swarm = new Swarm();
```
juangon, line 16: Due to VFS TempFileProvider storing temporal file paths in static section, it doesn't goes well under same test class (TempFileProvider won't be unloaded). Separating in two different test classes does the trick.

core/container/src/main/java/org/wildfly/swarm/container/runtime/cdi/TempFileProviderProducer.java
```diff
@@ -61,14 +61,12 @@ TempFileProvider tempFileProvider() {
...
 
-    void dispose(@Disposes TempFileProvider provider) {
-        // To ensure we only close the one we produce
-        if (this.tempFileProvider == provider) {
-            try {
-                this.tempFileProvider.close();
-            } catch (IOException e) {
-                SwarmMessages.MESSAGES.errorCleaningUpTempFileProvider(e);
-            }
+    @PreDestroy
```
juangon, line 64: This was curious to find. If having a non-static Dispose method, then class is constructed again, causing to this class to initialize again before Weld being shutdown (so 2 initializes, 1 destroy), hence having two temporal files, only 1 is removed.
Adding PreDestroy makes unloading safer and working.

core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
```diff
@@ -234,6 +234,7 @@ public void deploy(Archive<?> deployment, String asName) throws DeploymentExcept
 
             if (BootstrapProperties.flagIsSet(SwarmProperties.EXPORT_DEPLOYMENT)) {
                 final File out = new File(deployment.getName());
+                out.deleteOnExit();
```
Ladicek: This looks wrong. It's the very purpose of `swarm.export.deployment` to dump the deployment so that it can be inspected manually. Deleting it on JVM exit defeats the purpose.

core/container/src/main/java/org/wildfly/swarm/container/runtime/cdi/TempFileProviderProducer.java
Ladicek: Nice!

plugins/maven-enforcer-pattern-size/src/main/java/org/wildfly/swarm/plugin/enforcer/patternsize/RequireFilePatternSize.java
```diff
@@ -0,0 +1,158 @@
...
+                                 if (matcher.matches()) {
+                                     File file = filePath.toFile();
+                                     matchedFiles.add(file);
+                                 }
+
+                                return super.visitFile(filePath, attrs);
+                            }
+                 });
+            } catch (IOException e) {
+                e.printStackTrace();
```
Ladicek: I know it's just a compile-time thing, but I'd rather avoid `e.printStackTrace` completely.

testsuite/testsuite-container/src/test/java/org/wildfly/swarm/container/test/DeploymentFailureTest.java
Ladicek: Where again is the problem? Sorry, I don't understand.

Ladicek:
Otherwise LGTM.

juangon:
Yeah @Ladicek, that deleteOnExit you said was the only one I changed without any specific test failure, so I will try to not touch too much next time.

Just updated the PR with all your recommendations.

Thanks very much!

testsuite/testsuite-jsp/pom.xml
```diff
@@ -70,6 +70,20 @@
...
+              <execution>
+                <id>enforce</id>
+                <phase>stop</phase>
+                <configuration combine.self="merge"/>
+                <goals>
+                   <goal>enforce</goal>
+                    </goals>
+              </execution>
+            </executions>
+          </plugin>
```
Ladicek: This got added again by accident?

Ladicek:
Just one issue, otherwise LGTM.

juangon:
Thanks again  @Ladicek! Actually I forgot to remove this change. I've updated the branch (and PR).

plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StartMojo.java
```diff
@@ -328,7 +329,7 @@ protected SwarmExecutor executor(final Path appPath, final String name,
                         Files.createTempFile(Paths.get(this.projectBuildDir), "swarm-", "-cp.txt").toFile() :
                         Files.createTempFile("swarm-", "-cp.txt").toFile();
 
-                tmp.deleteOnExit();
+                getPluginContext().put("swarm-cp-file", tmp);
```
Ladicek: Actually I'm thinking here... this is OK for the `start` goal, because there should always be a corresponding `stop`, but what about `run`? `RunMojo extends StartMojo`, so if I do `mvn wildlfy-swarm:run` and then Ctrl+C that process, the temp file will stay on the filesystem. Right?

juangon: Awesome @Ladicek. I didn't take that into account. Fixed and (again) updated pull.

Ladicek:
LGTM, but see the failure on Windows.

juangon:
Oh, I thought it was about some of those problems in CI Windows I heard these days. But reading more in console, it's failing as some arquilian temp files exists there. Looking at that now...

juangon:
@bobmcwhirter @Ladicek seems (some) temporary files in Windows aren't removed. After some tries and failures in CI I am gonna install windows and try to reproduce locally. Please hold on this PR until then.

kenfinnigan:
Bear in mind we do have known issues on Windows due to the way processes end that means temp files can be left behind.

Not sure if that's the case in this situation, @bobmcwhirter might know

juangon:
Exactly @kenfinnigan. That's what I am checking and trying to have a better way to finish that process. Thanks!

juangon:
Sent other PR here: https://github.com/wildfly-swarm/wildfly-swarm/pull/559
Closing this one.

PR #534, opened 2017-06-08
SWARM-1392: Create a separate bom to manage versions in user project builds

heiko-braun:
Simple preview to discuss further steps. Not intended to be merged yet

### DIFF
--- a/boms/bom-dependency/pom.xml
+++ b/boms/bom-dependency/pom.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2015 Red Hat, Inc. and/or its affiliates.
+  ~
+  ~ Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+  -->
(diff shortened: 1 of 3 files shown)


heiko-braun:
@kenfinnigan Is this similar to what you had in mind?

kenfinnigan:
Yes, with a couple of changes. I think `dependency-bom` is a more appropriate name.

Also, I wasn't expecting http components to be in there, I was only expecting openjdk-orb and whatever else was causing the issues before

heiko-braun:
I think we would need all of the pending DR3 ones to managed. httpclient is one of them I used to test the setup.

kenfinnigan:
I didn't think there was a duplicate version issue with httpclient?

heiko-braun:
not problem with duplicates, but the version was not aligned with a prod one. see the DR3 tab in @mswiderski report

kenfinnigan:
I have no idea which report that is, can you share a link via email?

kenfinnigan:
Thanks for the changes.

If we can adjust name to `dependency-bom` I think we're good to merge, when it passes

heiko-braun:
retest this please

PR #538, opened 2017-06-12
SWARM-1381: add test for using the RESTEasy Client

heiko-braun:
Motivation
----------
Some dependency handling changes that went into Swarm 2017.6.0
broke the RESTEasy Client. A test that reproduces this is needed.

Modifications
-------------
Added `testsuite-jaxrs-client` which uses the JAX-RS Client,
provided by RESTEasy, to access a simple JAX-RS resource.

Result
------
More test coverage, no change to product.

### DIFF
--- a/testsuite/testsuite-jaxrs-client/pom.xml
+++ b/testsuite/testsuite-jaxrs-client/pom.xml
@@ -1,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2017 Red Hat, Inc. and/or its affiliates.
+  ~
+  ~ Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.wildfly.swarm</groupId>
+    <artifactId>testsuite</artifactId>
+    <version>2017.7.0-SNAPSHOT</version>
+    <relativePath>../</relativePath>
+  </parent>
+
+  <groupId>org.wildfly.swarm</groupId>
+  <artifactId>testsuite-jaxrs-client</artifactId>
+
+  <name>Test Suite: JAX-RS Client</name>
+  <description>Test Suite: JAX-RS Client</description>
+
+  <packaging>war</packaging>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.wildfly.swarm</groupId>
+      <artifactId>jaxrs</artifactId>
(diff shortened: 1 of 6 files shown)


heiko-braun:
This PR changes the parent inheritance of the jaxrs-client testsuite. Previously the dependency on the wildfly-swarm parent did also lead to an inherited wildfly-swarm feature pack, which brought back the exclusion policies on the rest-easy client and thus negated the effect the fix. This version decouples the test from the wildfly-swarm parent and depends on the bom inheritance, which brings it closer to the setup actual user projects would have.

Ladicek:
I don't see the difference in the PR, but this is a very good point. Maybe the `testsuite` module shouldn't even inherit from the main Swarm parent POM... but that would require significant reorganization of the Maven project structure I think. Hmm...

heiko-braun:
@Ladicek Sorry, one commit went missing. See the latest addition for the relevant changes.

Ladicek:
So this is way above my Maven skills. The root Swarm POM includes `testsuite/testsuite-jaxrs-client` as a `<module>`, but the module refers to a different parent POM -- is that OK? Sounds like I will, once again, start thinking about a separate testsuite project...

Ladicek:
BTW if the `testsuite-jaxrs-client` project no longer inherits from `testsuite`, then the entire specification of dependency on `graphene-webdriver` for the purpose of excluding everything can be removed, as that's inherited from `testsuite`.

heiko-braun:
@Ladicek Yes, that's the part I am not sure myself. I think it's fine, but maybe @kenfinnigan has a different opinion?

heiko-braun:
I think maven projects can only have one ultimate parent,  but a different sub module hierarchy. In this case the setup is as follows:

```
org.wildfly.swarm:parent
  |- org.wildfly.swarm:wildfly-swarm
      |- testsuite
  |- jaxrs-client-test
```
Everything points to `org.wildfly.swarm:parent`, but the jaxrs-client-test skips two elements in the parent hierarchy.

Ladicek:
Yes, I was thinking maybe the entire test suite should inherit from `org.wildfly.swarm:parent`, to avoid inheriting too much from `org.wildfly.swarm:wildfly-swarm`.

heiko-braun:
@bobmcwhirter @kenfinnigan are you up for reviewing the structural changes to the maven parent hierarchy?

testsuite/testsuite-jaxrs-client/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-client</artifactId>
+      <version>3.0.19.Final</version>
+      <scope>test</scope>
+    </dependency>
+
+    <!--
```
kenfinnigan: Based on @Ladicek's comments can we remove this section and verify it passes on CI?

kenfinnigan (changes requested):

heiko-braun:
@kenfinnigan done

kenfinnigan (approved):

kenfinnigan:
Thanks @heiko-braun, when CI is green we can merge

PR #542, opened 2017-06-14
SWARM-1398 - Added OpenTracing fraction

jpkrohling:
Motivation
----------
Adds OpenTracing to a deployment setting the GlobalTracer to a Tracer
resolved via the TraceResolver. This means that it gets concrete Tracer
implementation from the application being deployed.

Modifications
-------------
A new fraction was added, under fractions/opentracing.

Result
------
A new fraction is available, under the GA org.wildfly.swarm:opentracing

### DIFF
--- a/fractions/opentracing/src/main/java/org/wildfly/swarm/opentracing/OpenTracingFraction.java
+++ b/fractions/opentracing/src/main/java/org/wildfly/swarm/opentracing/OpenTracingFraction.java
@@ -1,0 +1,11 @@
+package org.wildfly.swarm.opentracing;
+
+import org.wildfly.swarm.spi.api.Fraction;
+import org.wildfly.swarm.spi.api.annotations.DeploymentModule;
+
+/**
+ * @author Juraci Paixão Kröhling
+ */
+@DeploymentModule(name = "org.wildfly.swarm.opentracing", slot = "deployment")
+public class OpenTracingFraction implements Fraction<OpenTracingFraction> {
+}
--- a/fractions/opentracing/src/main/java/org/wildfly/swarm/opentracing/runtime/OpenTracingInstaller.java
+++ b/fractions/opentracing/src/main/java/org/wildfly/swarm/opentracing/runtime/OpenTracingInstaller.java
@@ -1,0 +1,37 @@
+package org.wildfly.swarm.opentracing.runtime;
+
+import org.jboss.logging.Logger;
+import org.jboss.shrinkwrap.api.Archive;
+import org.wildfly.swarm.spi.api.DeploymentProcessor;
+import org.wildfly.swarm.spi.runtime.annotations.DeploymentScoped;
+import org.wildfly.swarm.undertow.WARArchive;
+import org.wildfly.swarm.undertow.descriptors.WebXmlAsset;
+
+import javax.inject.Inject;
+
+/**
+ * @author Juraci Paixão Kröhling
+ */
+@DeploymentScoped
+public class OpenTracingInstaller implements DeploymentProcessor {
+    private static final Logger logger = Logger.getLogger(OpenTracingInstaller.class);
+    private final Archive<?> archive;
+
+    @Inject
+    public OpenTracingInstaller(Archive archive) {
+        this.archive = archive;
+    }
+
+    @Override
+    public void process() throws Exception {
+        logger.info("Determining whether to install OpenTracing integration or not.");
+        if (archive.getName().endsWith(".war")) {
+            logger.logf(Logger.Level.INFO, "Installing the OpenTracing integration for the deployment %s", archive.getName());
+            WARArchive webArchive = archive.as(WARArchive.class);
+            WebXmlAsset webXml = webArchive.findWebXmlAsset();
+
+            logger.logf(Logger.Level.INFO, "Adding the listener org.wildfly.swarm.opentracing.deployment.OpenTracingInitializer");
+            webXml.addListener("org.wildfly.swarm.opentracing.deployment.OpenTracingInitializer");
+        }
+    }
+}
--- a/fractions/opentracing/src/main/resources/modules/io/opentracing/main/module.xml
+++ b/fractions/opentracing/src/main/resources/modules/io/opentracing/main/module.xml
@@ -1,0 +1,14 @@
+<module xmlns="urn:jboss:module:1.3" name="io.opentracing">
+  <resources>
+    <artifact name="io.opentracing:opentracing-api:${version.opentracing}"/>
+    <artifact name="io.opentracing:opentracing-util:${version.opentracing}"/>
+    <artifact name="io.opentracing:opentracing-noop:${version.opentracing}"/>
+    <artifact name="io.opentracing.contrib:opentracing-globaltracer:${version.globaltracer}"/>
+    <artifact name="io.opentracing.contrib:opentracing-tracerresolver:${version.tracerresolver}"/>
+  </resources>
+
+  <dependencies>
+    <module name="javax.api"/>
+    <module name="javax.servlet.api"/>
+  </dependencies>
+</module>
--- a/pom.xml
+++ b/pom.xml
@@ -1349,7 +1349,6 @@
     <module>fractions/keycloak</module>
     <module>fractions/microprofile</module>
     <module>fractions/monitor</module>
-    <module>fractions/opentracing</module>
     <module>fractions/topology</module>
     <module>fractions/topology-openshift</module>
     <module>fractions/topology-webapp</module>
@@ -1449,6 +1448,7 @@
         <module>fractions/jolokia</module>
         <module>fractions/logstash</module>
         <module>fractions/keycloak-server</module>
+        <module>fractions/opentracing</module>
         <module>fractions/spring</module>
         <module>fractions/swagger</module>
         <module>fractions/swagger-webapp</module>
--- a/pom.xml
+++ b/pom.xml
@@ -1349,6 +1349,7 @@
     <module>fractions/keycloak</module>
     <module>fractions/microprofile</module>
     <module>fractions/monitor</module>
+    <module>fractions/opentracing</module>
     <module>fractions/topology</module>
     <module>fractions/topology-openshift</module>
     <module>fractions/topology-webapp</module>
--- a/pom.xml
+++ b/pom.xml
@@ -100,6 +100,11 @@
     <version.zipkin.reporter>0.6.1</version.zipkin.reporter>
     <version.zipkin>1.13.1</version.zipkin>
 
+    <!-- OpenTracing related versions -->
+    <version.opentracing>0.22.0</version.opentracing>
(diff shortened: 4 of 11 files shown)


jpkrohling:
This is my first fraction, so, I'm still a bit confused on the dependencies part. I tried to remove everything non-essential, but there might still be something that could be removed.

In any case, this is a base fraction for future further integration. A sample application is available here:
https://github.com/jpkrohling/swarm-servlet/tree/JPK-AddedReadmeAndSwitchedToJaeger . This sample application can later be added to the Swarm examples, if desired.

Before merging, I would like to agree on an approach for further work (so that the review is done with the agreed scenario in mind).

OpenTracing provides "framework integration" libraries for Servlet, JAX-RS, Spring and so on. What would be the best way to add this? I can think of three ways. Using JAX-RS as an example (OT framework integration: https://github.com/opentracing-contrib/java-jaxrs)

1) Let the application developer add the framework integration library to their pom.xml , as it's done outside of Swarm (that's how the sample application is doing at the moment)
2) Create new fractions, like, `opentracing-jaxrs`. This would require `jaxrs` and `opentracing`, and would pull the OT JAX-RS framework integration library.
3) Add to the existing JAX-RS fraction. This would require JAX-RS to depend on OpenTracing, which *should not* cause any harm (as OT ships with a NoopTracer by default, so, nothing would break).

cc @objectiser , @pavolloffay

pom.xml
```diff
@@ -1352,6 +1352,7 @@
     <module>fractions/keycloak</module>
     <module>fractions/microprofile</module>
     <module>fractions/monitor</module>
+    <module>fractions/opentracing</module>
```
Ladicek: This module should probably be in the `unsupported` profile.

fractions/opentracing/src/main/java/org/wildfly/swarm/opentracing/OpenTracingFraction.java
```diff
@@ -0,0 +1,11 @@
+package org.wildfly.swarm.opentracing;
+
+import org.wildfly.swarm.spi.api.Fraction;
+import org.wildfly.swarm.spi.api.annotations.DeploymentModule;
+
+/**
+ * @author Juraci Paixão Kröhling
+ */
+@DeploymentModule(name = "org.wildfly.swarm.opentracing", slot = "deployment")
+public class OpenTracingFraction implements Fraction<OpenTracingFraction> {
```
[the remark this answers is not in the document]
kenfinnigan: Yes it's needed, that's how our plugin works

fractions/opentracing/src/main/java/org/wildfly/swarm/opentracing/runtime/OpenTracingInstaller.java
```diff
@@ -0,0 +1,37 @@
+package org.wildfly.swarm.opentracing.runtime;
```
kenfinnigan, line 1: I don't believe this class is needed at all?

The ServletContextListener being in `deployment` package and attaching that module to users archive with `@DeploymentModule` should be enough

fractions/opentracing/src/main/resources/modules/io/opentracing/main/module.xml
```diff
@@ -0,0 +1,14 @@
+<module xmlns="urn:jboss:module:1.3" name="io.opentracing">
+  <resources>
```
kenfinnigan, line 2: Are all these APIs only? ie. the user still needs to bring an impl in their project dependencies?

kenfinnigan:
Unless it's overly cumbersome, my preference would be option 1 as that allows the developer to choose a more recent impl version, provided there's API compatibility.

Option 2 would then require some auto detection for each to be added to the users application as appropriate, which might be tricky.

kenfinnigan:
Forgot to mention that it would be great if it was possible to add a /testsuite module to verify it's working as expected on each build

fractions/opentracing/src/main/resources/modules/io/opentracing/main/module.xml
jpkrohling: Right, the user needs to bring a concrete implementation for the Tracer to have something useful, but this won't break in absence of a Tracer, as a `NoopTracer` is provided by the API.

fractions/opentracing/src/main/java/org/wildfly/swarm/opentracing/runtime/OpenTracingInstaller.java
jpkrohling: Will try that out, thanks!

jpkrohling:
> Unless it's overly cumbersome, my preference would be option 1 as that allows the developer to choose a more recent impl version, provided there's API compatibility.

I'm also tending towards option 1. But just to play with ideas: a counter argument could be that Swarm comes with "batteries included", and by plugging a concrete Tracer implementation means that developers get automatic tracing capabilities to all frameworks and libraries (Servlet, JAX-RS, EJBs, ...) without adding anything extra to their pom files.

Option 2 as I imagined would be more like giving the developer the choice of adding either `org.wildfly.swarm:jaxrs` to their pom, or `org.wildfly.swarm:opentracing-jaxrs`.

kenfinnigan:
Another approach would be a combination.

So there's an `opentracing`, as now, that is purely the API side of the fence.

Then we have `opentracing-jaeger`, or whatever, that builds on `opentracing` by providing a specific impl.

I like the idea of having `opentracing` in addition in that scenario because it means if the jaeger impl in WF Swarm gets out of date users have the option of choosing a newer version as the impl directly

pavolloffay:
@jpkrohling this is a much better than `opentracing-hawkular`. I have one question does annotation `@Traced` work? https://github.com/opentracing-contrib/java-jaxrs/blob/master/opentracing-jaxrs2/src/main/java/io/opentracing/contrib/jaxrs2/server/Traced.java

jpkrohling:
> I have one question does annotation @Traced work? 

No. It could, if we decided to ship the framework integration libraries directly with this fraction, and that was the point for my question in https://github.com/wildfly-swarm/wildfly-swarm/pull/542#issuecomment-308455240 . 

In the end, it depends on whether Swarm prefers to have a "batteries included" approach, or a "here's the scaffold, do it your way" approach.

jpkrohling:
@kenfinnigan note that there are two main parts to OpenTracing: the first is about gathering data in the form of spans, the second is about sending this data "somewhere". The second is the so called "tracer", and where Jaeger/Zipkin/... fits and I agree that a natural next step would be to add a `opentracing-jaeger`, extending the fraction from this PR.

Just to check if we are on the same page regarding the first part: spans are gathered either manually (in case of business spans, like "add new book to inventory"), or automatically, via framework integration libraries (like, "process HTTP request", in case of Servlet integration).

The question on https://github.com/wildfly-swarm/wildfly-swarm/pull/542#issuecomment-308455240 is about the span gathering case. So, we could ship all possible framework integration libraries on this `opentracing` fraction (Servlet, JAX-RS, Spring, ...), or have fractions specific to the chosen stack (`opentracing-jaxrs`), or add OpenTracing support to the existing "stack fractions" (jaxrs), or let the user add the libraries they care about (by telling them to add `opentracing-contrib/java-jaxrs` to their `pom.xml`).

fractions/opentracing/src/main/java/org/wildfly/swarm/opentracing/runtime/OpenTracingInstaller.java
jpkrohling: It might be needed after all. I tried removing this `Installer` class, but the context listener is never called. I also tried removing the `@Vetoed` from the listener, as well as adding a `beans.xml` to `META-INF`. Perhaps I'm missing something ? 

https://gist.github.com/jpkrohling/789eadd3c39b6c82123ec552ac825772

jpkrohling:
I just rebased onto the latest master, moved the OpenTracing module to unsupported and added a couple of very basic tests. All they are doing so far is making sure the `GlobalTracer` is setting the tracer based on the outcome of the `TracerResolver`. Depending on the outcome of https://github.com/wildfly-swarm/wildfly-swarm/pull/542#issuecomment-308455240 , I may add more tests ensuring that spans are created for specific operations (HTTP request, for instance).

kenfinnigan:
Ok, I think I understand this better now.

I think my preference would be `opentracing-jaxrs`, `opentracing-servlet`, etc.

Then each could have auto detection such that `jaxrs` and `opentracing` API presence leads to `opentracing-jaxrs` being added ,etc.

jpkrohling:
Once this passes the review, I'll squash the commits and update the PR.

kenfinnigan (approved):

kenfinnigan:
ok to test

testsuite/testsuite-opentracing/src/test/java/org/wildfly/swarm/opentracing/OpenTracingWithTraceResolverOnDeploymentTest.java
```diff
@@ -0,0 +1,64 @@
...
+        deployment.addPackage(MockTracer.class.getPackage());
+        deployment.addAsServiceProvider(TracerResolver.class, MockTracerResolver.class);
+
+        return deployment;
+    }
+
+    @CreateSwarm
+    public static Swarm newContainer() throws Exception {
+        return new Swarm().fraction(new OpenTracingFraction());
+    }
```
Ladicek, line 58: Could this test just use `@DefaultDeployment`? It seems neither `@Deployment` nor `@CreateSwarm` are necessary, but maybe I'm missing something.

kenfinnigan: Probably could yes.

@jpkrohling please raise a new PR to fix

wyaeld:
@kenfinnigan @jpkrohling 

Have been trying in vain to get a non-trivial opentracing setup with wildfly-swarm to work.
Specificically when wanting to trace from the request through to the JDBC layer.
The community has contributed.
https://github.com/opentracing-contrib/java-jaxrs
https://github.com/opentracing-contrib/java-jdbc
The former works fine, and I've seen you mention it above, however the latter has to do some work wrapping database drivers, and after 2 days I've been able to get that to work with swarm's datasource layer.

Someone more experienced with how the datasource fractions work might be able to spot the issue.  I believe being able to show having both the jaxrs and jdbc tracing actually work is useful for swarm

jpkrohling:
@wyaeld , thanks for bringing this up. I've just added a Trello card for testing this: https://trello.com/c/mY9BP4I1/40-test-wildfly-swarm-jdbc-with-opentracing-fraction and I'll be looking into this soon.

If you are facing a specific error, or if you have a sample application with your setup, it would help me a lot in getting started on this task.

jpkrohling:
Opened: https://issues.jboss.org/browse/SWARM-1461 , see also https://github.com/jpkrohling/demo-swarm-jdbc-opentracing

PR #547, opened 2017-06-20
SWARM-1305 add wildfly-nosql integration + NoSQL fractions

scottmarlow:

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/util/Messages.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/util/Messages.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2017 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.swarm.container.util;
+
+import org.jboss.logging.BasicLogger;
+import org.jboss.logging.Logger;
+import org.jboss.logging.annotations.Cause;
+import org.jboss.logging.annotations.LogMessage;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageLogger;
+
+/**
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+@MessageLogger(projectCode = "WFSWARM", length = 4)
+public interface Messages extends BasicLogger {
+
+    Messages MESSAGES = Logger.getMessageLogger(Messages.class, "org.wildfly.swarm");
+
+    int OFFSET = 1120;
+
+    @LogMessage(level = Logger.Level.DEBUG)
+    @Message(id = OFFSET + 1, value = "Attempting to auto-detect driver for %s")
+    void attemptToAutoDetectDriver(String driverName);
+
+    @LogMessage(level = Logger.Level.INFO)
+    @Message(id = OFFSET + 2, value = "Auto-detected driver for %s")
+    void autodetectedDriver(String driverName);
+
+    @LogMessage(level = Logger.Level.ERROR)
+    @Message(id = OFFSET + 3, value = "Detected driver %s, but an error occurred while loading it")
+    void errorLoadingAutodetectedDriver(String driverName, @Cause Throwable t);
+
+    @Message(id = OFFSET + 4, value = "Cannot generate new module '%s' " +
+                "when module '%s' is already being used by application")
+    IllegalStateException cannotAddReferenceToModule(String module, String alreadyAssociatedModule);
+
+}
--- a/fractions/nosql/cassandra/src/main/java/org/wildfly/swarm/cassandra/runtime/CassandraCustomizer.java
+++ b/fractions/nosql/cassandra/src/main/java/org/wildfly/swarm/cassandra/runtime/CassandraCustomizer.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2017 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.swarm.cassandra.runtime;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Any;
+import javax.enterprise.inject.Instance;
+import javax.inject.Inject;
+
+import org.wildfly.swarm.cassandra.CassandraFraction;
+import org.wildfly.swarm.spi.api.Customizer;
+import org.wildfly.swarm.container.util.Messages;
+import org.wildfly.swarm.spi.runtime.annotations.Pre;
+
+/**
+ * CassandraCustomizer
+ *
+ * @author Scott Marlow
+ */
+@Pre
+@ApplicationScoped
+public class CassandraCustomizer implements Customizer {
+
+    @Inject
+    @Any
+    Instance<CassandraFraction> allDrivers;
+
+    @Override
+    public void customize() throws Exception {
+        customizeDrivers();
+    }
+
+    protected void customizeDrivers() {
+            this.allDrivers.forEach(this::attemptInstallation);
+        }
+
+    protected void attemptInstallation(CassandraFraction info) {
+        CassandraDriverInfo cassandraDriverInfo = new CassandraDriverInfo();
+        if (cassandraDriverInfo.detect(info)) {
+            Messages.MESSAGES.autodetectedDriver(cassandraDriverInfo.name());
+        }
+    }
+}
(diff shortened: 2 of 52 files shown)


kenfinnigan:
add to whitelist

scottmarlow:
looks like I missed a reference to org.wildfly.nosql:wildfly-cassandra-feature-pack:zip:1.0.0.Alpha1-SNAPSHOT

scottmarlow:
looks like the -Pnosql profile in the root pom.xml is not correct yet, as I didn't expect the NoSQL tests to run on CI but they did (which fail, since we don't yet have NoSQL DBs via Kube or Docker).

fractions/nosql/cassandra/src/main/java/org/wildfly/swarm/cassandra/runtime/CassandraCustomizer.java
```diff
@@ -0,0 +1,58 @@
...
+    public void customize() throws Exception {
+        customizeDrivers();
+    }
+
+    protected void customizeDrivers() {
+            this.allDrivers.forEach(this::attemptInstallation);
+        }
+
+    protected void attemptInstallation(CassandraFraction info) {
+        CassandraDriverInfo mongoDBDriverInfo = new CassandraDriverInfo();
```
Ladicek: `mongoDBDriverInfo` :-)

pom.xml
```diff
@@ -1360,6 +1508,11 @@
     <module>fractions/keycloak</module>
     <module>fractions/microprofile</module>
     <module>fractions/monitor</module>
+    <module>fractions/netflix</module>
+    <module>fractions/nosql/cassandra</module>
+    <module>fractions/nosql/mongodb</module>
+    <module>fractions/nosql/neo4j</module>
+    <module>fractions/nosql/orientdb</module>
```
Ladicek: These should probably be in the `unsupported` profile?

testsuite/testsuite-cassandra/src/test/java/org/wildfly/swarm/cassandra/CassandraArquillianTest.java
```diff
@@ -0,0 +1,146 @@
...
+                            .outboundSocketBindingRef("casstesthost")
+                        )
+                        .jndiName("java:jboss/cassandradriver/test")
+                        .id("cassandratestprofile")
+                        .module("org.cassandra.custom")
+                        .securityDomain("cassandraRealm")
+                        .ssl(false)
+
+                )
+        );
```
Ladicek: How hard would it be to use `project-defaults.yml` instead of custom `main`? YAML is the preferred way of configuring Swarm.

testsuite/testsuite-cassandra/src/test/java/org/wildfly/swarm/cassandra/CassandraArquillianTest.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Test
+    public void testNothing() throws InterruptedException, TimeoutException {
+    }
```
Ladicek: ? :-)

Ladicek:
A _lot_ of code duplication in the fractions.

pom.xml
scottmarlow: What is the purpose of the unsupported profile?  Is that for experimental fractions, like the NoSQL fractions?  Or is that more of a way to indicate that the fractions are not expected to be in product yet?

Easy enough for me to move them to the unsupported profile.  I would like to develop community around the NoSQL fractions and see them move to be supported at some point. :)

I'll remove the extra netflix module as well.

fractions/nosql/cassandra/src/main/java/org/wildfly/swarm/cassandra/runtime/CassandraCustomizer.java
scottmarlow: nice catch :-)

pom.xml
kenfinnigan: Right, basically things that aren't in product

testsuite/testsuite-cassandra/src/test/java/org/wildfly/swarm/cassandra/CassandraArquillianTest.java
scottmarlow: Excellent idea, I have no idea but that sounds like a good change for me to try.

testsuite/testsuite-cassandra/src/test/java/org/wildfly/swarm/cassandra/CassandraArquillianTest.java
scottmarlow: awesome find, I added that initially and later forgot to remove it. :-)  Just removed it.

scottmarlow:
There is very little code in the fractions, most of the code is generated in the nosql-config-api classes, which the fractions extend.  I'm not sure that I understand what you mean about there being a lot of code duplication in the fractions.

There is also more code in the underlying NoSQL subsystems (http://github.com/wildfly/wildfly-nosql/).

testsuite/testsuite-cassandra/src/test/java/org/wildfly/swarm/cassandra/CassandraArquillianTest.java
bobmcwhirter: Since you config-api, then you should get a *lot* of configurable-via-YAML for free.

Run your stuff with -Dswarm.logging.org.wildfly.swarm.config=DEBUG and you'll see the dotted-notation version of things that can go into YAML.

To translate into YAML, view the dotted-notation as a tree.

`swarm.foo.bar=whatever`

becomes

    swarm:
      foo:
        bar: whatever

scottmarlow: Do we have anything built in for OutboundSocketBinding?  
https://paste.fedoraproject.org/paste/QUGYj04UHSI9X3sXEOG3IA shows the logged configuration for the MongoDBArquillianTest.  MongoDBFraction + SecurityFraction seem to included in the logged configuration but not OutboundSocketBinding.  

Do we need to add a Customizer, similar to RemoteConnectionSocketBindingCustomizer or SocketBindingCustomizer, to get configurable-via-YAML for OutboundSocketBinding?

Thanks,
Scott

Ladicek:
The `*Fraction` and `*Customizer` classes probably can't be shared (and are fairly small), but the `*DriverInfo` and `Messages` classes look all almost the same and a lot of code could IMHO be shared there.

scottmarlow:
Updated the *DriverInfo + Messages classes to use a common org.wildfly.swarm.container.util.DriverModuleBuilder class that is shared.  Perhaps datasource DriverInfo could use DriverModuleBuilder as well, at some point.

scottmarlow:
Still need to update the tests to use project-defaults.yml (after figuring out the OutboundSocketBinding part)

scottmarlow:
For the project-defaults.yml, I'm not yet sure of the best way to handle the OutboundSocketBinding part.  In theory, we could have a single NoSQL server configuration option that helps our NoSQL unit tests to use a projects-defaults.yml (e.g. similar to EnhancedSMTPServer supporting "swarm.mail.smtp.host" + "swarm.mail.smtp.port"), however, for multiple NoSQL server host/port pairs, we would need more than that, to support projects-defaults.yml for NoSQL.

What kind of Swarm class should I try to use, to solve adding multiple host/port pairs via OutboundSocketBinding in projects-defaults.yml?  Would a OutboundSocketBinding fraction help with that?  Or something else?

scottmarlow:
I'm not sure that a fraction would work, since the outbound socket binding in WildFly doesn't appear to be represented as a subsystem, instead the DMR representation is available via something like (from jboss-cli):

  ls /socket-binding-group=standard-sockets/remote-destination-outbound-socket-binding
  casstesthost    mail-smtp       mongotesthost   neo4jtesthost   orienttesthost

  ls /socket-binding-group=standard-sockets/remote-destination-outbound-socket-binding=mongotesthost
  fixed-source-port=false     host=localhost   port=27017   source-interface=undefined  source-port=undefined

bobmcwhirter:
I think a generic OutboundSocketBindingCustomizer in :container might work, fed by some @Configurable Map<> or information maybe.  We have a fair bit of plumbing in place, but just maybe lacking the YAML to SocketBinding bits.

e.g., we already have a datastructure for them and the capability of marshalling them:

https://github.com/wildfly-swarm/wildfly-swarm/blob/master/core/container/src/main/java/org/wildfly/swarm/container/runtime/marshal/SocketBindingGroupMarshaller.java

Just currently no real way to create new instances from the YAML as CDI-produced beans.

scottmarlow:
I should squash these commits into one after they are reviewed :)

scottmarlow:
need to remove unused imports from the tests.

scottmarlow:
Unused imports were removed from tests. :)

scottmarlow:
Note to self, remember to update [wildfly-nosql doc](https://github.com/wildfly/wildfly-nosql/blob/master/doc/README.adoc#connecting-to-neo4j-from-wildfly-swarm) to remove main classes, instead link to YAML files from wildfly-swarm unit tests.

core/container/src/main/java/org/wildfly/swarm/container/util/Messages.java
```diff
@@ -0,0 +1,52 @@
...
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+@MessageLogger(projectCode = "WFSWARM", length = 4)
+public interface Messages extends BasicLogger {
+
+    Messages MESSAGES = Logger.getMessageLogger(Messages.class, "org.wildfly.swarm");
+
+    int OFFSET = 1120;
+
+    @LogMessage(level = Logger.Level.DEBUG)
```
bobmcwhirter, line 36: Not blocking this merge, but I think moving these to a fraction-specific logging class with WFSMONGO or WFSNOSQL or whatever makes sense would be good.  We have a few other `Messages` classes scattered.

Consider it a refactoring, not a blocking comment.

bobmcwhirter (approved):

core/container/src/main/java/org/wildfly/swarm/container/util/Messages.java
scottmarlow: If we move the Messages to fractions/nosql/somethingCommon we probably could also move org.wildfly.swarm.container.util.DriverModuleBuilder there as well, unless we want to share DriverModuleBuilder for "bring your own JDBC drivers"?

PR #554, opened 2017-06-26
Added support for the deployment scanner subsystem

heiko-braun:
SWARM-1415:  Add support for the deployment scanner subsystem

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1349,7 +1349,6 @@
     <module>fractions/javaee/transactions</module>
     <module>fractions/javaee/undertow</module>
 
-    <module>fractions/wildfly/scanner</module>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -1349,6 +1349,7 @@
     <module>fractions/javaee/transactions</module>
     <module>fractions/javaee/undertow</module>
 
+    <module>fractions/wildfly/scanner</module>
```
Ladicek: Should be in the `unsupported` profile?

heiko-braun:
Thanks @Ladicek . I've moved it into the unsupported profile

Ladicek (approved):

PR #555, opened 2017-06-26
SWARM-1418 - De-manage MSC dependency.

bobmcwhirter:
Motivation
----------
By explicitly managing our dependency and module.xml for MSC,
we introduce an opportunity for version skew when going to
product.

Modifications
-------------
Remove and adjust our modules and versions related to JBoss MSC.

Result
------
Fewer things to skew about.

-----


Ladicek (approved):

Ladicek:
Though I'm not sure about removing the `module.xml` for `javax.sql.api` -- is that totally unnecessary?

bobmcwhirter:
Well, I figured while I was in there, I could see how much we could delete, particularly if it's just replicated from the WildFly/EAP feature-pack and module.xml's.

Seems like javax.sql.api and javax.api all could go away from our bootstrap module.  Might as well, to prevent possible future skew, even if not a problem immediately.

So, I removed those module.xml's and excluded them from the module-finders that were referencing them, I believe.

bobmcwhirter:
Yah, core builds, and -examples pass with both uberjar and non-uberjar with this PR.

So I think it's safe/good/right to remove all that we can.

Ladicek:
Ah, right, obviously we get the `module.xml` for `javax.sql.api` from WildFly. It's just that it isn't related to MSC what triggered my question.

bobmcwhirter:
Yeah, my commit message is sub-par. :)

PR #556, opened 2017-06-26
SWARM-1411 - Added Jaeger Fraction

jpkrohling:
Motivation
----------
This Fraction brings the Jaeger and OpenTracing dependencies and gives
a Swarm-way to configure the tracer.

The main advantage of using this fraction vs. the OpenTracing fraction
plus the Jaeger dependencies is that this brings the correct versions of
all the components required to have a fully working OpenTracing solution.

Modifications
-------------
* Added Jaeger fraction
* Added a property to the OpenTracing fraction to skip automatic resolution
of tracers

Result
------
Applications consuming this fraction now get automatic and effortless support
for a complete OpenTracing solution, including automatic capture of request
data, thanks to the TracingFilter from the OpenTracing fraction.

-----

### DIFF
--- a/fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/JaegerFraction.java
+++ b/fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/JaegerFraction.java
@@ -1,0 +1,84 @@
+package org.wildfly.swarm.jaeger;
+
+import org.jboss.logging.Logger;
+import org.wildfly.swarm.spi.api.Fraction;
+import org.wildfly.swarm.spi.api.annotations.Configurable;
+import org.wildfly.swarm.spi.api.annotations.DeploymentModule;
+
+import static com.uber.jaeger.Configuration.*;
+
+/**
+ * @author Juraci Paixão Kröhling
+ */
+@DeploymentModule(name = "org.wildfly.swarm.jaeger", slot = "deployment")
+@Configurable("swarm.jaeger")
+public class JaegerFraction implements Fraction<JaegerFraction> {
+    private static final Logger logger = Logger.getLogger(JaegerFraction.class);
+
+    private String serviceName = getDefault(JAEGER_SERVICE_NAME);
+
+    private String samplerType = getDefault(JAEGER_SAMPLER_TYPE);
+    private String samplerParameter = getDefault(JAEGER_SAMPLER_PARAM);
+    private String samplerManagerHost = getDefault(JAEGER_SAMPLER_MANAGER_HOST_PORT);
+
+    private String reporterLogSpans = getDefault(JAEGER_REPORTER_LOG_SPANS);
+    private String agentHost = getDefault(JAEGER_AGENT_HOST);
+    private String agentPort = getDefault(JAEGER_AGENT_PORT);
+    private String reporterFlushInterval = getDefault(JAEGER_REPORTER_FLUSH_INTERVAL);
+    private String reporterMaxQueueSize = getDefault(JAEGER_REPORTER_MAX_QUEUE_SIZE);
+
+    public String getServiceName() {
+        return serviceName;
+    }
+
+    public String getSamplerType() {
+        return samplerType;
+    }
+
+    public String getSamplerParameter() {
+        return samplerParameter;
+    }
+
+    public String getSamplerManagerHost() {
+        return samplerManagerHost;
+    }
+
+    public String getReporterLogSpans() {
+        return reporterLogSpans;
+    }
+
+    public String getAgentHost() {
+        return agentHost;
+    }
+
+    public String getAgentPort() {
+        return agentPort;
+    }
+
+    public String getReporterFlushInterval() {
+        return reporterFlushInterval;
+    }
+
+    public String getReporterMaxQueueSize() {
+        return reporterMaxQueueSize;
+    }
+
+    @Override
+    public String toString() {
+        return "JaegerFraction{" +
+                "serviceName='" + serviceName + '\'' +
+                ", samplerType='" + samplerType + '\'' +
+                ", samplerParameter='" + samplerParameter + '\'' +
+                ", samplerManagerHost='" + samplerManagerHost + '\'' +
+                ", reporterLogSpans='" + reporterLogSpans + '\'' +
+                ", agentHost='" + agentHost + '\'' +
+                ", agentPort='" + agentPort + '\'' +
+                ", reporterFlushInterval='" + reporterFlushInterval + '\'' +
+                ", reporterMaxQueueSize='" + reporterMaxQueueSize + '\'' +
+                '}';
+    }
+
+    private static String getDefault(String key) {
+        return System.getProperty(key, System.getenv(key));
+    }
+}
--- a/fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/JaegerFraction.java
+++ b/fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/JaegerFraction.java
@@ -1,9 +1,12 @@
 package org.wildfly.swarm.jaeger;
 
-import org.jboss.logging.Logger;
+import org.wildfly.swarm.config.runtime.AttributeDocumentation;
+import org.wildfly.swarm.spi.api.Defaultable;
 import org.wildfly.swarm.spi.api.Fraction;
 import org.wildfly.swarm.spi.api.annotations.Configurable;
 import org.wildfly.swarm.spi.api.annotations.DeploymentModule;
+
+import java.util.Optional;
 
 import static com.uber.jaeger.Configuration.*;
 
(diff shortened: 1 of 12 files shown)


jpkrohling:
cc @pavolloffay , @objectiser

fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/JaegerFraction.java
```diff
@@ -0,0 +1,84 @@
...
+import org.wildfly.swarm.spi.api.Fraction;
+import org.wildfly.swarm.spi.api.annotations.Configurable;
+import org.wildfly.swarm.spi.api.annotations.DeploymentModule;
+
+import static com.uber.jaeger.Configuration.*;
+
+/**
+ * @author Juraci Paixão Kröhling
+ */
+@DeploymentModule(name = "org.wildfly.swarm.jaeger", slot = "deployment")
```
jpkrohling, line 16: I kindly ask the Swarm developers to review this whole dependency thing for this PR... I have to admit that I don't really know how it works, and couldn't find a documentation about it, so, I kept trying things I've seen from other modules until I got a setup that works.

fractions/jaeger/src/main/resources/modules/com/uber/jaeger/main/module.xml
```diff
@@ -0,0 +1,18 @@
+<module xmlns="urn:jboss:module:1.3" name="com.uber.jaeger">
+  <resources>
+    <artifact name="com.uber.jaeger:jaeger-core:${version.jaeger}"/>
+    <artifact name="com.uber.jaeger:jaeger-tracerresolver:${version.jaeger}"/>
+    <artifact name="com.uber.jaeger:jaeger-thrift:${version.jaeger}"/>
+    <artifact name="org.apache.thrift:libthrift:0.9.2"/>
```
jpkrohling: The libraries that I fixed the version here are transitive dependencies. Shouldn't they be pulled automatically? It might become a PITA to maintain these versions in sync :(

fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/JaegerFraction.java
```diff
@@ -0,0 +1,84 @@
...
+@Configurable("swarm.jaeger")
+public class JaegerFraction implements Fraction<JaegerFraction> {
+    private static final Logger logger = Logger.getLogger(JaegerFraction.class);
+
+    private String serviceName = getDefault(JAEGER_SERVICE_NAME);
```
bobmcwhirter: I think it'd be worthwhile looking at Defaultable<String> and our dynamic defaults suppliers.  This would allow YAML to *know* a default value is applied, instead of an implicit default which just happens to be otherwise unset.

https://github.com/wildfly-swarm/wildfly-swarm/blob/master/core/spi/src/main/java/org/wildfly/swarm/spi/api/Defaultable.java#L21

jpkrohling:
@kevinearls , would you please test this?

bobmcwhirter (changes requested):

fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/JaegerFraction.java
```diff
@@ -0,0 +1,84 @@
...
+    private String samplerParameter = getDefault(JAEGER_SAMPLER_PARAM);
+    private String samplerManagerHost = getDefault(JAEGER_SAMPLER_MANAGER_HOST_PORT);
+
+    private String reporterLogSpans = getDefault(JAEGER_REPORTER_LOG_SPANS);
+    private String agentHost = getDefault(JAEGER_AGENT_HOST);
+    private String agentPort = getDefault(JAEGER_AGENT_PORT);
+    private String reporterFlushInterval = getDefault(JAEGER_REPORTER_FLUSH_INTERVAL);
+    private String reporterMaxQueueSize = getDefault(JAEGER_REPORTER_MAX_QUEUE_SIZE);
+
+    public String getServiceName() {
```
jpkrohling: I just realized that I'm missing the `@AttributeDocumentation` annotations.

fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/JaegerFraction.java
bobmcwhirter: The `*.deployment` module is a side-car non-active place you can stash classes that ultimately get mixed into a deployment.

With the above line, you're adding that classes in your `*.deployment` package to every deployment.  Which might be what you want, I dunno.

I do notice that one of those classes has `@Vetoed` on it, which may cause problems if this fraction can be used without CDI, as then there will be a `NoClassDefinitionFoundException` or such when it's attempted to be loaded as a part of the deployment.  Once again, I dunno the specifics of what's being attempted or the context in which this fraction may be used.

fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/deployment/JaegerInitializer.java
```diff
@@ -0,0 +1,93 @@
...
+import javax.servlet.ServletContextEvent;
+import javax.servlet.ServletContextListener;
+import javax.servlet.annotation.WebListener;
+
+import static com.uber.jaeger.Configuration.*;
+
+/**
+ * @author Juraci Paixão Kröhling
+ */
+@Vetoed
```
bobmcwhirter: This is the `@Vetoed` I mentioned above.  This may break if a user does not have CDI for whatever reason. If CDI is assumed to always be present, ignore me on this point.

fractions/jaeger/src/main/resources/modules/com/uber/jaeger/main/module.xml
bobmcwhirter: I'd investigate a few things here:

* Does WildFly have any of these artifacts already as compatible modules (commons-logging? commons-codec?)  If so, simply reference them the same was as you have `org.slf4j`.

* If WildFly does *not* have compatible versions of these modules, I'd recommend creating a new `module.xml` for each logical portion, possibly setting the `slot` to be something other than `main` as it might conflict with the `main` module coming from WildFly.  See our Ribbon bits for how, for instance, we brought in different versions of things like `guava` so as to not conflict with the core WildFly module, while also still keeping the `module.xml` as usual.

Additionally, *never* include versions direction within the `module.xml`, but prefer the `${version.whatever}` notations, using properties from the root `pom.xml`.  If required, add grist to the property name to include `jaeger` in case there are 3 other versions of `commons-codec` in use and defined via properties. in the `pom.xml`.

pom.xml
```diff
@@ -101,10 +101,11 @@
     <version.zipkin>1.13.1</version.zipkin>
 
     <!-- OpenTracing related versions -->
-    <version.opentracing>0.22.0</version.opentracing>
+    <version.opentracing>0.30.0</version.opentracing>
     <version.opentracing.tracerresolver>0.1.0</version.opentracing.tracerresolver>
-    <version.opentracing.servlet>0.0.8</version.opentracing.servlet>
+    <version.opentracing.servlet>0.0.9</version.opentracing.servlet>
     <version.opentracing.spanmanager>0.0.5</version.opentracing.spanmanager>
+    <version.jaeger>0.20.0</version.jaeger>
```
bobmcwhirter, line 107: such as here, you can define `version.jaeger.commons-codec` to reference from the above `module.xml`

fractions/jaeger/src/main/resources/modules/com/uber/jaeger/main/module.xml
bobmcwhirter: Transitive dependencies from Maven are *not* considered by JBoss-Modules and the `module.xml` files.  You do have to be explicit about defining module.xml's and what artifacts go into them.

jpkrohling: Alright, makes sense. I kept the versions fixed here, because this is something I did want to get feedback before getting it merged :)

fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/runtime/JaegerInstaller.java
```diff
@@ -0,0 +1,68 @@
...
+
+        if (archive.getName().endsWith(".war")) {
+            logger.logf(Logger.Level.INFO, "Installing the Jaeger integration for the deployment %s", archive.getName());
+            WARArchive webArchive = archive.as(WARArchive.class);
+            WebXmlAsset webXml = webArchive.findWebXmlAsset();
+
+            logger.logf(Logger.Level.INFO, "Adding the listener org.wildfly.swarm.jaeger.deployment.JaegerInitializer");
+            webXml.addListener("org.wildfly.swarm.jaeger.deployment.JaegerInitializer");
+
+            setContextParamIfNotNull(webXml, JAEGER_SERVICE_NAME, fraction.getServiceName());
```
objectiser, line 47: Just so I am clear, this is being done to allow the webXml to override the tracer config values provided in environment variables - otherwise the tracer could simply use them directly from env?

bobmcwhirter:
They could be sourced from YAML or other non-env sources

fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/runtime/JaegerInstaller.java
jpkrohling: Right, we are doing this because we want a "Swarm" native approach as well. The order should be: Swarm configuration (wherever it comes from), System property, Environment variables. The sysprop and env vars are the same as the "pure" Jaeger client.

fractions/jaeger/src/main/java/org/wildfly/swarm/jaeger/JaegerFraction.java
jpkrohling: > With the above line, you're adding that classes in your *.deployment package to every deployment. Which might be what you want, I dunno.

Perhaps if I explain the goal, it might be clearer. The main goal is to provide the developer with access to a proper combination of Jaeger+OpenTracing libraries. 

I tried to extend the OpenTracing fraction, to just add the Jaeger dependencies to the deployment, but couldn't get that working. The OpenTracing fraction has a `TracerResolver`, which tries to load a service via the `ServiceLoader` facility. The Jaeger fraction has a provider for this service, but it looks like the OpenTracing fraction can't "see" this provider class (they are possibly in different classloaders?). If a `Tracer` is resolved by the `TracerResolver`, it is then fed into the `GlobalTracer`, which can be consumed by the target application to create "business spans".

So, the final solution is to just "depend" on the OpenTracing fraction and implement a flag, set by the Jaeger fraction, telling the OpenTracing to not try to resolve the tracer via the `TraceResolver`. This way, the Jaeger fraction will register the tracer by itself with the `GlobalTracer`. One advantage of this is that it allows the target application to configure the Fraction like other Swarm fractions, the disadvantage is that this is far more complex than just adding the Jaeger dependencies to the target WAR.

jpkrohling:
PR updated with changes from the review. Once the PR is approved, I'll squash the commits.

jpkrohling:
The Windows build failed with `org.wildfly.swarm.management.console.ManagementConsoleTest.testHomePageShouldHaveCorrectTitle` , which doesn't seem to be caused by this PR.

objectiser:
@jpkrohling The problem (I believe) is that the current `TracerResolver` uses the system classloader - if it was changed to use the context classloader then it may work. Could you try a local test to see if this works - and if so raise a PR on the `TracerResolver`?

bobmcwhirter:
That's ignorable spurious intermittent.

Not at all related to your PR.

jpkrohling:
Talked to @objectiser and will look into `TracerResolver` as part of another JIRA ( https://issues.jboss.org/browse/HWKAPM-914 ).

If the fixes (based on the last review) are OK, I'll squash the commits and this will be ready to be merged.

bobmcwhirter (approved):

PR #557, opened 2017-06-27
SWARM-1419 - Provide an alternate tmpdir for Swarm jars.

bobmcwhirter:
Motivation
----------
As reported, WildFly Swarm produces a lot of tmpdir entries
that are required for the length of the process, which, since
we are enterprisey, might be a very very long time.  When
using the OS tempfile directory, a WFS process might run
afoul of eager OS tempfile directory clean-up activities,
resulting in undefined behaviour.  Or maybe defined-but-crappy
behaviour.

Modifications
-------------
Allow a user to specify a system property of swarm.io.tmpdir
(to mirror java.io.tmpdir), to point to a directory where
a WFS process should place it's long-lived temp files, separate
and apart from java.io.tmpdir, which may be used by the applicaitons
themselves.

If unspecified, the normal temp directory is used of course, so
order or precedence for WFS long-lived temp files is:

1) swarm.io.tmpdir
2) java.io.tmpdir
3) Whatever the JVM/OS decide ($TMP and friends)

Result
------
Better management of less-than-temporary temporary files.

-----

### DIFF
--- a/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/TempFileManager.java
+++ b/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/TempFileManager.java
@@ -17,6 +17,7 @@ package org.wildfly.swarm.bootstrap.util;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.util.Collections;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


Ladicek (approved):

heiko-braun:
lgtm

PR #558, opened 2017-06-28
SWARM-1420: Add dependencies that don't get aligned to bom-dependency

michalszynkiewicz:
Motivation
----------
Enable prod alignment of the affected dependencies

Modifications
-------------
The affected dependencies have been added to bom-dependency/pom.xml
Additionally, org.apache.httpcomponents:httpclient has been downgraded to 4.5.0 (from 4.5.2)
and its version has been moved to the parent pom

Result
------
All the dependencies of the prod build are now aligned.

-----

### DIFF
--- a/boms/bom-dependency/pom.xml
+++ b/boms/bom-dependency/pom.xml
@@ -25,6 +25,16 @@
 
   <properties>
     <version.openjdk.orb>8.0.6.Final</version.openjdk.orb>
+    <version.classmate>1.1</version.classmate>
+    <version.jackson-module-jaxb-annotations>2.5.4</version.jackson-module-jaxb-annotations>
+    <version.jcip-annotations>1.0</version.jcip-annotations>
+    <version.xnio-api>3.3.6.Final</version.xnio-api>
+    <version.httpclient>4.5</version.httpclient>
+    <version.httpcore>4.4.1</version.httpcore>
+    <version.apache-mime4j>0.6</version.apache-mime4j>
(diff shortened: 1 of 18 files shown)


Ladicek:
Not sure, but if you want to pin the dependency on `org.apache.httpcomponents:httpclient`, you'll probably also want to pin the dependency on `org.apache.httpcomponents:fluent-hc` (to the same version number). It's used in a bunch of modules of the test suite.

michalszynkiewicz:
@Ladicek Done

Ladicek (dismissed):

bobmcwhirter (dismissed):

Ladicek (dismissed):

Ladicek (approved):

kenfinnigan:
Based on what the changes involve, are we aligning testsuite dependencies?

To me that seems unnecessary

Ladicek:
I think it's because PME modifies the testsuite modules as well. I briefly talked to @michalszynkiewicz if it would be possible to skip the testsuite modules when doing dependency alignment, he even did some experiments, but I don't know the final results.

michalszynkiewicz:
yes, I tried to exclude testsuite, but PME goes through everything...
And it was failing with having two versions of the same library (it caused a conflict in alignment)

kenfinnigan:
Doesn't that seem like a bug that PME can't "ignore" certain things?

It's a waste of time for us to "align" things that aren't even required for a user building their project

PR #559, opened 2017-06-29
SWARM-1327: Fix file leaks in tests and add new maven enforcer rule to check if a file exist by pattern to avoid same problem in future

juangon:
Motivation
----------

Modifications
-------------

Result
------

- [] Have you built the project locally prior to submission with `mvn clean install`?

-----
@Ladicek @bobmcwhirter have other branches with different levels of changes for windows. This one fails in my environment, but as some tests doesn't fail always, let's see how it behaves in CI.

Together with this Pull I have other 2 options in a branch each:
- One with failing Windows tests disabling maven enforcer check. There are 16 tests in this situation. Not sure if this can be acceptable.
- Other where I just had to disable maven enforcer plugin for file leaks in just 1 test. A drawback is all jar files are exploded (even nested ones) in order to those files to be removed completely. BUT this option makes all tests execution soooooo slow in Windows I want to let it as the last option.

### DIFF
--- a/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/Main.java
+++ b/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/Main.java
@@ -15,6 +15,21 @@
  */
 package org.wildfly.swarm.bootstrap;
 
+import static java.nio.file.StandardWatchEventKinds.*;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
+import java.nio.file.WatchEvent.Kind;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
 import org.wildfly.swarm.bootstrap.env.ApplicationEnvironment;
 import org.wildfly.swarm.bootstrap.modules.BootModuleLoader;
 import org.wildfly.swarm.bootstrap.performance.Performance;
@@ -25,6 +40,16 @@ import org.wildfly.swarm.bootstrap.util.BootstrapProperties;
  */
 public class Main {
 
+    public static final String MAIN_PROCESS_FILE = "org.wildfly.swarm.mainProcessFile";
+
+    private static WatchService watcher;
+
+    private static Map<WatchKey,Path> keys;
+
+    private static ExecutorService shutdownService;
+
+    private static MainInvoker mainInvoker;
+
     public Main(String... args) throws Throwable {
         this.args = args;
     }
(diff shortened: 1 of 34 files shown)


core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/ApplicationModuleFinder.java
```diff
@@ -121,7 +120,7 @@ protected void addAsset(ModuleSpec.Builder builder, ApplicationEnvironment env)
             name = name.substring(0, dotLoc);
         }
 
-        File tmp = TempFileManager.INSTANCE.newTempFile(name, ext);
+        File tmp = File.createTempFile(name, ext);
```
Ladicek, line 123: Typically we change it the other way round. Why can't we keep this using the `TempFileManager` and let the shutdown hooks registered below to only close the `ResourceLoader`s? Maybe we have to make sure that the `ResourceLoader`s are closed _before_ we delete the file?

testsuite/testsuite-container/src/test/java/org/wildfly/swarm/container/test/DeploymentFailureTest.java
```diff
@@ -0,0 +1,31 @@
...
+      try {
+          swarm.deploy(a);
+          fail("should have throw a DeploymentException");
+      } catch (DeploymentException e) {
+          // expected and correct
+          assertThat(e.getArchive()).isSameAs(a);
+          assertThat(e.getMessage()).contains("org.jboss.modules.ModuleNotFoundException: com.i.do.no.exist:main");
+      }
+
+      swarm.stop();
```
Ladicek, line 29: Should be in `finally`?

Ladicek:
The `stopMain` changes look quite invasive to me.

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/ApplicationModuleFinder.java
juangon: Good question @Ladicek. It's needed in this case as we need to close the jar/war loader first before deleting the file. And, as shutdown hooks are executed in a non deterministic order, we can't ensure when it would be.

testsuite/testsuite-container/src/test/java/org/wildfly/swarm/container/test/DeploymentFailureTest.java
juangon: Yeah, it makes sense. In this case I just copy/pasted the code and left it unchanged (separated in 2 test classes though)

juangon:
It may sound like that @Ladicek  but, do you know about other "convention" to follow on Main files to "stop" them (as it seems we are just invoking those Main classes from other "root" Main class)?

juangon:
@Ladicek @bobmcwhirter Approach #2 PR is made here: https://github.com/wildfly-swarm/wildfly-swarm/pull/560

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/Main.java
```diff
@@ -34,16 +59,86 @@ public static void main(String... args) throws Throwable {
...
             new Main(args).run();
         } catch (Throwable t) {
             t.printStackTrace();
             throw t;
         }
     }
 
+    private static void register(File directory, Path file) throws IOException {
+        watcher = FileSystems.getDefault().newWatchService();
+        keys = new HashMap<WatchKey,Path>();
```
Ladicek, line 95: This always replaces the `keys` map with a new one, so the code as is only works with a single `Key` by necessity. In that case, it could probably be drastically simplified.

juangon: Ok thanks, I will review that.

Ladicek:
> It may sound like that @Ladicek but, do you know about other "convention" to follow on Main files to "stop" them (as it seems we are just invoking those Main classes from other "root" Main class)?

I don't know. Maybe inside `Swarm.start`, we could add a shutdown hook that would call `stop` on the same instance?

juangon:
> I don't know. Maybe inside Swarm.start, we could add a shutdown hook that would call stop on the same instance?

Ok, I think these other kind of logic can be great for more Windows testing. Thanks!

juangon:
> I don't know. Maybe inside Swarm.start, we could add a shutdown hook that would call stop on the same instance?

Tried this @Ladicek and not luck. Some errors when stopping due to some SelfContainedContainer thread pool rejection in executing an stop task, and temp file still not removed.

juangon:
Approach 3 (exploded jar files) pull request was made here: https://github.com/wildfly-swarm/wildfly-swarm/pull/562

juangon:
Closing as we chose approach 3 here: https://github.com/wildfly-swarm/wildfly-swarm/pull/562

PR #560, opened 2017-06-29
SWARM-1327: Approach #2 (disable file leak check for failing tests in Windows)

juangon:
Fix file leaks in tests and add new maven enforcer rule to check if a file exist by pattern to avoid same problem in future.

Motivation
----------

Modifications
-------------

Result
------

-----


juangon:
@Ladicek @bobmcwhirter in this case I disabled the file leak check (through Maven enforcer plugin) for 16 failing tests. You can check those pom.xml files. Let's see if it passes all tests here.

juangon:
Closing as we agreed to choose approach #3 here: https://github.com/wildfly-swarm/wildfly-swarm/pull/562

PR #562, opened 2017-06-30
SWARM-1327: Approach 3. Explode jar files belonging to boot loader resource root into directories.

juangon:
Let's see the difference in tests executing between this approach and previous one due to the exploding.

Main differences are in changes made to core/boostrap...ApplicationModuleFinder and NestedJarResourceLoader.

### DIFF
--- a/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/Main.java
+++ b/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/Main.java
@@ -15,6 +15,19 @@
  */
 package org.wildfly.swarm.bootstrap;
 
+import static java.nio.file.StandardWatchEventKinds.*;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
+import java.nio.file.WatchEvent.Kind;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
 import org.wildfly.swarm.bootstrap.env.ApplicationEnvironment;
 import org.wildfly.swarm.bootstrap.modules.BootModuleLoader;
 import org.wildfly.swarm.bootstrap.performance.Performance;
@@ -25,6 +38,14 @@ import org.wildfly.swarm.bootstrap.util.BootstrapProperties;
  */
 public class Main {
 
+    public static final String MAIN_PROCESS_FILE = "org.wildfly.swarm.mainProcessFile";
+
+    private static WatchService watcher;
+
+    private static ExecutorService shutdownService;
+
+    private static MainInvoker mainInvoker;
+
     public Main(String... args) throws Throwable {
         this.args = args;
     }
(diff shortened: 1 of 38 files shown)


core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/BootstrapUtil.java
```diff
@@ -0,0 +1,73 @@
...
+            if (je.isDirectory()) {
+                continue;
+            }
+            InputStream is = null;
+            FileOutputStream fo = null;
+            try {
+                is = jarFile.getInputStream(je);
+                fo = new FileOutputStream(fl);
+                while (is.available() > 0) {
+                    fo.write(is.read());
```
Ladicek: Good for testing, but one byte at a time will be dog slow :-)

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/NestedJarResourceLoader.java
```diff
@@ -131,23 +132,42 @@ public static ResourceLoader loaderFor(URL base, String rootPath, String loaderP
...
             }
         } else if (urlString.startsWith("file:")) {
             if (loaderName.endsWith(".jar") || loaderName.endsWith(".war")) {
-                return ResourceLoaders.createJarResourceLoader(
-                        loaderName,
-                        new JarFile(new File(urlString.substring(5), loaderPath)));
+                final File file = new File(urlString.substring(5), loaderPath);
+                final JarFile jarFile = new JarFile(file);
+
+                File tmpDir = TempFileManager.INSTANCE.newTempDirectory("nextedjarloader", null);
```
Ladicek: `nexted` :-)

Ladicek:
Wow, it passed even on Windows? That's cool. Are we sure that the temporary exploded directories would be caught by the enforcer as well? The result is very nice!

I'm still not sure about the `stopMain` thing, it's too intrusive to me. But maybe there's no better way, I don't know.

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/BootstrapUtil.java
juangon: Sorry, too much copy/paste here :-)

juangon:
Yes @Ladicek it is also caught by enforcer plugin as it accomplish the rule of 5 numbers or more in the name (regardless if it's directory or not).

juangon:
FYI the only file leak check I had to disable was testsuite-webservices. It seems caused by server not releasing the provided HandledChain xml file on shutdown in Windows, which prevent files to be deleted.

bobmcwhirter (approved):

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/Main.java
```diff
@@ -34,16 +55,77 @@ public static void main(String... args) throws Throwable {
...
+            String processFile = System.getProperty(MAIN_PROCESS_FILE);
+
+            if (processFile != null) {
+                shutdownService = Executors.newSingleThreadExecutor();
+                shutdownService.submit(() -> {
+                        File uuidFile = new File(processFile);
+                        try {
+                            File watchedDir = uuidFile.getParentFile();
+                            register(watchedDir);
+                            processEvents(watchedDir, uuidFile.toPath());
```
Ladicek, line 68: I'd inline the `register` method, it's just 2 lines, and mainly, the `processEvents` should be called something like `waitForFileDeleted` or something like that.

Ladicek:
OK, just one minor comment, otherwise LGTM.

juangon:
Oh I was going to say tests are almost spending same time, so good to go :-)

PR #563, opened 2017-06-30
SWARM-1426 - Allow specification of deployment export directory.

bobmcwhirter:
Motivation
----------
Sometimes when using swarm.export.deployment, $PWD isn't the
best place to export the deployment.

Modifications
-------------
swarm.export.deployment, if set to "true" (or without a value),
continues to export in $PWD.  If set to some other value,
that value will be treated as a path to a directory where
the deployment should be exported.  If that directory does
not exist, it will be created.

Result
------
-Dswarm.export.deployment=./some/dir works as you might hope.

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
@@ -17,7 +17,9 @@ package org.wildfly.swarm.container.runtime;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeDeployer.java
```diff
@@ -232,8 +234,18 @@ public void deploy(Archive<?> deployment, String asName) throws DeploymentExcept
                 }
             }
 
-            if (BootstrapProperties.flagIsSet(SwarmProperties.EXPORT_DEPLOYMENT)) {
-                final File out = new File(deployment.getName());
+            //if (BootstrapProperties.flagIsSet(SwarmProperties.EXPORT_DEPLOYMENT)) {
+            String exportLocation = System.getProperty(SwarmProperties.EXPORT_DEPLOYMENT);
```
Ladicek: I'm thinking if maybe someone sets this property to `false` (maybe by some automation), but that's probably too much of a corner case.

Ladicek (dismissed):

Ladicek (approved):

PR #564, opened 2017-07-03
SWARM-1374: Fix Weld version mismatch

mkouba:
Motivation
----------
Swarm core depends on weld-se artifacts which are not managed in WildFly feature packs.

Modifications
-------------
Remove weld-api and cdi-api version properties. Verify that Weld SE version is aligned
with version provided by the WildFly feature pack. Move version.weld-se property to root pom.

Result
------
It is not possible to remove the version.weld-se property completely. However, we can
minimize its impact and verify the correct version is used.

### DIFF
--- a/core/container/src/main/resources/modules/org/jboss/weld/se/main/module.xml
+++ b/core/container/src/main/resources/modules/org/jboss/weld/se/main/module.xml
@@ -1,12 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module xmlns="urn:jboss:module:1.5" name="org.jboss.weld.se">
   <resources>
-    <!--<resource-root path="weld-se-core-${version.weld}.jar"/>-->
-    <artifact name="org.jboss.weld.environment:weld-environment-common:${version.weld}"/>
-    <artifact name="org.jboss.weld:weld-api:${version.weld-api}"/>
-    <artifact name="org.jboss.weld:weld-spi:${version.weld-api}"/>
-    <artifact name="org.jboss.weld:weld-core-impl:${version.weld}"/>
-    <artifact name="org.jboss.weld.se:weld-se-core:${version.weld}"/>
+    <artifact name="org.jboss.weld.environment:weld-environment-common:${version.weld-se}"/>
+    <artifact name="org.jboss.weld:weld-api:${org.jboss.weld:weld-api:jar.version}"/>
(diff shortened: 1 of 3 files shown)


Ladicek (dismissed):

mkouba:
It's seems we have a lot of similar test failures. Let me find the cause.

core/container/src/main/resources/modules/org/jboss/weld/se/main/module.xml
```diff
@@ -1,12 +1,11 @@
...
 <module xmlns="urn:jboss:module:1.5" name="org.jboss.weld.se">
   <resources>
-    <!--<resource-root path="weld-se-core-${version.weld}.jar"/>-->
-    <artifact name="org.jboss.weld.environment:weld-environment-common:${version.weld}"/>
-    <artifact name="org.jboss.weld:weld-api:${version.weld-api}"/>
-    <artifact name="org.jboss.weld:weld-spi:${version.weld-api}"/>
-    <artifact name="org.jboss.weld:weld-core-impl:${version.weld}"/>
-    <artifact name="org.jboss.weld.se:weld-se-core:${version.weld}"/>
+    <artifact name="org.jboss.weld.environment:weld-environment-common:${version.weld-se}"/>
+    <artifact name="org.jboss.weld:weld-api:${org.jboss.weld:weld-api:jar.version}"/>
```
kenfinnigan, line 5: Are these version properties coming from WF somewhere?

mkouba: ` version.weld-se` is defined in swarm root pom. `org.jboss.weld:weld-api:jar.version` and `org.jboss.weld:weld-core-impl:jar.version` are set by [dependencyversion-maven-plugin](https://github.com/wildfly-swarm/wildfly-swarm/pull/564/files#diff-1d9cc2c72a3cdd14f37073b9bd158ddaR41) and correspond to  https://github.com/wildfly/wildfly/blob/master/pom.xml#L200 (when importing a pom in <dependencyManagement> only dependencies are considered, not properties, so we need to take the current version of the weld artifacts).

kenfinnigan: ok cool

kenfinnigan (approved):

PR #572, opened 2017-07-06
SWARM-1435: Introduce AsciidoctorJ fraction

gastaldi:
Motivation
----------
This includes the AsciidoctorJ fraction to the list of available fractions

Modifications
-------------
Added asciidoctorj fraction + tests

Result
------
AsciidoctorJ can now be executed inside a WildFly Swarm application without much fuzz.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -580,6 +580,12 @@
 
       <dependency>
         <groupId>org.wildfly.swarm</groupId>
+        <artifactId>asciidoctorj</artifactId>
+        <version>2017.8.0-SNAPSHOT</version>
+      </dependency>
+
+      <dependency>
+        <groupId>org.wildfly.swarm</groupId>
         <artifactId>batch-jberet</artifactId>
         <version>2017.8.0-SNAPSHOT</version>
       </dependency>
@@ -1499,6 +1505,7 @@
         <module>fractions/javaee/resource-adapters</module>
         <module>fractions/javaee/webservices</module>
 
+        <module>fractions/asciidoctorj</module>
         <module>fractions/camel</module>
         <module>fractions/drools-server</module>
         <module>fractions/fluentd</module>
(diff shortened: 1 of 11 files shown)


gastaldi:
Retest this please

pom.xml
```diff
@@ -1381,6 +1387,8 @@
 
     <module>client-apis/health-api</module>
 
+    <module>fractions/asciidoctorj</module>
```
Ladicek: Should move to the `unsupported` profile? Same with the testsuite module below.

Ladicek:
I'm just wondering about the value of this. Shouldn't I be able to do this just as part of the deployment without any issues? Or are there issues that this fraction eliminates?

pom.xml
kenfinnigan: +1 to move to unsupported

kenfinnigan:
In theory yes, but there are manifest or module.xml dependency machinations required

Ladicek:
If I'd have to touch module stuff when trying to deploy AsciidoctorJ, then a fraction is a perfect solution, thanks!

gastaldi:
The Windows Build fails because of https://github.com/jruby/jruby/issues/3657. 

@bobmcwhirter any chance to have the JRuby team look at it?

enebo:
@gastaldi we will try and reexamine our stale dll files issue.  The reason we have not fixed this so far is it is harder then it looks.  One engineer is really afraid of putting the dlls into a predictable location as hackers on Windows use that as a mechanism for replacing the dll.  Once we load the dll then windows is incapable of unloading it so we cannot delete the file from the JVM (although just saying that makes me wonder if there is an unload function).  Our binary launcher tries to overlay the JVM process so we do not necessarily have wrapper script which could attempt to clean the files up.

gastaldi:
Apparently upgrading Asciidoctor helped on making the Windows tests pass. Let me ask to run the tests again, just in case

gastaldi:
Retest this please

bobmcwhirter (approved):

PR #575, opened 2017-07-12
SWARM-1447: default datasource not set in the EE subsystem

Ladicek:
Motivation
----------
When a default datasource is created (see
`DatasourceAndDriverCustomizer.customizeDefaultDatasource`),
it doesn't get set as the default datasource in the `ee` subsystem.
This is the primary place where WildFly goes to for the default
datasource.

Since databases are most often accessed via JPA, this issue wasn't
visible much, because the `jpa` fraction takes that default datasource
from the `datasources` fraction and sets it as a default for the `jpa`
subsystem. This means that JPA works with the default datasource just
fine, but other parts of the application server don't (e.g. injecting
`@Resource DataSource` into a servlet).

Modifications
-------------
The default datasource is set in the `ee` subsystem now.
It is no longer set in the `jpa` subsystem, which is now forced
to consult the `ee` subsystem.

Result
------
Default datasource, if exists, is properly available to the entirety
of the application server. This improves Java EE 7 compatibility.

-----

### DIFF
--- a/fractions/javaee/datasources/src/main/java/org/wildfly/swarm/datasources/runtime/DefaultDatasourceCustomizer.java
+++ b/fractions/javaee/datasources/src/main/java/org/wildfly/swarm/datasources/runtime/DefaultDatasourceCustomizer.java
@@ -13,14 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.wildfly.swarm.jpa.runtime;
+package org.wildfly.swarm.datasources.runtime;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Instance;
 import javax.inject.Inject;
 
-import org.wildfly.swarm.datasources.runtime.DefaultDatasource;
-import org.wildfly.swarm.jpa.JPAFraction;
+import org.wildfly.swarm.config.EE;
 import org.wildfly.swarm.spi.api.Customizer;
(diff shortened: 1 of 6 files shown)


kenfinnigan (dismissed):

Ladicek:
Oh, one thing I wanted to ask about: the `datasources` fraction currently doesn't depend on the `ee` fraction. That will typically be brought in by another fraction (e.g. `undertow`), but I was thinking if maybe we should add that dependency. Default datasource is a Java EE 7 feature and if I bring in the `datasources` fraction, I expect I will be able to use it. Of course, the `ee` subsystem brings in a bunch of other unrelated stuff...

kenfinnigan:
I'd not add `ee` as there may be situations where `datasources` are required but not `ee` so would be good to avoid the hard dependency

Ladicek:
retest this please

Ladicek:
Ah, interesting, the batch test seems to fail because of a missing datasource. Hmm...

Ladicek:
I don't understand why the batch test would require a datasource, but it does now. So I added a dependency on H2 to that testsuite module.

Ladicek:
Aha! There were 2 issues: I (probably) used the Config API wrongly, and mainly, the customizer checked if the `@DefaultDatasource Instance<String>` is satisfied or not per CDI rules, but that's wrong -- it's always satisfied, just sometimes has the `null` value.

bobmcwhirter (approved):

PR #577, opened 2017-07-14
Fix ReflectionUtil.getMethodsWithAnnotation()

sweidac:
SWARM-1449: Fix ReflectionUtil.getMethodsWithAnnotation()

Motivation
----------
To enable the use of DRY and SRP in tests, a superclass should handle container/deployment - setup.

Modifications
-------------
Changed call from getDeclaredMethods() to getMethods() to also get methods of superclasses.

Result
------
Now test classes can derive from superclasses that deal with the setup of swarm and the deployment.


bobmcwhirter:
Okay to test

kenfinnigan (approved):

PR #580, opened 2017-07-20
SWARM-1458: Load JAXP API’s from container module

heiko-braun:
Leverage JAXP redirect facility in jboss modules to load custom JAXP implementation

### DIFF
--- a/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/MainInvoker.java
+++ b/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/MainInvoker.java
@@ -1,5 +1,10 @@
 package org.wildfly.swarm.bootstrap;
 
+import org.jboss.modules.Module;
+import org.jboss.modules.ModuleIdentifier;
+import org.jboss.modules.ModuleLoadException;
+import org.wildfly.swarm.bootstrap.modules.BootModuleLoader;
+
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
@@ -8,11 +13,6 @@ import java.lang.reflect.Modifier;
 import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.List;
-
-import org.jboss.modules.Module;
-import org.jboss.modules.ModuleIdentifier;
-import org.jboss.modules.ModuleLoadException;
-import org.wildfly.swarm.bootstrap.modules.BootModuleLoader;
 
 /**
  * @author Bob McWhirter
--- a/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/MainInvoker.java
+++ b/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/MainInvoker.java
@@ -1,10 +1,5 @@
 package org.wildfly.swarm.bootstrap;
 
-import org.jboss.modules.Module;
-import org.jboss.modules.ModuleIdentifier;
-import org.jboss.modules.ModuleLoadException;
-import org.wildfly.swarm.bootstrap.modules.BootModuleLoader;
-
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
@@ -13,6 +8,11 @@ import java.lang.reflect.Modifier;
 import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.List;
+
+import org.jboss.modules.Module;
+import org.jboss.modules.ModuleIdentifier;
+import org.jboss.modules.ModuleLoadException;
+import org.wildfly.swarm.bootstrap.modules.BootModuleLoader;
 
 /**
  * @author Bob McWhirter
--- a/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/ContainerModuleFinder.java
+++ b/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/ContainerModuleFinder.java
@@ -89,6 +89,40 @@ public class ContainerModuleFinder extends AbstractSingleModuleFinder {
                         null,
                         ModuleIdentifier.create("javax.enterprise.api"), false));
 
+        builder.addDependency(
+                DependencySpec.createModuleDependencySpec(
+                        PathFilters.acceptAll(),
+                        PathFilters.acceptAll(),
+                        PathFilters.acceptAll(),
+                        PathFilters.acceptAll(),
+                        ClassFilters.acceptAll(),
+                        ClassFilters.acceptAll(),
+                        null,
+                        ModuleIdentifier.create("org.apache.xalan"), true));
+
+        builder.addDependency(
+                DependencySpec.createModuleDependencySpec(
+                        PathFilters.acceptAll(),
+                        PathFilters.acceptAll(),
+                        PathFilters.acceptAll(),
+                        PathFilters.acceptAll(),
+                        ClassFilters.acceptAll(),
+                        ClassFilters.acceptAll(),
+                        null,
+                        ModuleIdentifier.create("org.apache.xerces"), false));
+
+        builder.addDependency(
+                DependencySpec.createModuleDependencySpec(
+                        PathFilters.acceptAll(),
+                        PathFilters.acceptAll(),
+                        PathFilters.acceptAll(),
+                        PathFilters.acceptAll(),
+                        ClassFilters.acceptAll(),
+                        ClassFilters.acceptAll(),
+                        null,
+                        ModuleIdentifier.create("org.codehaus.woodstox"), false));
+
+
         ApplicationEnvironment environment = ApplicationEnvironment.get();
 
         environment.bootstrapModules()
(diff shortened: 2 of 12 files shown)


core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/MainInvoker.java
```diff
@@ -9,11 +14,6 @@
 import java.util.Arrays;
 import java.util.List;
 
-import org.jboss.modules.Module;
-import org.jboss.modules.ModuleIdentifier;
-import org.jboss.modules.ModuleLoadException;
-import org.wildfly.swarm.bootstrap.modules.BootModuleLoader;
-
```
kenfinnigan: Can you check your code formatting, as I don't think this should be a change.

fractions/javaee/undertow/src/main/java/org/wildfly/swarm/undertow/UndertowFraction.java
```diff
@@ -31,15 +31,8 @@
...
-import static org.wildfly.swarm.spi.api.Defaultable.ifAnyExplicitlySet;
-import static org.wildfly.swarm.spi.api.Defaultable.integer;
-import static org.wildfly.swarm.undertow.UndertowProperties.DEFAULT_AJP_PORT;
-import static org.wildfly.swarm.undertow.UndertowProperties.DEFAULT_HOST;
-import static org.wildfly.swarm.undertow.UndertowProperties.DEFAULT_HTTPS_PORT;
-import static org.wildfly.swarm.undertow.UndertowProperties.DEFAULT_HTTP_LISTENER;
-import static org.wildfly.swarm.undertow.UndertowProperties.DEFAULT_HTTP_PORT;
-import static org.wildfly.swarm.undertow.UndertowProperties.DEFAULT_SERVER;
+import static org.wildfly.swarm.spi.api.Defaultable.*;
+import static org.wildfly.swarm.undertow.UndertowProperties.*;
```
kenfinnigan: Also check code formatting here.

testsuite/testsuite-jsp/pom.xml
```diff
@@ -27,6 +27,11 @@
       <artifactId>undertow</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.wildfly.swarm</groupId>
```
kenfinnigan, line 31: Are you sure this is needed? With recent changes it shouldn't be

testsuite/testsuite-jsp/src/test/java/org/wildfly/swarm/jsp/test/JstlIT.java
```diff
@@ -0,0 +1,37 @@
...
+import java.io.IOException;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+/**
+ * @author Heiko Braun
+ */
+public class JstlIT {
+
+    @Test
```
kenfinnigan, line 31: How does this test work without any Arquillian deployments?

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/ContainerModuleFinder.java
```diff
@@ -89,6 +89,40 @@ public void buildModule(ModuleSpec.Builder builder, ModuleLoader delegateLoader)
                         null,
                         ModuleIdentifier.create("javax.enterprise.api"), false));
 
+        builder.addDependency(
```
kenfinnigan, line 92: Does adding these to the container module prevent a user from overriding them, ie. by providing their own implementation in a WAR?

Might be worth adding a test to verify that works

heiko-braun: If we want overrides, we may need to provide hooks similar to this one: https://github.com/jboss-modules/jboss-modules/blob/1.x/src/main/java/org/jboss/modules/Main.java#L391

heiko-braun: But I would say we can still add it when users request it.

heiko-braun: Rereading your comment: "in a WAR". Yes, that's something we can verify. I think it should work ootb, but I am not sure.

kenfinnigan: Right, I'm less concerned about someone wanting to override the module directly than being able to add dependencies to a WAR and have it work

kenfinnigan:
Given the low level change, it's probably also a good idea to run -examples locally to make sure it doesn't impact anything we don't expect, like Camel. If you haven't already that is

testsuite/testsuite-jsp/src/test/java/org/wildfly/swarm/jsp/test/JstlIT.java
heiko-braun: Seems to be using the default deployment of that module.

testsuite/testsuite-jsp/pom.xml
heiko-braun: the arq dep seems to be required.

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/ContainerModuleFinder.java
heiko-braun: I've added another test: It seems that WEB-INF/lib overrides work as expected

heiko-braun:
@kenfinnigan Examples passes for me

testsuite/testsuite-jsp/src/test/java/org/wildfly/swarm/jsp/test/JstlIT.java
kenfinnigan: I should have clarified that this appears to be a plain JUnit test, no Arquillian involved at all

testsuite/testsuite-jsp/pom.xml
```diff
@@ -22,15 +22,30 @@
   <packaging>war</packaging>
 
   <dependencies>
+
```
kenfinnigan: What was missing to need this?

testsuite/testsuite-jsp/pom.xml
```diff
@@ -49,6 +64,12 @@
             <version>${project.version}</version>
             <executions>
               <execution>
+                <id>package</id>
```
kenfinnigan, line 59: Doesn't this configuration come from the parent?

Likewise with `useUberJar`, also should only be activated with profile `uberjar`.

testsuite/testsuite-jsp/src/test/java/org/wildfly/swarm/jsp/test/JstlIT.java
heiko-braun: The pre and post phases start/stop Swarm. I guess that's why it works to use a plain test that pokes into `localhost`. Honestly I didn't think much about it and merely took the same approach than the existing tests. Would you like to have this reworked?

testsuite/testsuite-jsp/pom.xml
heiko-braun: No, `package` is not executed by default. All the tests currently run based off `wildfly-swarm:start/stop`. 

I'll move the test to the `uberjar` profile

heiko-braun: @kenfinnigan Can you explain what you with profile `uberjar `? I don't any such thing in the testsuite modules. There is a profile `uberjar ` in standalone-servers. Am I missing something?

kenfinnigan: Sorry, was getting confused thinking this was in -examples.

I'll ask a different question, why does the test need to be run as an uberjar? Does it not work without it?

testsuite/testsuite-jsp/src/test/java/org/wildfly/swarm/jsp/test/JstlIT.java
kenfinnigan: There are other tests that work this way? I thought all tests used `@RunAs`

testsuite/testsuite-jsp/pom.xml
heiko-braun: Right, it doesnt work without it, because it requires proper jboss-module loading in order replace the JAXP APIs that are also provided by the JDK. The majority of testsuite uses wildfly-swarm:run, which juggles classloading scopes outside modules and here it doesn't work properly.

kenfinnigan: ah ok

heiko-braun:
retest this please

heiko-braun:
Good morning @kenfinnigan. Could you briefly outline what changes are necessary from your point if view to accept the PR?

kenfinnigan (approved):

PR #589, opened 2017-07-31
SWARM-1372: Apply LicenseMojo to community BOM.

mkouba:
NOTE: THIS PR IS NOT READY FOR MERGE YET. Depends on https://github.com/wildfly-swarm/wildfly-swarm-fraction-plugin/pull/41 and subsequent fraction plugin release.

Motivation
----------
Gather license data of all artifacts neededed for community BOM.

Modifications
-------------
Added configuration to boms/bom/pom.xml and complementary resources
(xsl, etc.).
    
Result
------
It is possible to use licenses profile to generate/transform
liceses.xml.

### DIFF
--- a/boms/src/main/resources/licenses-transform.xsl
+++ b/boms/src/main/resources/licenses-transform.xsl
@@ -1,0 +1,204 @@
+<?xml version="1.0"?>
+<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+
+  <xsl:output method="xml" encoding="utf-8"/>
+
+  <!-- Global constants -->
+  <xsl:variable name="lowercase" select="'abcdefghijklmnopqrstuvwxyz'"/>
+  <xsl:variable name="uppercase" select="'ABCDEFGHIJKLMNOPQRSTUVWXYZ'"/>
+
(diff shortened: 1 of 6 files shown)


boms/src/main/resources/licenses-transform.xsl
```diff
@@ -0,0 +1,204 @@
...
+    <xsl:param name="origin" select="'Fixed'"/>
+    <license>
+      <!-- For debug purposes only -->
+      <!-- xsl:comment><xsl:value-of select="$origin"/></xsl:comment-->
+      <name><xsl:value-of select="$name"/></name>
+      <url><xsl:value-of select="$url"/></url>
+    </license>
+  </xsl:template>
+
+</xsl:stylesheet>
```
Ladicek, line 204: I fear we'll soon have a gazillion copies of this file, each with their own changes. We should coordinate around a single source of truth that would be just released to Maven Central. Just a comment, not a request for changing anything in this PR.

mkouba: Well, the idea is to have one file for all Swarm BOMs. I'm afraid that if we try to create a similar xsl for all projects the file will become really large and not so easy to maintain :-(

kenfinnigan (approved):

PR #590, opened 2017-07-31
SWARM-1470: topology-webapp doesn't work together with jaxrs

Ladicek:
Motivation
----------
When the `topology-webapp` fraction is included, a small deployment
`topology-webapp.war` is automatically generated by the fraction
and deployed.

When the `jaxrs` fraction is included, all `.war` deployments that
aren't JAX-RS-enabled explicitly, get a a class looking like

```
@ApplicationPath("/")
public class WildFlySwarmDefaultJAXRSApplication extends Application {
    ...
}
```

When the two fractions are included simultaneously, all the servlet
and static content from `topology-webapp.war` are shadowed by
the JAX-RS stuff, so `topology-webapp` can't possibly work.

Modifications
-------------
Exclude the `jaxrs` subsystem from `topology-webapp.war` via
`jboss-deployment-structure.xml`.

The existing code for manipulating `jboss-deployment-structure.xml`
didn't support that, so support for excluding subsystems
programmatically is also added.

Result
------
`topology-webapp` and `jaxrs` live in peace and harmony.

-----

### DIFF
--- a/core/spi/src/main/java/org/wildfly/swarm/spi/api/JBossDeploymentStructureAsset.java
+++ b/core/spi/src/main/java/org/wildfly/swarm/spi/api/JBossDeploymentStructureAsset.java
@@ -25,12 +25,14 @@ import org.jboss.shrinkwrap.api.asset.Asset;
 import org.jboss.shrinkwrap.descriptor.api.Descriptors;
 import org.jboss.shrinkwrap.descriptor.api.jbossdeployment13.DependenciesType;
 import org.jboss.shrinkwrap.descriptor.api.jbossdeployment13.DeploymentType;
+import org.jboss.shrinkwrap.descriptor.api.jbossdeployment13.ExcludeSubsystemsType;
 import org.jboss.shrinkwrap.descriptor.api.jbossdeployment13.ExclusionsType;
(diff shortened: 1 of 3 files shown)


kenfinnigan:
As stated in the JIRA, this is not an appropriate approach.

It's perfectly valid to have `jaxrs` and `topology-webapp` in the same deployment.

Ladicek:
Let's continue the discussion in the JIRA, but I'm pretty confident I can convince you :-)

bobmcwhirter:
This relates to SWARM-1462 also, I think.

The problem is implicit deployments (and deployments-via-GAV) should be considered "complete" usually, and not have our DeployProcessors applied to them.

This affected the swagger-webapp deployment, causing it to have a JAXRS Application generated for is, wrongly.

I've taken a more general approach which should (I hope) remedy much of this later today.

Ladicek:
That totally makes sense and would resolve this too.

Ladicek:
Marked _hold_ for now, but I suspect this will get closed eventually, because what @bobmcwhirter describes sounds like a more correct solution.

Ladicek:
#591 is a more general solution and fixes this issue as well.

PR #596, opened 2017-08-03
Upgrade to WildFly-Camel-4.8.0

tdiesler:

-----

### DIFF
--- a/fractions/camel/components/opentracing/module.conf
+++ b/fractions/camel/components/opentracing/module.conf
@@ -1,0 +1,2 @@
+org.apache.camel.component.opentracing
+org.wildfly.swarm.camel.core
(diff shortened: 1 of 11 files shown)


kenfinnigan:
Why the need to make the maven enforcer rule empty?

bobmcwhirter:
I know camel would routinely violate the rule if tests failed. It doesn't clean up itself very well. I think we should fix camel not disable the rule.

tdiesler:
Unfortunately, I don't know what to do with these rules or what they even are. It also seems that these rules were introduced in previous releases regardless of breaking the camel tests in the process. 

Perhaps we can start by explaining what these rules are for and then talk about what can be done for the camel integration

kenfinnigan:
@tdiesler, basically the rules ensure that fractions clean up after themselves.

Whether that be temp space or other erroneous files that a test might create.

kenfinnigan:
Closing this

PR #599, opened 2017-08-04
SWARM-1480: Don't use Arquillian test classes in client mode to detect fractions

juangon:
Motivation
----------
There are cases when test classes want to use some client tools that could cause some additional fraction to be detected. These client tools are usually used in Arquillian client mode tests (@RunAsClient or testable="false"). These fractions aren't needed and can cause some unexpected errors.

Modifications
-------------
Added a ClassReader annotation visitor to check if a test class is an Arquillian client mode one. In that case, test class will not count on fraction detection.

Result
------
Arquillian test classes in client mode won't affect fraction detection, avoiding unexpected results due to unwanted fractions being added in these cases.

-----

Sorry for not adding a test. Tried but unsucessfully. I may try on 14th with the help of @bobmcwhirter or @Ladicek if needed

### DIFF
--- a/meta/fraction-metadata/src/main/java/org/wildfly/swarm/fractions/ArquillianClientAnnotationSeekingClassVisitor.java
+++ b/meta/fraction-metadata/src/main/java/org/wildfly/swarm/fractions/ArquillianClientAnnotationSeekingClassVisitor.java
@@ -1,0 +1,87 @@
+/**
+ * Copyright 2015-2017 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.fractions;
+
+import org.objectweb.asm.AnnotationVisitor;
+import org.objectweb.asm.ClassVisitor;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.Type;
+
+/**
+ * @author Juan Gonzalez
+ */
+public class ArquillianClientAnnotationSeekingClassVisitor extends ClassVisitor {
+
+    static final String DEPLOYMENT_ANNOTATION = "org/jboss/arquillian/container/test/api/Deployment";
+    static final String RUN_AS_CLIENT_ANNOTATION = "org/jboss/arquillian/container/test/api/RunAsClient";
+
+    private int clientCounter;
+    private int containerCounter;
+
+    public ArquillianClientAnnotationSeekingClassVisitor() {
+        super(Opcodes.ASM5);
+    }
+
+    public boolean isClient() {
+        return this.clientCounter > 0 && this.containerCounter == 0;
+    }
+
+    @Override
+    public MethodVisitor visitMethod(final int __,
+                                     final String ___,
+                                     final String desc,
+                                     final String signature,
+                                     final String[] exceptions) {
+
+        return new MethodVisitor(Opcodes.ASM5) {
+            @Override
+            public AnnotationVisitor visitAnnotation(final String desc,
+                                                     final boolean __) {
+
+                Type type = Type.getType(desc);
+                String className = type.getInternalName();
+
+                if (className.equals(DEPLOYMENT_ANNOTATION)) {
+                    containerCounter++;
+                    return new ArquillianAnnotationVisitor();
+                }
+
+                if (className.equals(RUN_AS_CLIENT_ANNOTATION)) {
+                    clientCounter++;
+                }
+
+                return null;
+            }
+       };
+    }
+
+    private final class ArquillianAnnotationVisitor extends AnnotationVisitor {
+
+            public ArquillianAnnotationVisitor() {
+                super(Opcodes.ASM5);
+            }
+
+            @Override
+            public void visit(final String name, final Object value) {
+
+                if (name.equals("testable") && value.equals(Boolean.FALSE)) {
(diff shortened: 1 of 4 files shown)


kenfinnigan (dismissed):

meta/fraction-metadata/src/main/java/org/wildfly/swarm/fractions/ArquillianClientAnnotationSeekingClassVisitor.java
```diff
@@ -0,0 +1,100 @@
...
+
+                if (clazz.equals(DEPLOYMENT_ANNOTATION) &&
+                        name.equals("testable")) {
+
+                    if (value.equals(Boolean.FALSE)) {
+                        clientCounter++;
+                        containerCounter--;
+                    } else {
+                        containerCounter++;
+                    }
```
Ladicek: Personally, I don't quite understand the interactions between `ArquillianClientAnnotationSeekingClassVisitor` and `ArquillianAnnotationVisitor` around how the counters are incremented and decremented. It seems to me that `visitAnnotationDefault` doesn't apply to our situation, that method is called when looking at annotation definitions, not their usages, right? If so, I think this code could be simplified a bit.

kenfinnigan: @Ladicek do you want this resolved before merging?

juangon:
Hi @Ladicek and @kenfinnigan. I think 
Ladislav is right and visitAnnotationDefault could be removed. 
About why I was using counters, is a way to balance appearance of Client and in container annotations along the way, as it isn't possible (AFAIK) to gather this info in advance before visiting each annotation and their values.
I can fix the extra method when I come back on Monday.

Thanks to both!

Ladicek:
I was actually thinking maybe `visitAnnotationDefault` is also called on annotation usages to let the caller know about the default value of an annotation attribute, in case the attribute isn't set directly. (ASM javadoc is so terse...!) Let me test this quick.

BTW using counters is OK from my POV. I just don't understand the sequence in which they are incremented and decremented, but I'll check it out in a debugger.

Ladicek:
OK, so I stepped through this in the debugger and:

- the `visitAnnotationDefault` method is never called (and as it is right now, it would never work, because it passes a type descriptor to the constructor of `ArquillianAnnotationVisitor`, but that expects an internal class name), so should be deleted
- the `ArquillianAnnotationVisitor` is only invoked if the `@Deployment` annotation has some parameters defined; if it's only used like `@Deployment`, is doesn't get called at all
- the `else` branch in the `ArquillianAnnotationVisitor` increments the `containerCounter` for the _second_ time, this should be deleted

I propose we merge this and then I send a follow-up PR with a little cleanup.

juangon:
@Ladicek be aware of the different use cases that could exist for Client or in-container mode (multiple deployment annotations,  Deployment without or with testable true or false, etc). Be aware that removing some code could remove some use case check.
Anyway I won't be able to check until next week.

Ladicek:
@juangon When it comes to the `@Deployment` annotation, I tried these 3 cases:

- `@Deployment`: `ArquillianAnnotationVisitor` is never called -- which is OK, the `containerCounter` has already been incremented (because `testable` defaults to `true`)
- `@Deployment(testable = false)`: `ArquillianAnnotationVisitor` is called and decrements the `containerCounter` -- which is OK, it was incremented before and that needs correction
- `@Deployment(testable = true)`: `ArquillianAnnotationVisitor` is called and increments the `containerCounter` for the second time -- which is wrong, it was already incremented

juangon:
Ok it sounds good. Will check next week whenever I remember why I added that increment.
Thanks!

meta/fraction-metadata/src/main/java/org/wildfly/swarm/fractions/ArquillianClientAnnotationSeekingClassVisitor.java
```diff
@@ -0,0 +1,93 @@
...
+
+            public ArquillianAnnotationVisitor(String clazz) {
+                super(Opcodes.ASM5);
+                this.clazz = clazz;
+            }
+
+            @Override
+            public void visit(final String name, final Object value) {
+
+                if (clazz.equals(DEPLOYMENT_ANNOTATION) &&
```
Ladicek: I'd remove the `clazz` field and this check, because `ArquillianAnnotationVisitor` is only used when this condition already holds. So just check if `name` equals `"testable"`.

juangon: Thanks @Ladicek. Yep, I think I initially added those conditions as I used them for other classes during my tests, but doesn't apply in this final version. Thanks for checking.

meta/fraction-metadata/src/main/java/org/wildfly/swarm/fractions/FractionUsageAnalyzer.java
```diff
@@ -284,6 +312,13 @@ private void loadDetectorsAndScanners() {
         detectorsLoaded = true;
     }
 
+    private String transformClassSource(String entry) {
+        entry = entry.replace("WEB-INF/classes/", "");
+        entry = entry.replace(".class", "");
+        entry = entry.replace(File.separator, ".");
```
Ladicek: I'm just thinking -- did you try this on Windows? I'm not sure what separators will be used e.g. for .zip entries...?

juangon: Hmm no I didn't. Let me check..

juangon: Ok, so thanks to this I could fix this case for Windows too, I updated the PR. Seems no File.separator is needed after all. Thanks!

Ladicek: OK, thanks!

Ladicek (approved):

kenfinnigan:
retest this please

kenfinnigan:
@juangon is this all good to be merged now?

juangon:
Yes it's @kenfinnigan .

Ladicek:
+1

PR #600, opened 2017-08-08
SWARM-1472: Arquillian fraction has test scope in all boms.

mkouba:
Motivation
----------
Arquillian fraction should have "test" scope in all boms so that its
dependencies are not added to the resulting artifact during compilation.

Modifications
-------------
Added <swarm.fraction.scope>test</swarm.fraction.scope> to the
arquillian fraction. The property will be currently ignored (arquillian
fraction is handled in a special way) but it's good to be consistent.

Result
------
Nothing changes.

### DIFF
--- a/arquillian/arquillian/pom.xml
+++ b/arquillian/arquillian/pom.xml
@@ -24,6 +24,7 @@
 
   <properties>
     <swarm.fraction.stability>stable</swarm.fraction.stability>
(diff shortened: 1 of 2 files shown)


mkouba:
See also https://github.com/wildfly-swarm/wildfly-swarm-fraction-plugin/pull/42

Ladicek (approved):

kenfinnigan (approved):

PR #606, opened 2017-08-16
SWARM-1519: GradleResolver is ignoring the classifier for artifact do…

ncabanis:
…wnloads

Motivation
----------
The GradleResolver is using a wrong download url for artifacts with a
classifier. If an artifact has a classifier, it is downloading an
artifact without classifier and storing it in the Gradle cache with the
classifier added to the name. This leads to artifact a to be stored as
aritfact b.
The problem arose with jbossws-cxf-resources-5.1.5.Final-wildfly1000.

The management console was another place where this problem arose. There
was the problem that an artifact was not found as the was no artifact
without classifier existent.

Modifications
-------------
The classifier is added to the download url.

Result
------
Artifacts are found and downloaded correctly.
How does this commit affect the product in behavior or usage.

-----


bobmcwhirter:
okay to test

bobmcwhirter (approved):

PR #609, opened 2017-08-17
SWARM-1522: Consul TTL service check not recovering from errors

Lasrik:
Motivation
----------
If any error occurs while sending ttl check (network error, consul unreachable..) the thread will silently die. All other consul components are resilient and will continue working once consul is reachable again.
But the check will then never be called again.

Modifications
-------------
Simply catch and log any exception in the thread so it keeps on trying.

Result
------
makes this part more resilient

-----


bobmcwhirter:
Okay to test

bobmcwhirter:
Add to whitelist

kenfinnigan (approved):

PR #612, opened 2017-08-22
SWARM-1525: Leverage maven args to determine build file

heiko-braun:

-----

With these changes in place, any codepath that previously did assume `pom.xm`, scans for `env.MAVEN_CMD_LINE_ARGS` and leverage the argument to `mvn -f <BUILD_FILE_NAME>` to determine which build file should be used.

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/internal/MavenArgsParser.java
+++ b/core/container/src/main/java/org/wildfly/swarm/internal/MavenArgsParser.java
@@ -1,0 +1,89 @@
+package org.wildfly.swarm.internal;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.StringTokenizer;
+
+/**
+ * Created by hbraun on 22.08.17.
+ */
+public class MavenArgsParser {
+
+    public enum ARG {
+            F("-f", "--file"),
+            P("-P", "--activate-profiles");
+
+        // the flag is the short form
+        private final String flag;
+
+        // the alt flag is the long form (different syntax applies)
+        private final String altFlag;
+
+        ARG(String flag, String altFlag) {
+            this.flag = flag;
+            this.altFlag = altFlag;
+        }
+
+        public String getFlag() {
+            return flag;
+        }
+        public String getAltFlag() {
+            return altFlag;
+        }
+    }
+
+    private Map<ARG, String> argValues = new HashMap<>();
+
+    private MavenArgsParser(String commandLine) {
+
+        StringTokenizer tok = new StringTokenizer(commandLine, " ");
+        while (tok.hasMoreTokens()) {
+            String token = tok.nextToken();
+            ARG[] args = ARG.values();
+            for (ARG arg : args) {
+
+                // irregular values, without whitespace between flag and value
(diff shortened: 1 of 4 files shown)


core/container/src/test/java/org/wildfly/swarm/internal/FileSystemUtilsTest.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.Optional;
+
+/**
+ * Created by hbraun on 22.08.17.
+ */
+public class FileSystemUtilsTest {
+
+    @Test
+    public void testMavenArgsPresent() {
+        String cmd = "clean -f pom.xml install";
```
kenfinnigan: Should this test actually use a name other than `pom.xml` to ensure the code is actually taking any name from -f and not just ending up with pom.xml by default?

heiko-braun: It only tests the `MavenArgsParser` that gives access to tokenised values. In this case anything passed to `-f`. There is no "fallback" in the parser itself.

kenfinnigan: Ok, but shouldn't it test what we're planning to use, ie. `pom-product.xml`?

heiko-braun: done

core/container/src/main/java/org/wildfly/swarm/internal/MavenArgsParser.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.StringTokenizer;
+
+/**
+ * Created by hbraun on 22.08.17.
+ */
+public class MavenArgsParser {
```
Ladicek, line 11: In addition to `-f`, you can also do `--file`.

Also you can do `-fanotherpom.xml` and it works too. This only applies to the single-character option, for `--file`, there needs to be a space.

heiko-braun: I'll add support for `--file` but would ignore the `-fanotherpom.xml` case for now. This is mostly about us developing and verifying the boosters

kenfinnigan: I agree that this is important for our immediate use case, but I would prefer we implement all cases otherwise we only introduce problems down the road as we think something should work but it doesn't

Ladicek: Yea at the very least, I'd add a big TODO comment. The current usecase for this is our work on boosters, but anyone could use `-f` with their own Swarm project, totally unrelated to boosters.

heiko-braun:
@kenfinnigan @Ladicek The additional commits should cover the other cases you described.

heiko-braun:
Bummer, i've introduced another bug. Working on it.

heiko-braun:
@kenfinnigan Once this passes I think we are good to go

kenfinnigan (approved):

PR #615, opened 2017-08-23
SWARM-1526: topology.js should have a mechanism to wait for initial topology

Ladicek:
Motivation
----------
When `topology()` is called, it subscribes to the SSE stream,
which immediately sends back the current topology. There
should be a mechanism to wait for that, otherwise users are
susceptible to race conditions.

Modifications
-------------
The `topology()` function no longer returns the `Topology` object.
Instead, it returns a promise that is resolved with the `Topology`
object once the first event from the stream arrives. That way,
the race condition can never occur, the user can only get hold
of the `Topology` object after the initial topology is obtained.

This is obviously a breaking change. Migration path is pretty simple,
though; instead of `var topo = topology(); ... use topo ...`, users
should use `topology().then(function(topo) { ... use topo ... });`.

Result
------
Users are no longer susceptible to race conditions with topology.js.
Of course, they have to migrate to the new solution.

-----

### DIFF
--- a/fractions/topology-webapp/src/main/resources/topology.js
+++ b/fractions/topology-webapp/src/main/resources/topology.js
@@ -182,16 +182,28 @@ var topology = (function() {
       return o;
     }
 
+    var initialTopologyObtained = false;
+    var result = deferred();
 
     var sse = new EventSource( options.context + "/system/stream" );
     sse.addEventListener('topologyChange', function(message) {
       console.log('topology.js: topology changed: ', message.data);
       topology = JSON.parse(message.data);
       scheduler = schedule(topology);
+
+      if (!initialTopologyObtained) {
+        initialTopologyObtained = true;
+        result.resolve(_topology);
+      }
     });
 
     sse.onerror = function(e) {
       console.log( "topology.js: topology SSE error", e );
+
+      if (!initialTopologyObtained) {
+        initialTopologyObtained = true;
+        result.reject(_topology);
+      }
     };
 
     function onTopologyChange(f) {
@@ -213,7 +225,8 @@ var topology = (function() {
       enumerable: true
     });
     Object.freeze(_topology);
-    return _topology;
+
+    return result.promise;
   }
 
   function merge(defaults, provided) {


bobmcwhirter:
I was thinking along the lines of

    var topo = topology( function(initial) { ... } );

or return a future from topology() constructor

    var topo = topology().then( function(initial){...} )

But I'm also not a js expert.  Maybe we should ask @lance

Ladicek:
Yea, `topology().then(...)` is exactly what I was thinking about too, but that's a breaking change. If that's fine, I can change this.

bobmcwhirter:
Ah, in that case, the first one, with an optional callback arg to the `topology()` factory maybe.

Ladicek:
Except `topology()` can already accept an argument, which is an object with options. Obviously, it can accept a 2nd optional argument... and I guess JS world is just used to this... oh well. I'll take a stab at this.

lance:
In principle, I'm against adding surface area to an API no matter how small. So my initial reaction is -1 on this. But I do see the issue you are facing could be a real problem. The goal for me would be to minimize any noticeable changes to the external API. 

So I understand correctly, the race condition you are trying to avoid is when calling the factory method to obtain the initial topology, the object returned has a `topology` property which may or may not yet be initialized. Since neither the factory function

Ref:
https://github.com/Ladicek/wildfly-swarm/blob/6c832c58475387c871333e6aa0b58201b6f7a3ad/fractions/topology-webapp/src/main/resources/topology.js#L11

Nor the returned object's `topology` property wait around for initialization:

Ref:
https://github.com/Ladicek/wildfly-swarm/blob/6c832c58475387c871333e6aa0b58201b6f7a3ad/fractions/topology-webapp/src/main/resources/topology.js#L220-L233

Personally, I think my preferred change would be to just change the factory function to return a `Promise` that resolves when everything has been fully initialized. But as you said, @Ladicek that would be a breaking change. How are those handled in general with swarm?

Another option, although still a breaking change, would be to just change the `topology` property to be a function that returns a promise which resolves once the initial topology has been obtained, and after that returns `Promise.resolve(_localCachedTopology)`. Honestly, I kind of like this option a little better than changing the factory function to return a promise because it allows for deprecation. 

The existing `.topology` property accessor can stay the same, but issue a deprecation warning with a link to some description of the race condition, and a new async function, e.g. `currentTopology()` could be added to takes its place. Eventually `.topology` property is removed, and API surface area reduced to initial size.

Adding `awaitInitialTopology` feels a little clunky to me, personally since it requires changes to existing code to avoid the race condition (as would all of the solutions discussed), but it also never goes away, and is, in my opinion less elegant than changing either the factory or the property to behave correctly.

Ladicek:
So changing the `topology` property itself isn't enough, because there are also the `getJSON`, `postJSON` and `ajax` functions that also assume that topology has already been obtained.

Instead of changing this PR, I added a second version here: https://github.com/Ladicek/wildfly-swarm/commits/await-initial-topology-2 After we decide on the approach to use, I will of course amend this PR.

When it comes to breaking changes: I think Swarm generally breaks quite often, but I just don't share that mindset. Maybe that's a mistake on my side :-) If I hear from @bobmcwhirter or @kenfinnigan that breaking change here is fine, I'd gladly change the `topology()` function to return a promise. It's the cleanest solution I can think of.

lance:
I like your second implementation, fwiw.

Ladicek:
Thanks @lance. I amended this PR.

lance (dismissed):

bobmcwhirter:
IRC vote was "it's okay to break backwards compatibility" if it makes it more correct.  (right @kenfinnigan?)

kenfinnigan:
indeedy

Ladicek:
Ah I didn't see that, I closed my IRC client a while ago. I'll take a stab at that tomorrow.

Ladicek:
Pushed a third solution, where the `topology()` function returns a promise of the `Topology` object, which is resolved when initial topology is obtained.

scottmarlow:
I'm not sure why yet but looks like https://ci.wildfly-swarm.io/job/wildfly-swarm-docker-test-linux/94/console had a transient failure:

> curl -v POST http://neo4j:neo4j@localhost:7474/user/neo4j/password -dpassword=changethis
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 
>   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Could not resolve host: POST; Name or service not known
> * Closing connection 0
> curl: (6) Could not resolve host: POST; Name or service not known
> * About to connect() to localhost port 7474 (#1)
> *   Trying ::1...
> * Connected to localhost (::1) port 7474 (#1)
> * Server auth using Basic with user 'neo4j'
> > POST /user/neo4j/password HTTP/1.1
> > Authorization: Basic bmVvNGo6bmVvNGo=
> > User-Agent: curl/7.29.0
> > Host: localhost:7474
> > Accept: */*
> > Content-Length: 19
> > Content-Type: application/x-www-form-urlencoded
> > 
> } [data not shown]
> * upload completely sent off: 19 out of 19 bytes
> * Recv failure: Connection reset by peer
> 
> 100    19    0     0  100    19      0   2324 --:--:-- --:--:-- --:--:--  2324
> * Closing connection 1
> curl: (56) Recv failure: Connection reset by peer

scottmarlow:
retest this please

scottmarlow:
https://ci.wildfly-swarm.io/job/wildfly-swarm-docker-test-linux/95/ passed this time.

Ladicek:
I had to push a small change -- the function didn't return the promise, but the "promise controller". Fixed now.

kenfinnigan:
@lance, would appreciate if you can take another quick look at this and give your ok. Thanks

fractions/topology-webapp/src/main/resources/topology.js
```diff
@@ -182,16 +182,28 @@ var topology = (function() {
...
+        initialTopologyObtained = true;
+        result.resolve(_topology);
+      }
     });
 
     sse.onerror = function(e) {
       console.log( "topology.js: topology SSE error", e );
+
+      if (!initialTopologyObtained) {
+        initialTopologyObtained = true;
```
lance, line 204: I am a little confused by this. Why is this set to `true` in the case of an error?

lance (approved):
Looks good to me with the exception of one question inline.

fractions/topology-webapp/src/main/resources/topology.js
Ladicek: Because the promise is rejected and can't be fulfilled again. I was contemplating about this a little -- maybe in case of an error, I shouldn't reject the promise in hope that the stream will heal itself later, but ultimately decided against this. Maybe I should rename the variable a bit, `initialTopologyObtained` only explains the success case, not the failure one.

bobmcwhirter:
Are we ready to merge this?

Ladicek:
Thanks, can we also merge https://github.com/wildfly-swarm/wildfly-swarm-users-guide/pull/33?

PR #616, opened 2017-08-24
ENTSWM-80 - Bundle javadoc .zip

bobmcwhirter:
Turns out that the public-javadoc project respects -Dswarm.product.build, so that's good.

It also references bom-all, which might be less good.  Not quite sure how we adjust that for product.

Nonetheless, bundle it as a .zip to go into the repo.


kenfinnigan:
completely forgot we had a separate javadoc module for it.

Is it possible to "smoosh" what's in that module to work inside boms/ and then upstream just publishes the result of bom-all for javadoc?

bobmcwhirter:
Yah. Don't merge. Let me think on this and improve it.

kenfinnigan (approved):

PR #620, opened 2017-08-30
SWARM-1033 -- Support usage.txt

bobmcwhirter:
Motivation
----------
Some -swarm.jar apps might enjoy being able to display some
usage information after booting.

Modifications
-------------

If a user's application has one of:

* `usage.txt`
* `META-INF/usage.txt`
* `WEB-INF/usage.txt`

it will be displayed immediately prior to the message that
states "WildFly Swarm is Ready".

Variables within this file, in the form of ${foo} will be interpolated
if the associated configurable is available.  This is useful for
pointing out which port or URL might be available, such as by using
${swarm.http.port} within the file.

If the configurable is un-set, then the variable will *not* be
interpolated, and you will see a literal `${whatever}` in the
output.  `usage.txt` is *not* intended to be a fancy scripting
language. The variables *are* exposed from defaults after all
configurables have been processed from the fraction configurations,
so they do not need to be explicitly set.

Result
------
Ability to display customized usage information at boot.

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/Swarm.java
+++ b/core/container/src/main/java/org/wildfly/swarm/Swarm.java
@@ -640,6 +640,8 @@ public class Swarm {
             tryToStopAfterStartupError(t, swarm);
             throw t;
         }
+
+        displayUsage();
     }
 
     public static void stopMain() throws Exception {
@@ -666,6 +668,10 @@ public class Swarm {
                 System.exit(1);
             }
         }
+    }
+
+    private static void displayUsage() throws Exception {
+        swarm.server.displayUsage();
     }
 
     private static ArtifactLookup artifactLookup() {
--- a/core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/ConfigurableManagerUsageVariableSupplier.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/ConfigurableManagerUsageVariableSupplier.java
@@ -1,0 +1,36 @@
+package org.wildfly.swarm.container.runtime.usage;
+
+import java.util.Optional;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+import org.wildfly.swarm.container.runtime.ConfigurableHandle;
+import org.wildfly.swarm.container.runtime.ConfigurableManager;
+
+/**
+ * Created by bob on 8/30/17.
+ */
+@ApplicationScoped
+public class ConfigurableManagerUsageVariableSupplier implements UsageVariableSupplier {
+
+    @Inject
+    public ConfigurableManagerUsageVariableSupplier(ConfigurableManager manager) {
+        this.manager = manager;
+    }
+
+    @Override
+    public Object valueOf(String name) throws Exception {
+        Optional<ConfigurableHandle> configurable = this.manager.configurables().stream()
+                .filter(e -> e.key().toString().equals(name))
+                .findFirst();
+
+        if (configurable.isPresent()) {
+            return configurable.get().currentValue();
+        }
+
+        return null;
+    }
+
+    private final ConfigurableManager manager;
+}
--- a/core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/ModuleUsageProvider.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/ModuleUsageProvider.java
@@ -1,0 +1,50 @@
+package org.wildfly.swarm.container.runtime.usage;
+
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.stream.Collectors;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.jboss.modules.Module;
+import org.jboss.modules.ModuleIdentifier;
+
+/**
+ * Created by bob on 8/30/17.
+ */
+@ApplicationScoped
+public class ModuleUsageProvider implements UsageProvider {
+
+    String USAGE_TXT = "usage.txt";
+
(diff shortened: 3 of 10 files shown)


core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/ConfigurableManagerUsageVariableSupplier.java
```diff
@@ -0,0 +1,36 @@
...
+        this.manager = manager;
+    }
+
+    @Override
+    public Object valueOf(String name) throws Exception {
+        Optional<ConfigurableHandle> configurable = this.manager.configurables().stream()
+                .filter(e -> e.key().toString().equals(name))
+                .findFirst();
+
+        if (configurable.isPresent()) {
```
gastaldi: `returnconfigurable.orElseGet(null)`

core/container/src/main/java/org/wildfly/swarm/Swarm.java
```diff
@@ -668,6 +670,17 @@ private static void tryToStopAfterStartupError(final Throwable errorCause, final
         }
     }
 
+    private static void displayUsage() throws Exception {
+        /*
```
gastaldi: Commented out code must go

core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/ConfigurableManagerUsageVariableSupplier.java
gastaldi, line 28: `return configurable.orElse(null);`

gastaldi (dismissed):
Looks great to me! Fantastic work!

core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/ModuleUsageProvider.java
```diff
@@ -0,0 +1,48 @@
...
+
+        if (in == null) {
+            in = cl.getResourceAsStream(USAGE_TXT);
+        }
+
+        if (in != null) {
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(in))) {
+                return String.join("\n",
+                                   reader.lines()
+                                           .collect(Collectors.toList()));
```
gastaldi: `reader.lines().stream().collect(Collectors.joining("\n"))`

core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/UsageCreator.java
```diff
@@ -0,0 +1,66 @@
...
+
+            cur = result.end();
+        }
+
+        replaced.append(raw.substring(cur));
+
+        return replaced.toString();
+    }
+
+    private static Pattern PATTERN = Pattern.compile("[^\\\\]?(\\$\\{([^}]+)\\})");
```
gastaldi: `private static final `

core/container/src/main/java/org/wildfly/swarm/internal/SwarmMessages.java
```diff
@@ -162,6 +162,11 @@
     // ------------------------------------------------------------------------
 
     @LogMessage(level = Logger.Level.INFO)
+    @Message(id = 88888, value = "\n\n========================================================================\n\n%s\n\n========================================================================\n")
```
gastaldi: Using `%n` instead of `\n` will use the platform-specific line separator as returned by System.getProperty("line.separator").

gastaldi: If the Formatter class is used, otherwise ignore that

core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/ConfigurableManagerUsageVariableSupplier.java
bobmcwhirter: Not quite, since that'd just swap `isPresent()` for a null check, since I then have to, if present, also call `currentValue()` which will NPE if I return null from `orElse(null)`.

core/container/src/main/java/org/wildfly/swarm/internal/SwarmMessages.java
bobmcwhirter: Not explicitly using a formatter class.  No idea what is used internally.  @jamezp ?

jamezp: Internally it will end up using `String.format()` which uses `java.util.Formatter`.

core/container/src/test/java/org/wildfly/swarm/container/runtime/usage/UsageCreatorTest.java
```diff
@@ -0,0 +1,48 @@
+package org.wildfly.swarm.container.runtime.usage;
```
jamezp, line 1: This is missing a copyright too. Not sure how strict you all are about that though.

bobmcwhirter:
@gastaldi -- if you approve, can you approve via the review+approve workflow?

gastaldi (approved):
Approved

core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/ConfigurableManagerUsageVariableSupplier.java
gastaldi: Ah yes, you're right.

core/container/src/test/java/org/wildfly/swarm/container/runtime/usage/UsageCreatorTest.java
gastaldi: All the new classes are missing a copyright tbh ;)

PR #622, opened 2017-08-31
SWARM-1355 -- Support configuration value filtering.

bobmcwhirter:
Motivation
----------
Some people may encrypt or otherwise obscure configuration
values and desire a generic way to filter the values from
the YAML files (or properties) through a decryption service.

Modifications
-------------
* Added `spi` to the BOM
* Added `ConfigurationFilter` class to the SPI
* Scan applications for implementations of aforementioned class
* Provide all implementations an opportunity to filter configuration values.

Result
------
Users can ROT13 or otherwise mutate configuration values specified
in YAML or as properties.

-----

### DIFF
--- a/boms/src/main/resources/bom-template.xml
+++ b/boms/src/main/resources/bom-template.xml
@@ -55,6 +55,11 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
+        <groupId>org.wildfly.swarm</groupId>
+        <artifactId>spi</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
         <groupId>org.jboss.shrinkwrap</groupId>
         <artifactId>shrinkwrap-bom</artifactId>
         <version>${version.shrinkwrap}</version>
--- a/core/container/src/main/java/org/wildfly/swarm/Swarm.java
+++ b/core/container/src/main/java/org/wildfly/swarm/Swarm.java
@@ -15,6 +15,7 @@
  */
 package org.wildfly.swarm;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Method;
@@ -27,6 +28,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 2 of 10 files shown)


bobmcwhirter:
I'd like to verify that this isn't a terrible performance impact on build-times on CI prior to merging.

Ladicek:
Isn't this solved by the notion of a vault in WildFly? I didn't try using vault with Swarm yet, though.

bobmcwhirter:
retest this please

bobmcwhirter:
@Ladicek Yah, it probably is solved by Vault, but I'm ignorant of Vault, and the user requested this behavior specifically, probably because they have their own method for value obfuscation.

bobmcwhirter:
Seems like our config-api probably also doesn't support the equivalent of the `standalone.xml` `<vault>` tag.  We should investigate that as a separate issue.

bobmcwhirter:
Added JIRA https://issues.jboss.org/browse/SWARM-1534

Ladicek:
I'm thinking, even without vault concerns, we're already delegating to WildFly for `${value.interpolation}`, and with generic value filtering, we're adding a second possibility to do this kind of things. Which might lead to surprising errors. E.g. when someone adds a dependency on a library that accidentally also provides a Swarm config value filter... Oh well. Great power & great responsibility.

bobmcwhirter:
retest this please

boms/src/main/resources/bom-template.xml
```diff
@@ -55,6 +55,11 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
+        <groupId>org.wildfly.swarm</groupId>
```
kenfinnigan, line 58: What's the need for spi to be directly present in BOM?

kenfinnigan (approved):

boms/src/main/resources/bom-template.xml
bobmcwhirter: If a user wants to implement the `ConfigurationFilter` interface they will need to drag in the SPI explicitly, instead of just getting it implicitly/transitively. Hence the need for it to be BOM-able.

kenfinnigan: no problem

lborupj:
Has anybody used this feature for writing a ConfigurationFilter inside an archive (WAR)? I did and it is unfortunately not "picked up", eventhough the implementing class is seen as a resource, it is unable to load it as the URL points to a location in relation for CURRENT_DIR and not the web archive embedded within the fatjar.

`
private void initializeConfigFiltersFatJar() throws ModuleLoadException, IOException, ClassNotFoundException {
        Indexer indexer = new Indexer();

        Module appModule = Module.getBootModuleLoader().loadModule(APPLICATION_MODULE_NAME);
        Iterator<Resource> iter = appModule.iterateResources(PathFilters.acceptAll());
        while (iter.hasNext()) {
            Resource each = iter.next();
            if (each.getName().endsWith(".class")) {
                if (!each.getName().equals("module-info.class")) {
                    try (InputStream is = each.openStream()) {
                        indexer.index(is);
                    } catch (IOException e) {
                        // ignore
                    }
                }
            }
        }
`

PR #627, opened 2017-09-05
SWARM-1536: topology-openshift uses wrong namespace when pod runs in a different namespace than the image was built in

Ladicek:
Motivation
----------
`org.wildfly.swarm.topology.openshift.runtime.NamespaceService`
contains a simple algorithm to select which namespace to target
when reading the topology. This checks the `KUBERNETES_NAMESPACE`
env var, then `OPENSHIFT_BUILD_NAMESPACE` and then possibly
the list of all projects iff it contains a single entry.

However, the `KUBERNETES_NAMESPACE` env var isn't a feature
of Kubernetes or OpenShift. It only gets added to the pod template
by the Fabric8 Maven plugin, using the Kubernetes Downward API.
If the deployment method used doesn't involve Fabric8 Maven plugin,
this env var will be missing. Then, the `OPENSHIFT_BUILD_NAMESPACE`
env var contains the name of the namespace _where the image was
built_, which doesn't have to be the name of the namespace in which
the pod _is currently running_ (though it's probably quite common).

It seems reading `/var/run/secrets/kubernetes.io/serviceaccount/namespace`
is the best option, see also
https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#accessing-the-api-from-a-pod

Modifications
-------------
Changed the `NamespaceService` to first try reading
`/var/run/secrets/kubernetes.io/serviceaccount/namespace`.
If it fails, continue with the original algorithm.

Result
------
If the pod is running in a different namespace than the image
was built in, and Fabric8 Maven plugin wasn't involved, the namespace
is still selected correctly.

-----

### DIFF
--- a/fractions/topology-openshift/src/main/java/org/wildfly/swarm/topology/openshift/runtime/NamespaceService.java
+++ b/fractions/topology-openshift/src/main/java/org/wildfly/swarm/topology/openshift/runtime/NamespaceService.java
@@ -15,6 +15,10 @@
  */
 package org.wildfly.swarm.topology.openshift.runtime;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
 
 import com.openshift.restclient.IClient;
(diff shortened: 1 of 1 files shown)


kenfinnigan:
You tested this works with a booster or the like?

Ladicek:
No booster uses topology-openshift. I didn't test with our topology-openshift test (which is currently failing because of this issue), because that expects a Swarm release available in some Maven repo, but I tested manually.

kenfinnigan:
Should this check happen before KUBERNETES_NAMESPACE or just before OPENSHIFT_BUILD ?

Ladicek:
Definitely before `OPENSHIFT_BUILD_NAMESPACE`. We can check `KUBERNETES_NAMESPACE` before or after, I don't mind.

kenfinnigan:
Can we switch to do `KUBERNETES_NAMESPACE` first?

Ladicek:
Done.

kenfinnigan (approved):

PR #629, opened 2017-09-06
SWARM-1537 - Added configuration option for OpenTracing's skipPattern

jpkrohling:

-----

> Have you built the project locally prior to submission with `mvn clean install`?

Only the relevant parts.

### DIFF
--- a/fractions/opentracing/src/main/java/org/wildfly/swarm/opentracing/runtime/OpenTracingInstaller.java
+++ b/fractions/opentracing/src/main/java/org/wildfly/swarm/opentracing/runtime/OpenTracingInstaller.java
@@ -1,12 +1,15 @@
 package org.wildfly.swarm.opentracing.runtime;
 
+import io.opentracing.contrib.web.servlet.filter.TracingFilter;
 import org.jboss.logging.Logger;
 import org.jboss.shrinkwrap.api.Archive;
+import org.wildfly.swarm.opentracing.OpenTracingFraction;
 import org.wildfly.swarm.spi.api.DeploymentProcessor;
 import org.wildfly.swarm.spi.runtime.annotations.DeploymentScoped;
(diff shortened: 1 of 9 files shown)


fractions/opentracing/src/main/java/org/wildfly/swarm/opentracing/runtime/OpenTracingInstaller.java
```diff
@@ -18,12 +21,17 @@
     private final Archive<?> archive;
 
     @Inject
```
kenfinnigan, line 22: I think this can be modified to:

`@Inject
private OpenTracingFraction fraction`

as we're inside that fraction so it will always be present for this code

kenfinnigan (changes requested):
Is it possible to also have a test that verifies the yaml configuration?

jpkrohling:
> Is it possible to also have a test that verifies the yaml configuration?

Is it necessary to have this test if we have the one using YAML?

jpkrohling:
I added another commit to this PR, changing the test to use the configuration file instead of setting the property directly into the fraction. Unless I'm missing something, I don't think the old test is needed anymore. If you agree, I can squash the commits and update the PR.

testsuite/testsuite-opentracing/src/test/java/org/wildfly/swarm/opentracing/SkipPatternTest.java
```diff
@@ -42,7 +40,7 @@
  * @author Juraci Paixão Kröhling
  */
 @RunWith(Arquillian.class)
-public class OpenTracingWithSkipPatternTest {
+public class SkipPatternTest {
```
jpkrohling: I changed the name of the tests to be more concise. If it's not appropriate for this PR, I can revert it.

kenfinnigan (approved):

PR #631, opened 2017-09-07
SWARM-1539 - Accomodate port offset.

bobmcwhirter:
Motivation
----------
usage.txt needs to be able to display the actual port in use,
not just bare swarm.http.port.

Modifications
-------------
In general move away from ${swarm.port.offset:0} format for expressions
resolved by WF into forcing our own YAML/configurable to handle it prior
because we need it resolved earlier for usage.txt.

Now support:

${swarm.INTERFACE_NAME.host} such as ${swarm.public.host} for learning
of hostnames for interfaces.

${swarm.BINDING.port}, ${swarm.BINDING.multicast-port} and ${swarm.BINDING.multicast-address}
for general socket-bindings etc, taking into account the port-offset.

Just so happens that makes ${swarm.http.port} tell you the correctly-offsetted
port.

Additionally, for ease of use, ${swarm.public.url.base} will resolve the same
as:

    http://${swarm.public.host}:${swarm.http.port}/

It's the only friendly short-hand we've got.

Result
------
See above.

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/NetworkVariableSupplier.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/NetworkVariableSupplier.java
@@ -1,0 +1,94 @@
+package org.wildfly.swarm.container.runtime.usage;
+
+import org.wildfly.swarm.container.Interface;
+import org.wildfly.swarm.spi.api.SocketBinding;
+import org.wildfly.swarm.spi.api.SocketBindingGroup;
+
+/**
+ * Created by bob on 9/7/17.
+ */
+public class NetworkVariableSupplier implements UsageVariableSupplier {
+
+    public NetworkVariableSupplier(Iterable<Interface> interfaces, Iterable<SocketBindingGroup> socketBindings, UsageVariableSupplier delegate) {
+        this.interfaces = interfaces;
+        this.socketBindings = socketBindings;
+        this.delegate = delegate;
+    }
+
+    @Override
+    public Object valueOf(String name) throws Exception {
+        // very special case
+        if (name.equals("swarm.public.url.base")) {
+            return "http://" + valueOf("swarm.public.host") + ":" + valueOf("swarm.http.port") + "/";
+        }
+
+        String[] parts = name.split("\\.");
+        if (parts.length > 0 && parts[0].equals("swarm")) {
+            if (parts.length == 3) {
+                for (Interface each : this.interfaces) {
+                    if (parts[1].equals(each.getName())) {
+                        if (parts[2].equals("host")) {
+                            return each.getExpression();
+                        }
+                    }
+                }
+            }
+
+            Object value = null;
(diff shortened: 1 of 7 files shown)


gastaldi (dismissed):

core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/NetworkVariableSupplier.java
```diff
@@ -0,0 +1,92 @@
...
+        String[] parts = name.split("\\.");
+        if (parts.length == 3) {
+            for (Interface each : this.interfaces) {
+                if (parts[1].equals(each.getName())) {
+                    if (parts[2].equals("host")) {
+                        return each.getExpression();
+                    }
+                }
+            }
+        }
```
Ladicek: Is it guaranteed that `parts[0]` is always `"swarm"`?

Ladicek:
In general, I'm a bit concerned about resolving some parts of configuration in Swarm instead of leaving it to WildFly, because then WildFly won't know that it's actually supposed to be an expression. But I guess it's OK since we're by no means supporting managed domain, and nothing else comes to mind that might be affected.

Well maybe setting system properties via WildFly management... but I guess noone does that with Swarm. Not sure if that's even possible.

core/container/src/main/java/org/wildfly/swarm/container/runtime/usage/NetworkVariableSupplier.java
bobmcwhirter: Ah, we should probably confirm that it is.

bobmcwhirter:
@Ladicek fwiw, 98% of our cases we're controlling anything that had been previously an expression with our Configurable stuff (YAML, props etc).  And our method even makes *everything* expression-like instead of only the subset supported by WF itself.

I think we've explicitly stopped caring about domain mode or anything terribly mutable via management interfaces.

Ladicek:
Sounds good. I could also think of `standalone.xml` that we're I think still supporting (?), but I guess these are all corner cases.

bobmcwhirter:
@Ladicek Yeah, if someone is using `standalone.xml` then we are kinda screwed.  I guess we could instead of feeding these values from the YAML configurables, we could scrub around in the DMR, which is after-the-fact, except for any ${expr} used within a `standalone.xml` maybe.

If it turns into an issue, we can also scrub around in the `Service<T>` for socket-bindings after WF has booted, and avoid having to math ourselves, and it would represent the *actual true bonafide values*.

But not in this PR.

bobmcwhirter:
Actually, @Ladicek convinced me.  I'm going to reformulate this in terms of WF runtime `Service<T>`s to avoid corners/edges and sources of confusion and bugs.  Added a _HOLD_ label.  

DO NOT MERGE.

bobmcwhirter:
We only currently process standalone.xmls for subsystems, so not going to pursue reformulating this currently.  MERGE ON!

kenfinnigan (approved):

PR #633, opened 2017-09-11
SWARM-1497 - Avoid removing non-temporal jar files

juangon:
Motivation
----------
As part of the fix for SWARM-1327, some non temporal jar files that are being exploded are being removed, which is wrong. Only exploded jar files should be removed.

Modifications
-------------
Fix code in order to not remove original unexploded jar files that are going to be exploded later.

Result
------
User jar files aren't being removed.

-----

### DIFF
--- a/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/NestedJarResourceLoader.java
+++ b/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/NestedJarResourceLoader.java
@@ -154,7 +154,6 @@ public class NestedJarResourceLoader {
                     BootstrapUtil.explodeJar(jarFile, tmpDir.getAbsolutePath());
 
                     jarFile.close();
-                    file.delete();
 
                     return ResourceLoaders.createFileResourceLoader(loaderName, tmpDir);
                 } else {
@@ -171,7 +170,6 @@ public class NestedJarResourceLoader {
                 BootstrapUtil.explodeJar(jarFile, tmpDir.getAbsolutePath());
 
                 jarFile.close();
-                file.delete();
 
                 return ResourceLoaders.createFileResourceLoader(loaderName, tmpDir);
             }


vveloso (commented):
If it is safe to leave the original file in its place, why explode it at all? SWARM-1327 introduced jar file explosion to allow deletion of the original jar. So if we're no longer deleting the jar why is jar explosion still necessary?

vveloso (changes requested):
org.wildfly.swarm.bootstrap.modules.ApplicationModuleFinder#addAsset was also changed by SWARM-1327 to explode and delete original jars. Should it be updated as well?

juangon:
@vveloso the problem is we can't use actual jar file without being exploded, as Windows doesn't release them. This applies to ApplicationModuleFinder mainly, as NestedJarResourceLoader doesn't use temporal jar files (ApplicationModuleFinder does, which causes those to not be removed). I wrongly deleted them in NestedJarResourceLoader, but as this case isn't the same as ApplicationModuleFinder, this isn't needed at all. About your 2nd question, ApplicationModuleFinder seems fine, as there it uses temporal jar files.

vveloso:
@juangon if I understand correctly, NestedJarResourceLoader does not use temporary jar files but it still explodes loaded jars because it may be used to load resources from temporary files created elsewhere. Is this correct?

Also, is the locked file restriction you mentioned in the context of SWARM-1327? Is that a problem only in a test environment? If so I'd argue that it does not make sense to spend resources (CPU time, I/O and storage) to explode jars in production. What about a feature switch, disabled by default, to enable jar explosion?

juangon:
@vveloso Not exactly. NestedJarResourceLoader explodes the jar files they are loading (nested in this case).
Yes, that problem arised in the context of SWARM-1327 for Windows env, but as we don't want to make any source difference between environments, we let as is. About performance, I doubt that, in Microservices context, there would be extremely huge jar files that would cause a lot of resources and HDD memory to explode it. This improvement had been there for a while and no complaint since then.

juangon:
@vveloso BTW SWARM-1327 isn't a problem in test environment, but in Windows environment.

vveloso:
@juangon I understand your point. That makes Swarm ill-suited for simplifying deployment of larger services (not micro. mind you) with some thousands of Java classes organized in modules with a few hundred jars. Startup is delayed by 2 minutes due to jar explosion alone on SSDs. Thank you for your input.

kenfinnigan (approved):

PR #640, opened 2017-10-02
SWARM-1542 Run Cassandra,MongoDB,Neo4j,OrientDB test with docker via mvn -Pdocker clean install verify

scottmarlow:
"-Pdocker verify" doesn't work from root yet, due to "verify errors that seem out of scope for this change (RedundantImport imports from logger https://paste.fedoraproject.org/paste/8arldovKmUrxpJD-M4pCyA).  Also stopped using @DefaultDeployment, to workaround failure with beans.xml not being loaded with integration tests (not sure why).


scottmarlow:
Currently working, is that we can change into each of the NoSQL testsuite folders and do a "mvn -Pdocker clean install verify" and expect that to pass.  I'm thinking that we can do this in the wildfly-swarm-pull-request-linux job.

bobmcwhirter:
Per IRC...

This can work (for some small value of "work") on my Mac if you change the `<wait>` section to include:

    <wait>
      <tcp>
          <mode>mapped</mode>
          <ports>
            <port>27017</port>
         </ports>
      </tcp>
    </wait>

With the `<mode>mapped</mode>` being the important tidbit.

bobmcwhirter:
With the above-mentioned change, both OrientDB and Mongo give me the same `MALFORMED` deployment error, along with OrientDB failing it's actual tests (but Mongo seems to pass other than that)

https://gist.github.com/bobmcwhirter/a5f83c4634b0fbd67374c5179fc38495

bobmcwhirter:
Also seems the failsafe-plugin runs twice for each testsuite:

    [INFO] --- maven-failsafe-plugin:2.19.1:integration-test (integration-test) @ testsuite-neo4j ---

and

    [INFO] --- maven-failsafe-plugin:2.19.1:integration-test (default) @ testsuite-neo4j ---

The differences being the executions;`integration-test` which results in the `MALFORMED` error, and `default` which may or may not pass (on my Mac).

Neo4J doesn't quite fail, but lots of NPEs on the console:

https://gist.github.com/bobmcwhirter/eea36baa8e23491d9cc18694f677f252

testsuite/testsuite-cassandra/pom.xml
```diff
@@ -65,4 +65,75 @@
...
+                <image>
+                  <name>cassandra:3.0</name>
+                  <run>
+                    <ports>
+                      <port>9042:9042</port>
+                    </ports>
+                    <wait>
+                        <time>120000</time>
+                        <tcp>
+                            <ports>
```
bobmcwhirter: Add `<mode>mapped</mode>`

testsuite/testsuite-mongodb/pom.xml
```diff
@@ -69,4 +69,102 @@
...
+                    </ports>
+                    <wait>
+                        <time>120000</time>
+                        <!-- TODO: get MongoDB db.createUser to work for authentication testing
+                        <exec>
+                            <postStart>mongo mongotestdb "db.createUser({ user: 'devuser', pwd:  'changethis', roles: ['dbAdmin'] }); db.auth('devuser', 'changethis'); "</postStart>
+                        </exec>
+                        -->
+                        <tcp>
+                            <ports>
```
bobmcwhirter: Add `<mode>mapped</mode>`

testsuite/testsuite-neo4j/pom.xml
```diff
@@ -65,4 +65,85 @@
...
+                    </env>
+                    <privileged>true</privileged>
+                    <ports>
+                      <port>7687:7687</port>
+                      <port>7474:7474</port>
+                    </ports>
+                    <wait>
+                        <time>120000</time>
+                        <tcp>
+                            <ports>
```
bobmcwhirter, line 105: Add `<mode>mapped</mode>`

testsuite/testsuite-orientdb/pom.xml
```diff
@@ -103,4 +103,82 @@
...
+                    <env>
+                        <ORIENTDB_ROOT_PASSWORD>admin</ORIENTDB_ROOT_PASSWORD>
+                    </env>
+                    <ports>
+                      <port>2424:2424</port>
+                    </ports>
+                    <wait>
+                        <time>120000</time>
+                        <tcp>
+                            <ports>
```
bobmcwhirter, line 142: Add `<mode>mapped</mode>`

bobmcwhirter:
So, the `pom.xml` one level up includes the `maven-failsafe-plugin` and a non-`default` execution defined, which might be what's causing some of the double-running I'm seeing. 

https://github.com/wildfly-swarm/wildfly-swarm/blob/master/testsuite/pom.xml#L80-L87

scottmarlow:
Instead of merging this pr, switch to using Arquillian Cube.

PR #648, opened 2017-10-12
ENTSWM-186 - Document BOM usage.

bobmcwhirter:
Motivation
----------
Document the usage of BOMs from both upstream and downstream
perspectives, using conditional blocks.

Modifications
-------------
Provide verbiage on both community and product flavors.

Result
------
Resulting doc is controlled by -Dswarm.product.build as usual.

-----

### DIFF
--- a/docs/howto/use-a-bom/index.adoc
+++ b/docs/howto/use-a-bom/index.adoc
@@ -7,6 +7,8 @@ BOMs (bill of materials) which you may use to avoid
 having to track and update Maven artifact versions
 in several places.
 
+ifndef::product[]
+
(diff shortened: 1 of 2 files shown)


tradej (approved):
LGTM.

Having said that, I am not sure if `:product:` is the best attribute name to use, as some products use it for product name. In BxMS, where I used to work before, we would use `PRODUCT_ONLY` and `COMMUNITY_ONLY` as complements.

RHOAR uses `:ProductName:`, so if you insist on using `:product:`, we should generally be OK.

What do you think?

kenfinnigan:
Is this still hold?

PR #650, opened 2017-10-13
ENTSWM-171: Add fractions tags to ribbon-secured-client.

mkouba:
Motivation
----------
ribbon-secured-client is not included in bom-certified. In fact,
it is not considered a fraction and so it's completely ignored by
fraction registry, therefore it's not added to the fraction-list.txt
and ignored by CertifiedBomMojo.

Modifications
-------------
Added "Service Discovery,Netflix OSS,Security" fraction tags. Also set
the stability level to STABLE.

Result
------
ribbon-secured-client will be listed in bom-certified.

### DIFF
--- a/fractions/netflix/ribbon-secured-client/pom.xml
+++ b/fractions/netflix/ribbon-secured-client/pom.xml
@@ -18,7 +18,7 @@
   <artifactId>ribbon-secured-client</artifactId>
 
   <name> Ribbon Secured Client</name>
-  <description> Ribbon Secured Client</description>
+  <description>Ribbon Secured Client</description>
 
   <packaging>jar</packaging>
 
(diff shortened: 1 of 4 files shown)


kenfinnigan (dismissed):

kenfinnigan:
retest this please

kenfinnigan:
retest this please

bobmcwhirter:
I think we also have a pom.xml property to force inclusion in a BOM regardless of if its fraction-like or not.  I could be wrong though.

mkouba:
Bob is right, it seems there is `swarm.bom` property! I'm going to try it out and send a new PR if it works.

mkouba:
Unfortunately, `swarm.bom` does not work here because `CertifiedBomMojo` loads data from `fraction-list.txt`:
https://github.com/wildfly-swarm/wildfly-swarm-fraction-plugin/blob/master/src/main/java/org/wildfly/swarm/plugin/bom/CertifiedBomMojo.java#L81

kenfinnigan (approved):

PR #655, opened 2017-10-15
[SWARM-1597] Topology OpenShift: advertise cluster IP

tadamski:
Shouldn't OpenShift Topology plugin advertise clusterIP? Using currently advertised data listeners are unable to talk to the service that is being discovered.


Ladicek:
I think we already had this change in and had to take it back out, see SWARM-1333

tadamski:
Thanks. I made my comments there. Can you please check if they makes any sense to you :)

heiko-braun:
We had this change in place, but it was reverted. IIRC the argument was that internally, the service name works as the lookup and their would not be a reason for accessing the IP. I.e within Openshift you can call `http://greeting` for a service that's named `greeting`

heiko-braun:
test this please

heiko-braun:
I'd suggest we close this one because it didn't really move forward.

PR #656, opened 2017-10-16
Move to a single assembly for upstream build.

bobmcwhirter:
Assembles the modules from docs/index.adoc into
a single doctype=book output.

-----

### DIFF
--- a/docs/howto/create-a-datasource/index.adoc
+++ b/docs/howto/create-a-datasource/index.adoc
@@ -41,7 +41,7 @@ scope) to your application:
 [source,xml]
 ----
 <dependencies>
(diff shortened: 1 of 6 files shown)


tradej (approved):
In the context of this pull request, I have no objections.

However, I have noticed that the items in various lists are indented wrongly (especially in procedures). This can be fixed using open blocks, for example:
```
. First list element with some text
+
--
Some more text.

[source]
----
a code example
----

Some more text still.
--

. Second list element.
```

In that example, everything enclosed in double dashes will be appended to the first list element.

Will you file an issue about the list elements, or should I?

PR #689, opened 2017-11-10
Mentions of hollow JARs Conditional

tradej:
Made the mentions of hollow JARs conditional. Needed for fabric8-launch/appdev-documentation#823.


kenfinnigan:
@tradej the conditionals are not something I'm familiar with.

Do they mean that the hollow jar doc IS included in product or IS NOT included? The conditional reads like the former, but I think it's the latter

tradej:
@kenfinnigan The `ifndef` directive means that whatever is surrounded by it is shown if the attribute in question is **not** defined, i. e. in this case, the hollow JAR is **not** shown in product builds.

Ordinarily, I would use `ifdef::community[]` or some such variant, but we only use the `:product:` attribute at the moment.

kenfinnigan (approved):

PR #697, opened 2017-11-20
Fix and enable Microprofile TCKS within testsuite

heiko-braun:
This fixes assorted minor problems and allows to run the microprofile TCK's from the main testsuite


heiko-braun:
retest this please

kenfinnigan:
Does this enable the TCK for each MP spec as part of the normal build?

Ideally we want to have that as a separate profile that can be activated for PRs, and a separate PR job, but not on every CI build

heiko-braun:
No, this just allows to run the tck's at all. Without these fixes it didn't work. 

What you describe is a separate jira i think.

kenfinnigan (approved):

PR #700, opened 2017-11-20
SWARM-1635, Part 1: Added HealthFraction detector

heiko-braun:


fractions/microprofile/microprofile-health/src/main/java/org/wildfly/swarm/health/detect/HealthPackageDetector.java
```diff
@@ -0,0 +1,33 @@
+/**
+ * Copyright 2015-2016 Red Hat, Inc, and individual contributors.
```
kenfinnigan: Should the copyright be 2017?

juangon: Wouldn't be possible to say 2015-present? This is what they do in my previous employer, and it avoids updating thousands of source code files each year.

kenfinnigan (approved):

PR #706, opened 2017-11-24
SWARM-1691 | Microprofile OpenTracing fraction

pavolloffay:
Swam jira: https://issues.jboss.org/browse/SWARM-1691
Umbrella task (contains resteasy tasks): https://issues.jboss.org/browse/MCRPRO-9

It's not done yet. I am sharing this to address remaining issues:
- [x] Async handlers seems to not work https://pastebin.com/22QBg0yU
- [x] CDI providers seems to work in testsuite but not in an example app see https://pastebin.com/PcjB8hSu
- [x] Remove `MockTracer` from `TracerProducer`
- [x] (resteasy) Client tracing feature/filter is not automatically registered. The solution might be to use extend `ResteasyClientBuilder` and register it via service loader so far without a luck for me. I need a help here! Other solution might be to use `ResteasyProviderFactory.getInstance().getClientRequestFilters()` but it does not seem to work either (see http://lists.jboss.org/pipermail/resteasy-dev/2017-November/000509.html). - **Fixed via** https://github.com/wildfly/wildfly-core/pull/3082
- [x] (resteasy) TCK `testException` - runtime exception is thrown from handler, response code in filter is 200 and should be 500. - **TEMPORAL FIX**: in TCK deployment I have added exception mapper for runtime exception.

If somebody wants to try it here is example https://github.com/wildfly-swarm/wildfly-swarm-examples/pull/162

### DIFF
--- a/fractions/microprofile/microprofile-opentracing/pom.xml
+++ b/fractions/microprofile/microprofile-opentracing/pom.xml
@@ -1,0 +1,138 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.wildfly.swarm</groupId>
+    <artifactId>build-parent</artifactId>
+    <version>2018.6.0-SNAPSHOT</version>
+    <relativePath>../../../build-parent/pom.xml</relativePath>
+  </parent>
+
+  <artifactId>microprofile-opentracing</artifactId>
+  <packaging>jar</packaging>
+
+  <name>Microprofile-OpenTracing</name>
+  <description>OpenTracing integration for Microprofile</description>
+
+  <properties>
+    <swarm.fraction.stability>experimental</swarm.fraction.stability>
+    <swarm.fraction.tags>Microprofile, OpenTracing, Monitoring</swarm.fraction.tags>
+  </properties>
+
+  <build>
+    <resources>
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>true</filtering>
+      </resource>
+    </resources>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.wildfly.swarm</groupId>
+      <artifactId>spi</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.wildfly.swarm</groupId>
+      <artifactId>meta-spi</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.wildfly.swarm</groupId>
+      <artifactId>container</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.wildfly.swarm</groupId>
+      <artifactId>jaxrs</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.wildfly.swarm</groupId>
+      <artifactId>cdi</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-client</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <!-- OpenTracing -->
+    <dependency>
+      <groupId>org.eclipse.microprofile.opentracing</groupId>
+      <artifactId>microprofile-opentracing-api</artifactId>
+      <version>${version.microprofile-opentracing}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.opentracing</groupId>
+      <artifactId>opentracing-api</artifactId>
+      <version>${version.opentracing}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.opentracing</groupId>
+      <artifactId>opentracing-util</artifactId>
+      <version>${version.opentracing}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.opentracing</groupId>
+      <artifactId>opentracing-noop</artifactId>
+      <version>${version.opentracing}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.opentracing.contrib</groupId>
+      <artifactId>opentracing-tracerresolver</artifactId>
+      <version>${version.opentracing.tracerresolver}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.opentracing.contrib</groupId>
+      <artifactId>opentracing-web-servlet-filter</artifactId>
+      <version>${version.opentracing.servlet}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.opentracing.contrib</groupId>
+      <artifactId>opentracing-jaxrs2</artifactId>
+      <version>${version.opentracing.jaxrs}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.wildfly.core</groupId>
+      <artifactId>wildfly-core-feature-pack</artifactId>
+      <type>zip</type>
+      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.wildfly</groupId>
+      <artifactId>wildfly-feature-pack</artifactId>
+      <type>zip</type>
+      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.wildfly</groupId>
+      <artifactId>wildfly-servlet-feature-pack</artifactId>
+      <type>zip</type>
+      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+  </dependencies>
+</project>
--- a/fractions/microprofile/microprofile-opentracing/src/main/java/org/wildfly/swarm/mpopentracing/deployment/MicroprofileServerTracingFeature.java
+++ b/fractions/microprofile/microprofile-opentracing/src/main/java/org/wildfly/swarm/mpopentracing/deployment/MicroprofileServerTracingFeature.java
@@ -1,0 +1,50 @@
+/**
+ * Copyright 2015-2018 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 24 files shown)


kenfinnigan:
add to whitelist

pavolloffay:
I have added testsuite to run MP-OpenTracing TCK. Strange thing is that tests works but example app (linked in description) fails on https://pastebin.com/Z5514Pyu 

```
 Caused by: org.jboss.as.server.deployment.DeploymentUnitProcessingException: WFLYWELD0021: Service class org.wildfly.swarm.mpopentracing.deployment.OpenTracingCDIExtension didn't implement the javax.enterprise.inject.spi.Extension interface"
```

cc @mkouba @kenfinnigan @pilhuhn

scottmarlow:
[docker test](https://ci.wildfly-swarm.io/blue/organizations/jenkins/wildfly-swarm-docker-test-linux/detail/wildfly-swarm-docker-test-linux/258/pipeline) is failing the './mvnw -B -U clean install -DskipTests=true -e' due to [error](https://paste.fedoraproject.org/paste/-c63M9v6BkN3X-3NKdFe4w)

pavolloffay:
@scottmarlow It's WIP and as the first comment says it depends on snapshots.

fractions/microprofile/microprofile-opentracing/src/main/java/org/wildfly/swarm/mpopentracing/deployment/MicroprofileServerTracingFeature.java
```diff
@@ -0,0 +1,32 @@
...
+ * @author Pavol Loffay
+ */
+@Provider
+public class MicroprofileServerTracingFeature implements DynamicFeature {
+
+  private ServerTracingDynamicFeature delegate;
+
+  public MicroprofileServerTracingFeature() {
+    // TODO tracer CDI injection didn't work
+    this.delegate = new Builder(GlobalTracer.get())
```
pavolloffay: @mkouba do you think we can inject the tracer via CDI?

I have tried but it didn't work. Tracer was always null. I have tried constructor and field injection.

kenfinnigan:
@pavolloffay What's the state of this?

It needs a rebase because of a pom.xml change, but beyond that how does it stand?

pavolloffay:
Client tracing does not work, the traced builder is not used at runtime. The second issuse - in an example app it says that cdi extension does not implement extension interface. And also the third issue is that async endpoints do not work.

fractions/microprofile/microprofile-opentracing/src/main/java/org/wildfly/swarm/mpopentracing/deployment/MicroprofileServerTracingFeature.java
mkouba: I have no idea how `DynamicFeature` works and whether JAX-RS supports injection there. If there is a bean for  `GlobalTracer` then `CDI.current().select(GlobalTracer.class).get()` should work but using `CDI.current()` might be tricky in Java EE.

bobmcwhirter:
retest this please

bobmcwhirter:
retest this please

bobmcwhirter:
retest this please

pavolloffay:
@mkouba hi, could you please look at CDI issue described in the PR description?

fractions/microprofile/microprofile-opentracing/src/main/java/org/wildfly/swarm/mpopentracing/deployment/MicroprofileServerTracingFeature.java
pavolloffay: There should be `Tracer.class` bean, I will try that, although is it safe to use?

mkouba:
@pavolloffay Will definitely do early next year. I had few other tasks with higher priority in the queue..

pavolloffay:
>  Async handlers seems to not work

This is also related to https://issues.jboss.org/browse/SWARM-1695. Fraction has to add a filter which runs for every request. 

Currently the filter is added as:
```java
    servletContext
        .addFilter("tracingFilter", new SpanFinishingFilter(tracer))
        .addMappingForUrlPatterns(EnumSet.of(DispatcherType.ASYNC, DispatcherType.REQUEST), false, "*");
```

However it still throws an exception `IllegalStateException: UT010026: Async is not supported for this request`. We cannot use only `DispatcherType.ASYNC` because it has to be executed for every request. 

The question is how can fraction add a filter which is executed for every request? cc @kenfinnigan @bobmcwhirter

kenfinnigan:
In addition to setting both dispatcher types, it might be necessary to set `asyncSupported` on the Filter?

Not sure how to do it programmatically, but annotation has it: https://docs.oracle.com/javaee/7/api/javax/servlet/annotation/WebFilter.html

pavolloffay:
@kenfinnigan I have the solution. It was simple, although by googling it I could not find any references to it.

```java
    Dynamic filterRegistration = servletContext
        .addFilter("tracingFilter", new SpanFinishingFilter(tracer));
    filterRegistration.setAsyncSupported(true);
```

kenfinnigan:
So what I suggested then?

pavolloffay:
You suggested the right thing, it seems that all swarm issues are sorted out. TCK now looks like this

```
Tests run: 15, Failures: 5, Errors: 0, Skipped: 0
```

These 5 are resteasy related - 1.client features registration, 2. wrong status code if an exception is thrown

kenfinnigan:
Do those 5 failures require fixes in RESTEasy to be resolved?

pavolloffay:
Probably yes, I could not find any solution. @liweinan and  @asoldano are looking into this.

Related issues are https://issues.jboss.org/browse/RESTEASY-1758 and https://issues.jboss.org/browse/RESTEASY-1759

pavolloffay:
@mkouba I need more your help with OpenTracing TCK. 

Tests are using just tracer API not a specific tracer implementation. Then there is an endpoint which returns reported data back to test for validation. Endpoint uses reflection to access data in the tracer and it assumes that it is `MockTracer`. Deployment in TCK is here https://github.com/eclipse/microprofile-opentracing/blob/master/tck/src/main/java/org/eclipse/microprofile/opentracing/tck/OpentracingClientTests.java#L88 - it's not adding `MockTracer`. We somehow have to add `MockTracer` to the deployment. Not only the tracer but tracer-resolver SPI binding. Other TCKs are using arquillian extensions to change things. Could be this approach used in this scenario?

In swarm we use service loader to provide tracer instance. The app developer just pulls in mp-opentracing fraction and adds dependency like this  https://github.com/jaegertracing/jaeger-client-java/tree/master/jaeger-tracerresolver. MP does not define how tracer if configured it just says that it is available to the app via CDI.

mkouba:
@pavolloffay Hm, I'm not sure I really understand how the test should work. `Tracer` does not seem to be part of MP opentracing API, is it? Which endpoint expects `MockTracer`?

In any case - yes, you can use Arquillian SPI to enhance the test archive. See for example https://github.com/wildfly-swarm/wildfly-swarm/blob/master/testsuite/microprofile-tcks/config/src/test/java/org/wildfly/swarm/microprofile/config/tck/ConfigAuxiliaryArchiveAppender.java.

fractions/microprofile/microprofile-opentracing/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+      <groupId>org.wildfly.swarm</groupId>
+      <artifactId>microprofile</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-client</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
```
kenfinnigan: Remove `wildfly-jaxrs`, `microprofile`and `javaee-api` as dependencies to the fraction

kenfinnigan:
@pavolloffay, I debugged through to see that the `TracedClientBuilder` is being instantiated, but fails because `CDI.current() returns null.

There are some dependencies in `pom.xml` that can be cleaned up, and can you rebase this PR on latest master to remove conflicts. Thanks

fractions/microprofile/microprofile-opentracing/pom.xml
pavolloffay: When I remove dependency on `microprofile` TCK fails on 


```
java.lang.NoClassDefFoundError: org/apache/commons/logging/LogFactory
	at org.eclipse.microprofile.opentracing.tck.OpentracingClientTests.clearTracer(OpentracingClientTests.java:1075)
	at org.eclipse.microprofile.opentracing.tck.OpentracingClientTests.beforeEachTest(OpentracingClientTests.java:116)
Caused by: java.lang.ClassNotFoundException: org.apache.commons.logging.LogFactory
	at org.eclipse.microprofile.opentracing.tck.OpentracingClientTests.clearTracer(OpentracingClientTests.java:1075)
	at org.eclipse.microprofile.opentracing.tck.OpentracingClientTests.beforeEachTest(OpentracingClientTests.java:116)
```

kenfinnigan: So `commons-logging` or the like needs to be a dependency, maybe. And it looks like it's needed for the test more than the actual fraction.

`microprofile` refers to the entire collection of MP fractions in WF Swarm, so should not be used within a single MP fraction otherwise every other MP fraction will be included in open tracing

pavolloffay: I solved it, I have added commons-logging to the test module

pavolloffay:
Rebased and updated.

> I debugged through to see that the TracedClientBuilder is being instantiated, but fails because `CDI.current() returns null.

it is instantiated only in tests. In the app/deployment the TracedClientBuilder is not being used. I have changed CDI to get the tracer via static field. Now it does not fail but the `TracedClientBuilder` is not used in deployment.

kenfinnigan:
What example are you using to test in a real deployment?

So it's fixed in the TCK?

pavolloffay:
The example is in the first comment wildfly-swarm/wildfly-swarm-examples#162. No it's not solved in TCK as the deployment is not using traced client.

pavolloffay:
A short update here with: wildfly/wildfly-core#3082 100% TCK pass.

stuartwdouglas:
We have had to revert https://github.com/wildfly/wildfly-core/pull/3082 . I have attempted to work around the issue a different way using https://github.com/wildfly/wildfly/pull/10933 however I am not sure how to test it to make sure it resolves your issue.

kenfinnigan:
@stuartwdouglas, would that fix be possible in WF 11?

To verify it would require a local WF SNAPSHOT build and then updating the /build-parent/pom.xml to use the new WF version in the build. Then a `mvn clean install` on this PR will test existing stuff and this new PR.

stuartwdouglas:
Assuming it works then it should be possible to do something similar with Wildfly 11 using a DeploymentUnitProcessor (I assume your fractions can register additional DUP's?). You would need to install the DUP after JaxrsDependencyProcessor, and if a JAX-RS client is detected then iterate the module dependencies and add the appropriate filters to the dependency.

stuartwdouglas:
I attempted to test this but it looks like the EE8 changes are causing some issues for swarm. I managed to get the fraction plugin working with https://github.com/wildfly-swarm/wildfly-swarm-fraction-plugin/pull/69 but then further changes were required (I got as far as https://github.com/stuartwdouglas/wildfly-swarm-core/commit/6fe3d4e7cfe6c9493931efba0405716d8d7bca07 then gave up, mostly due to frustration with the intellij maven import now working correctly).

kenfinnigan:
@stuartwdouglas, we're not able to update to WF 12 at present, so that solution wouldn't work for us anyway unfortunately.

kenfinnigan:
test this please

mkouba:
Please send a new PR once the [migration to a new repo is over](https://issues.jboss.org/browse/THORN-2031).

PR #714, opened 2017-11-29
SWARM-1619: BOM should not include internal fractions

kenfinnigan:

-----

Motivation
----------
BOM currently includes fractions that are marked internal.

Users should not be adding internal fractions as dependencies to their applications.

Modifications
-------------
Update fraction-plugin to version that filters out internal fractions from BOM creation.

Result
------
BOM no longer includes internal fractions

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -22,7 +22,7 @@
   <description>Build Parent to bring in required dependencies</description>
 
   <properties>
-    <version.wildfly.swarm.fraction.plugin>76</version.wildfly.swarm.fraction.plugin>
+    <version.wildfly.swarm.fraction.plugin>77</version.wildfly.swarm.fraction.plugin>
 
     <version.org.snakeyaml>1.17</version.org.snakeyaml>
 
--- a/testsuite/testsuite-maven-plugin/src/test/java/org/wildfly/swarm/maven/plugin/TestingProject.java
+++ b/testsuite/testsuite-maven-plugin/src/test/java/org/wildfly/swarm/maven/plugin/TestingProject.java
@@ -84,7 +84,7 @@ public final class TestingProject {
         String pomXmlContent = new String(Files.readAllBytes(pomXml), StandardCharsets.UTF_8)
                 .replace("<!--PLACEHOLDER:swarm-version-->", swarmVersion)
                 .replace("<!--PLACEHOLDER:packaging-->", packaging.packagingType())
-                .replace("<!--PLACEHOLDER:dependencies-->", dependenciesSnippet())
+                .replace("<!--PLACEHOLDER:dependencies-->", dependenciesSnippet(swarmVersion))
                 .replace("<!--PLACEHOLDER:configuration-->", swarmPluginConfigurationSnippet());
         Files.write(pomXml, pomXmlContent.getBytes(StandardCharsets.UTF_8));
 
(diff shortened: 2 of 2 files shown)


Ladicek (dismissed):

kenfinnigan:
retest this please

kenfinnigan:
retest this please

kenfinnigan:
@Ladicek any insight you can provide as to why testsuite-maven-plugin keeps failing with this PR would be appreciated.

Had a look at the test but couldn't see anything that would cause it not to find SERVLET or JAX-RS

Ladicek:
@kenfinnigan Looking at https://ci.wildfly-swarm.io/job/wildfly-swarm-pull-request-linux/800/org.wildfly.swarm.testsuite$testsuite-maven-plugin/testReport/, it seems all the 3 failures are the same: when the test project is being built, Maven can't find a version for `org.wildfly.swarm:container`. That used to be a public API I think; is it no longer the case?

```
Exit code was non-zero: 1; command line and log = 
/opt/maven/apache-maven-3.3.9/bin/mvn -s /root/.m2/settings.xml -e --batch-mode -Dmaven.repo.local=/root/workspace/wildfly-swarm-pull-request-linux/.repository org.apache.maven.plugins:maven-clean-plugin:clean verify
[INFO] Error stacktraces are turned on.
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for org.wildfly.swarm:container:jar is missing. @ line 37, column 17
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
org.apache.maven.project.ProjectBuildingException: Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for org.wildfly.swarm:container:jar is missing. @ line 37, column 17
```

kenfinnigan:
From looking at the BOMs, yes `container` is no longer present in the BOM, but `container` will still be a dependency of other Fractions.

Maybe I need to rework the test a bit

Ladicek:
So all the tests that are failing have a custom `main` method (aside: do we still wanna test that? I guess yes, for now), so they have an explicit dependency on `org.wildfly.swarm:container` right in their POM. I don't know how importing a `<dependencyManagement>` section from a BOM works, but I guess this doesn't work transitively?

kenfinnigan:
@Ladicek tests work with this fix.

We can evaluate another time the validity of testing main() usage

Ladicek:
I don't quite like the test change -- it bypasses the BOM, which is how Swarm is supposed to be consumed. At this point, I believe the BOM should include `container` -- it's fraction-ish for the purpose of generating the BOM, but not really a fraction, so the notion of "being internal" doesn't quite apply. I think.

kenfinnigan:
Right, but `container` is always present as a transitive dependency through user facing fractions.

It's adding it directly that's the problem as users should not be adding that themselves.

Now I'm happy to remove the main() tests so that such a dependency isn't needed at all

Ladicek:
I see. Makes sense. I filed https://issues.jboss.org/browse/SWARM-1709 and I'm now +1 to merge this.

Ladicek:
Ah silly me -- I thought the fix you did bypasses the BOM for _all_ fractions -- but it's only for `container`. So it's actually fine.

Ladicek (approved):

PR #719, opened 2017-12-01
SWARM-1644: Make MP TCKs work.

mkouba:
Motivation
----------
We need to setup a proper TCK runners for all MP specs. Each MP spec should
have a dedicated configuration. Note that MP health TCK currently
works only by luck (there are no test running in container mode).

Modifications
-------------
Added maven module for each MP spec and "microprofile-tck" profile in
the top-level pom.xml. This profile is not activated by default but
could be activated by means of "swarm.microprofile.tck" system property.

Result
------
It's possible to run all MP TCKs, TCK for a separate MP spec or even a
single test/method from a specific TCK.

## NOTE: metrics and fault tolerance TCKs contain a lot of failures. I'm going to send a separate PRs with fixes for both specs.


testsuite/microprofile-tcks/pom.xml
```diff
@@ -24,224 +24,74 @@
...
     <modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <groupId>org.wildfly.swarm.testsuite</groupId>
-        <artifactId>testsuite-parent</artifactId>
-        <version>2017.12.0-SNAPSHOT</version>
-        <relativePath>../</relativePath>
-    </parent>
+   <parent>
+      <groupId>org.wildfly.swarm</groupId>
```
kenfinnigan, line 28: any reason we inherit from build-parent and not testsuite-parent?

kenfinnigan (approved):

PR #720, opened 2017-12-01
[SWARM-1710] Fix test-jar

luan-cestari:
Found a problem when using wildfly swarm with arquilian and I want to use one or more classes of a dependecy's test source code. This scenario I can just import the dependency in maven specifying the type as test-jar. This will make the test source code of that depdency available to me. The problem fixed was that it was using the type to get the specific file extension but in this scenario is not the case

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/internal/ArtifactManager.java
+++ b/core/container/src/main/java/org/wildfly/swarm/internal/ArtifactManager.java
@@ -47,6 +47,9 @@ public class ArtifactManager implements ArtifactLookup {
 
     private static final String COLON = ":";
 
+    private static final String JAR = "jar";
+    private static final String TEST_JAR = "test-jar";
+
     public ArtifactManager() {
     }
 
@@ -122,7 +125,7 @@ public class ArtifactManager implements ArtifactLookup {
 
         String groupId = parts[0];
         String artifactId = parts[1];
-        String packaging = "jar";
+        String packaging = JAR;
         String version = null;
         String classifier = "";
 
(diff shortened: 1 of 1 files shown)


core/container/src/main/java/org/wildfly/swarm/internal/ArtifactManager.java
```diff
@@ -131,12 +134,13 @@ private File findFile(String gav) throws IOException, ModuleLoadException {
         }
 
         if (parts.length == 4) {
-            packaging = parts[2];
+            // the type "test-jar" is packaged as a .jar file more info: https://maven.apache.org/plugins/maven-jar-plugin/examples/create-test-jar.html
+            packaging = TEST_JAR.equals(parts[2]) ? JAR : TEST_JAR;
```
kenfinnigan: Should the test be changed to:
`packaging = TEST_JAR.equals(parts[2]) ? JAR : parts[2]`

At present it looks like any packaging other than `test-jar` will result in the packaging being set to `test-jar`, which doesn't seem right

luan-cestari: Thanks @kenfinnigan  , sorry , I thought that but I wrote something else. 

One thing that I forgot to mention, the unit tests of this class are all commented, is there something wrong wit them?

kenfinnigan:
ok to test

kenfinnigan:
As far as I can tell the test has been commented out for a very long time, and I have no idea why.

If you want to uncomment and see if it works, by all means give it a shot

luan-cestari:
It did't work =/  It was commented in eb913d83cd281826b7cf02047b255ddbb271990f due the internal refactor made which created some of the Maven dependency resolvers   . I can try to create the unit test later, but one thing that I'm not sure is if I still follow the same approach which is more like an integration test or do I mock the dependencies and just see if the logic of that class works fine. What do you think?

kenfinnigan (approved):

PR #721, opened 2017-12-03
SWARM-1641: Fix MP metrics fraction.

mkouba:
Motivation
----------
We need to fix MP Metrics fraction to pass the TCK.

Modifications
-------------
This commit contains a lot of changes but the main change in the
integration code is that we don't add "metrics-api" directly to the
deployment but instead add a module dependency.

Result
------
MP metrics TCK should pass. This commit also fixes SWARM-1639 and
partially SWARM-1640.


client-apis/metrics-api/pom.xml
```diff
@@ -61,6 +61,10 @@
       <groupId>org.wildfly.swarm</groupId>
       <artifactId>cdi</artifactId>
     </dependency>
+    <dependency>
```
kenfinnigan, line 64: Does this mean the client-api is depending on the Fraction?

I think that's backwards from what we've done in the past

mkouba: Yes, metrics-api (which is actually not a client API but merely a CDI extension) depends on fraction API, namely on https://github.com/mkouba/wildfly-swarm/blob/SWARM-1641/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/api/RegistryFactory.java. The reason is that the CDI extension does a JNDI lookup to find a registry factory provided by fraction. We can improve this later...

kenfinnigan: So should the Metrics Client API actually live in https://github.com/wildfly-swarm/wildfly-swarm/tree/master/fractions/cdi-extensions?

I ask because what's present in client-api are not things that we expect a user to directly add a dependency for, but this sounds like it's something that should be? Am I wrong?

mkouba: Yes, I think it should live somewhere else. For me "client API" is something a user would like to communicate with and that's what `metrics-api` is not. See also https://issues.jboss.org/browse/SWARM-1631?focusedCommentId=13496588&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13496588

`metrics-api` should be probably called `metrics-cdi-extension`. But it's not a fraction so I'm not sure whether it should live in `fractions/cdi-extensions`.

kenfinnigan (approved):

PR #728, opened 2017-12-05
SWARM-1711: Update hystrix to 1.5.12 to align with MP Fault Tolerance Impl

heiko-braun:

-----
HOLD: This is for verification purposes

### DIFF
--- a/fractions/netflix/hystrix/pom.xml
+++ b/fractions/netflix/hystrix/pom.xml
@@ -25,6 +25,7 @@
   <properties>
     <swarm.fraction.stability>stable</swarm.fraction.stability>
     <swarm.fraction.tags>Fault Tolerance,Netflix OSS</swarm.fraction.tags>
+    <version.hdrHistogram>2.1.9</version.hdrHistogram>
   </properties>
 
   <build>
@@ -89,7 +90,12 @@
       <groupId>com.netflix.hystrix</groupId>
       <artifactId>hystrix-metrics-event-stream</artifactId>
     </dependency>
-
+    <dependency>
+      <groupId>org.hdrhistogram</groupId>
+      <artifactId>HdrHistogram</artifactId>
+      <version>${version.hdrHistogram}</version>
+      <scope>compile</scope>
+    </dependency>
 
     <dependency>
       <groupId>org.wildfly.core</groupId>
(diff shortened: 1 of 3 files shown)


fractions/netflix/pom.xml
```diff
@@ -26,7 +26,7 @@
     <version.netflix.commons>0.1.1</version.netflix.commons>
     <version.archaius>0.6.6</version.archaius>
     <version.ribbon>2.1.0</version.ribbon>
-    <version.hystrix>1.4.21</version.hystrix>
+    <version.hystrix>1.5.12</version.hystrix>
```
mkouba, line 29: This is not enough IMHO. `hystrix-code` depends on `HdrHistogram`.

heiko-braun: Can you elaborate on this?

kenfinnigan:
In future, I'd prefer that we're testing a PR locally before submitting it as then we at least know it should pass

heiko-braun:
The latest commit adds the `hdrHistogram` deps to the module. The changes have been verified with the CB booster. All that's left is to consolidate the hystrix version property at a shared location. But I am leaving this to @mkouba to do when the tck work is completed.

kenfinnigan (approved):

fractions/netflix/hystrix/pom.xml
```diff
@@ -25,6 +25,7 @@
   <properties>
     <swarm.fraction.stability>stable</swarm.fraction.stability>
     <swarm.fraction.tags>Fault Tolerance,Netflix OSS</swarm.fraction.tags>
+    <version.hdrHistogram>2.1.9</version.hdrHistogram>
```
Ladicek, line 28: Should be in the root `pom.xml`, together with a `<dependencyManagement>` entry, like all other dependecies?

heiko-braun: yes, but as I said, I would leave this to @mkouba to fix together with the MP TCK suite he is working on

heiko-braun: I was a bit eager, this is hdr dep, yes maybe you are right

heiko-braun: not sure, @kenfinnigan what's you opinion: where should deps like this be managed?

kenfinnigan: Currently all netflix fraction dependencies are depMan'd in https://github.com/wildfly-swarm/wildfly-swarm/blob/master/fractions/netflix/pom.xml

Maybe there?

mkouba: +1 for netflix parent

PR #744, opened 2017-12-11
SWARM-1638: Simplify deployment module dependencies.

mkouba:
Motivation
----------
Simplify MP Auth deployment module dependencies. Get rid of hardcoded
versions in module descriptors.

Modifications
-------------
A single deployment module dependency is defined - fraction deployment
module which exports MP Auth API and jose4j modules.
MPJWTAuthExtensionArchivePreparer does not have to add
jwt-auth-wfswarm.jar - a workaround for
github.com/shrinkwrap/shrinkwrap/issues/112.

Result
------
Everything should work as before.


fractions/microprofile/microprofile-jwt/pom.xml
```diff
@@ -24,6 +24,7 @@
   <properties>
     <swarm.fraction.stability>experimental</swarm.fraction.stability>
     <swarm.fraction.tags>MicroProfile,JWT,Security,Web</swarm.fraction.tags>
+    <version.jose4j>0.6.0</version.jose4j>
```
Ladicek: Should be in `/build-parent/pom.xml` like everything else?

mkouba: I'm not sure. I suppose that `build-parent` is mostly useful for dependencies which might be shared and this is probably not the case. Also there are few places where we don't use build-parent, e.g. netflix: https://github.com/wildfly-swarm/wildfly-swarm/blob/master/fractions/netflix/pom.xml#L26 or even hardcoded versions in fluentd: https://github.com/wildfly-swarm/wildfly-swarm/blob/master/fractions/fluentd/src/main/resources/modules/org/fluentd/main/module.xml#L7.

@kenfinnigan @heiko-braun What do you think?

kenfinnigan: I don't think it's necessary for all versions of things to be in build-parent, but where it makes the most sense for updating versions of things together.

Beyond that I don't have a particular preference. As you point out @mkouba, there are plenty of places where putting things in build-parent are not followed

Ladicek: Ayea. I figured I'd rather ask, but I'm fine with this.

mkouba:
Linux build failure:
```
java.lang.AssertionError: 
JSON path 'metricTest.test1.histogram'.mean doesn't match.
Expected: <499.5F>
  Actual: 500.96945
	at org.eclipse.microprofile.metrics.test.MpMetricTest.testApplicationMetricsJSON(MpMetricTest.java:438)
```

kenfinnigan (approved):

PR #750, opened 2017-12-19
Mp 1.1

pilhuhn:

-----

This is SWARM-1746 implementing the Eclipse MicroProfile Metrics Spec version 1.1

### DIFF
--- a/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/MicroprofileMetricsFraction.java
+++ b/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/MicroprofileMetricsFraction.java
@@ -17,13 +17,13 @@
 package org.wildfly.swarm.microprofile.metrics;
 
 import org.wildfly.swarm.spi.api.Fraction;
-import org.wildfly.swarm.spi.api.annotations.Configurable;
 import org.wildfly.swarm.spi.api.annotations.DeploymentModule;
+import org.wildfly.swarm.spi.api.annotations.WildFlyExtension;
 
 /**
  * @author Heiko W. Rupp
  */
 @DeploymentModule(name = "org.wildfly.swarm.microprofile.metrics", slot = "deployment", export = true, metaInf = DeploymentModule.MetaInfDisposition.IMPORT)
-@Configurable("swarm.microprofile.metrics")
+@WildFlyExtension(module = "org.wildfly.extension.microprofile.config")
 public class MicroprofileMetricsFraction implements Fraction<MicroprofileMetricsFraction> {
 }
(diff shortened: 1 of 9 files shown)


fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/MicroprofileMetricsFraction.java
```diff
@@ -17,13 +17,13 @@
...
 import org.wildfly.swarm.spi.api.Fraction;
-import org.wildfly.swarm.spi.api.annotations.Configurable;
 import org.wildfly.swarm.spi.api.annotations.DeploymentModule;
+import org.wildfly.swarm.spi.api.annotations.WildFlyExtension;
 
 /**
  * @author Heiko W. Rupp
  */
 @DeploymentModule(name = "org.wildfly.swarm.microprofile.metrics", slot = "deployment", export = true, metaInf = DeploymentModule.MetaInfDisposition.IMPORT)
-@Configurable("swarm.microprofile.metrics")
```
kenfinnigan, line 27: So this means it can only be configured through MP Config methods, correct?

fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/MicroprofileMetricsFraction.java
```diff
@@ -17,13 +17,13 @@
...
+@WildFlyExtension(module = "org.wildfly.extension.microprofile.config")
```
kenfinnigan, line 27: Why does the MP Metrics Fraction need to refer to MP Config WF Extension?

The presence of MP Config should be what's required to trigger the WF Extension to be loaded.

fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/MicroprofileMetricsFraction.java
pilhuhn: Yes.

fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/MicroprofileMetricsFraction.java
pilhuhn: I have no idea - I think when starting this I needed it to pull in config. May have changed since. Or I just misunderstood things. 
I guess at the time there was no overall-MP-fraction to pull in, but Metrics as standalone needed to tell that it needs MP-config.

michalszynkiewicz:
@pilhuhn could you also disable the flickering tests as you did in https://github.com/eclipse/microprofile-metrics/pull/191/files ?

pilhuhn:
@michalszynkiewicz That should happen via the PR you cited, as the tests are pulled in from the microprofile test jar.

kenfinnigan:
Merged manually with some minor changes

PR #751, opened 2017-12-28
SWARM-1757 - chkconfig does not configure runlevels properly

nelsongraca:
Just two small changes, one for chkconfig (two commented lines that are required), another is a small extra check to avoid an error message when start-stop-daemon is not present.


kenfinnigan:
add to whitelist

kenfinnigan (approved):

PR #754, opened 2018-01-03
ENTSWM-256: switched to fraction plugin 79

michalszynkiewicz:

-----

### DIFF
--- a/boms/pom.xml
+++ b/boms/pom.xml
@@ -53,6 +53,7 @@
         <plugin>
           <groupId>org.wildfly.swarm</groupId>
           <artifactId>wildfly-swarm-fraction-plugin</artifactId>
+          <version>${version.wildfly.swarm.fraction.plugin}</version>
           <executions>
             <execution>
               <id>generate-bom</id>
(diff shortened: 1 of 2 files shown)


boms/pom.xml
```diff
@@ -53,6 +53,7 @@
         <plugin>
           <groupId>org.wildfly.swarm</groupId>
           <artifactId>wildfly-swarm-fraction-plugin</artifactId>
+          <version>${version.wildfly.swarm.fraction.plugin}</version>
```
kenfinnigan, line 56: Is this needed?

Would expect the version to be picked up by the definition in pluginManagement of build-parent

michalszynkiewicz: When I used 79-SNAPSHOT locally and switched the version in the build-parent, bom was still using the latest released version. When I added this line, it started using the desired version.
I didn't observe this behavior with a released version, maybe the newest released version is used when no version is provided?

I haven't found any info about inheriting between pluginmanagement's. IntelliJ is not aware it should be inherited.

kenfinnigan (approved):

PR #756, opened 2018-01-08
SWARM-1457: Swarmtool unable to resolve undertow artifacts

AdlerFleurant:
Motivation
----------
In certain circumstances, Wildfly Swarm Dependency Management may define an artifact version while while a dependency may require some specific different version of that same artifact. If using Gradle, the current code would only add the defined dependency management version inside the m2repo of the uberjar. Furthermore Gradle only considers compile dependencies and would not include dependencies in the runtime configurations.

Modifications
-------------
org.wildfly.swarm.plugin.gradle.GradleArtifactResolvingHelper:doResolve no longer tries to dependency managed everything by forcing modules.
org.wildfly.swarm.plugin.gradle.PackageTask:packageForSwarm resolves all artifacts in the default configurations-all runtime and compile dependencies.

Result
------
The m2repo of an uberjar will now contain all runtime and compile artifacts required by all modules. This may cause the uberjar to be bigger in size.

- [X ] Have you followed the guidelines in our [Contributing](http://wildfly-swarm.io/community/contributing/) document?

-----


kenfinnigan:
add to whitelist

kenfinnigan (approved):

PR #757, opened 2018-01-08
SWARM-1677 - Too many temp files.

bobmcwhirter:
Motivation
----------
We explode a lot of temp files from within uberjars. We shouldn't do that
because it wastes space and time.

Modifications
-------------
* Add an uberjar:// URL handler, get it registered with the jboss-modules
  modular URL stream factory.
* Adjust our usage of MavenResolvers to a possibly-stream-based ArtifactResolver
* Still wrap our new ArtifactResolvers into a JBoss-Modules happy file-based
  MavenResolver.
* Monkey-patch the jboss-modules MavenArtifactUtils to not insist on files
  being on disk.
* Iff possible, stream any jars-within-uberjars into JBoss-Module ClassSpec
  byte[] arrays directly in-memory, instead of copying them out of the
  uberjar onto the filesystem.
* But don't do it for org.wildfly.swarm:* artifacts, which are bean archives,
  because uberjar:// annoys the crap out of Weld (Hi Martin).

Result
------
Less disk-space used!

-----

### DIFF
--- a/core/bootstrap/src/main/java/org/jboss/modules/maven/MavenArtifactUtil.java
+++ b/core/bootstrap/src/main/java/org/jboss/modules/maven/MavenArtifactUtil.java
@@ -40,6 +40,9 @@ import org.jboss.modules.Module;
 import org.jboss.modules.ResourceLoader;
 import org.jboss.modules.ResourceLoaders;
 import org.wildfly.swarm.bootstrap.logging.BootstrapLogger;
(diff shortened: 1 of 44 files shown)


kenfinnigan (approved):

Ladicek:
We'll need to measure it for sure, but can you guess what the memory usage impact might be?

bobmcwhirter:
I think negligible.  JBoss Modules already loads up byte[] for the classes.
But yes we should measure.

PR #760, opened 2018-01-09
ENTSWM-255: Turn transformer tests into integration tests 

heiko-braun:
Bypass classloading issues the interfere with test execution. The assertions remains the same, but the test have been turned into integration tests (running against the uberJar, predictable classloading isolation) opposed to the former arquillian tests (fishy classloading, unknow error sources)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,6 +85,13 @@
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-failsafe-plugin</artifactId>
+        <configuration>
+          <excludedGroups>${testCategory.excluded}</excludedGroups>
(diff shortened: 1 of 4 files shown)


heiko-braun:
@kenfinnigan @mkouba I wasn't able to tell why the error occurs exactly, but it seems related to the changes in cc712476338c0fd0fd7d0bec9b43598e84fa38e4. Revisiting the assertion however, I think we can move this to the integration tests and work around the classloading issues in arquillian tests. Looking at the other IT's this is what we did there as well and as far as I recall, there was always an issues with JAXP in arquillian which led to this decision.

Wether or not you think this change is appropriate or not needs to be discussed.

kenfinnigan (approved):

PR #761, opened 2018-01-11
SWARM-1762: Update to MP Config 1.2 impl

heiko-braun:
- Disable TCK tests that have been challenged
- Update to MP config 1.2 and WF Config 1.2.1

### DIFF
--- a/testsuite/microprofile-tcks/config/pom.xml
+++ b/testsuite/microprofile-tcks/config/pom.xml
@@ -94,6 +94,13 @@
                <dependenciesToScan>
                   <dependency>org.eclipse.microprofile.config:microprofile-config-tck</dependency>
                </dependenciesToScan>
+
+               <properties>
+                  <property>
+                     <name>listener</name>
+                     <value>org.wildfly.swarm.microprofile.config.tck.TestNGAnnotationTransformer</value>
+                  </property>
+               </properties>
             </configuration>
          </plugin>
       </plugins>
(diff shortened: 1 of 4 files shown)


kenfinnigan (approved):

testsuite/microprofile-tcks/config/pom.xml
```diff
@@ -94,6 +94,13 @@
                <dependenciesToScan>
                   <dependency>org.eclipse.microprofile.config:microprofile-config-tck</dependency>
                </dependenciesToScan>
+
+               <properties>
+                  <property>
+                     <name>listener</name>
+                     <value>org.wildfly.swarm.microprofile.config.tck.TestNGAnnotationTransformer</value>
```
mkouba, line 101: @heiko-braun Why don't you add the exclusions to https://github.com/wildfly-swarm/wildfly-swarm/blob/master/testsuite/microprofile-tcks/config/src/test/tck-suite.xml ?

heiko-braun: None of the ways I've tried to add exclusion to `test-suite.xml` did work. It seems testng internally uses annotations to categorize tests and these can be used for exclusion, but not class names or other direct references. This was the best way I could figure out.

mkouba: Hm, `test-suite.xml` should work (we're using it e.g. in CDI TCK; https://github.com/cdi-spec/cdi-tck/blob/master/impl/src/main/resources/tck-tests.xml#L28) and imho should be the preferred way of exclusion if using testng. I'll try it.

PR #787, opened 2018-02-14
SWARM-1834: Hystrix is a dep of MP FT and needs to be present in supported profile

heiko-braun:
This is followup commit to SWARM-1834. The hystrix dep was missing.


Ladicek:
I wonder what are the supportability implications. Probably should discuss internally.

heiko-braun:
Thanks @Ladicek. I started a discussion on the ML

PR #788, opened 2018-02-15
Further netflix deps are required for MP FT

heiko-braun:
All the hystrix deps need to present in supported profile as well.


kenfinnigan:
Was this approved by anyone before merging? If not, in the future please wait for a review before merging

Also, did CI pass? It doesn’t matter how simple the change is, we need to ensure CI passes before merging

PR #793, opened 2018-02-20
SWARM-1841: Create a fraction that covers all EE related fractions

heiko-braun:


fractions/javaee/full/README.adoc
```diff
@@ -0,0 +1,16 @@
+= Web
```
Ladicek: Should probably be updated :-)

kenfinnigan:
Couple of things:

- Does `datasource` need to be added directly? Would expect it to come in via `jpa` of `web`
- Should change `Full Profile` to `Full Platform` as John is forever telling me there's no such thing as the `Full Profile`

fractions/javaee/full/README.adoc
```diff
@@ -0,0 +1,25 @@
+= Full
+
+Provides a collection of fractions equivalent to the _Full Platform:
```
Ladicek, line 3: Just noticed, missing underscore at the end of the line, before the colon :-)

PR #799, opened 2018-02-21
SWARM-1530 - Temp files exists after stop an uber jar in Windows

juangon:
Motivation
----------
Windows is more sensitive for used temp files when tried to delete them. This causes temp files not being removed when Swarm applications are shut down.

Modifications
-------------
* All JarFile objects should be closed prior to remove their files. Added a JarFileManager registry to close them properly.
* Removed temp files on shutdown for those being temporary created as needed Maven artifacts.
* Removed all Shutdownhooks except one, in order to unify and control where those files are being removed for every environment.
* Removed file.deleteOnExit. All files should be deleted under Swarm control in a determined Swarm lifecycle (when server stops).

Result
------
Majority of temp files being removed on shutdown and Windows environments.

- [] Have you built the project locally prior to submission with `mvn clean install`?

-----

### DIFF
--- a/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
+++ b/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
@@ -65,6 +65,7 @@ import org.wildfly.swarm.tools.DeclaredDependencies;
 import org.wildfly.swarm.tools.exec.SwarmExecutor;
 import org.wildfly.swarm.tools.exec.SwarmProcess;
 
+
 public class UberjarSimpleContainer implements SimpleContainer {
 
     private final ContainerContext containerContext;
@@ -274,7 +275,7 @@ public class UberjarSimpleContainer implements SimpleContainer {
                 System.err.println("-> " + each.getKey());
             }*/
 
-        File executable = File.createTempFile("arquillian", "-swarm.jar");
+        File executable = File.createTempFile(TempFileManager.WFSWARM_TMP_PREFIX + "arquillian", "-swarm.jar");
         wrapped.as(ZipExporter.class).exportTo(executable, true);
         executable.deleteOnExit();
 
@@ -286,15 +287,14 @@ public class UberjarSimpleContainer implements SimpleContainer {
 
         executor.withProperty("java.net.preferIPv4Stack", "true");
 
-        File processFile = File.createTempFile("mainprocessfile", null);
-        processFile.deleteOnExit();
+        File processFile = File.createTempFile(TempFileManager.WFSWARM_TMP_PREFIX + "mainprocessfile", null);
 
         executor.withProcessFile(processFile);
 
         executor.withJVMArguments(getJavaVmArgumentsList());
         executor.withExecutableJar(executable.toPath());
 
-        File workingDirectory = TempFileManager.INSTANCE.newTempDirectory("arquillian", null);
+        workingDirectory = TempFileManager.INSTANCE.newTempDirectory("arquillian", null);
         executor.withWorkingDirectory(workingDirectory.toPath());
 
         this.process = executor.execute();
@@ -356,6 +356,7 @@ public class UberjarSimpleContainer implements SimpleContainer {
     @Override
     public void stop() throws Exception {
         this.process.stop();
+        TempFileManager.deleteRecursively(workingDirectory);
     }
 
     private String ga(final MavenCoordinate coord) {
@@ -387,6 +388,22 @@ public class UberjarSimpleContainer implements SimpleContainer {
 
     }
 
+    private void deleteRecursively(File file) {
+        if (!file.exists()) {
+            return;
+        }
+        if (file.isDirectory()) {
+            File[] files = file.listFiles();
+            if (files != null) {
+                for (File child : files) {
+                    deleteRecursively(child);
+                }
+            }
+        }
+
+        file.delete();
+    }
+
     private final Class<?> testClass;
 
     private SwarmProcess process;
@@ -394,6 +411,8 @@ public class UberjarSimpleContainer implements SimpleContainer {
     private Set<String> requestedMavenArtifacts = new HashSet<>();
 
     private String javaVmArguments;
+
+    private File workingDirectory;
 
 }
 
--- a/core/bootstrap/src/main/java/org/jboss/modules/maven/MavenArtifactUtil.java
+++ b/core/bootstrap/src/main/java/org/jboss/modules/maven/MavenArtifactUtil.java
@@ -30,6 +30,8 @@ import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.util.List;
 import java.util.jar.JarFile;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.xml.xpath.XPath;
 import javax.xml.xpath.XPathExpression;
@@ -40,6 +42,7 @@ import org.jboss.modules.Module;
 import org.jboss.modules.ResourceLoader;
 import org.jboss.modules.ResourceLoaders;
 import org.wildfly.swarm.bootstrap.logging.BootstrapLogger;
+import org.wildfly.swarm.bootstrap.util.JarFileManager;
 import org.xml.sax.InputSource;
 
 /**
@@ -58,6 +61,8 @@ public final class MavenArtifactUtil {
     private static final XPath xpath = XPathFactory.newInstance().newXPath();
 
     private static XPathExpression snapshotVersionXpath;
+
+    private static final Pattern tempFilePattern = Pattern.compile("\\S+[0-9]{5,}.\\S{5,}");
 
     static {
         try {
@@ -265,7 +270,13 @@ public final class MavenArtifactUtil {
     public static ResourceLoader createMavenArtifactLoader(final MavenResolver mavenResolver, final String name) throws IOException {
         File fp = mavenResolver.resolveJarArtifact(ArtifactCoordinates.fromString(name));
         if (fp == null) return null;
-        JarFile jarFile = new JarFile(fp, true);
+        Matcher matcher = tempFilePattern.matcher(fp.getName());
+        JarFile jarFile = null;
+        if (matcher.matches()) {
+            jarFile = JarFileManager.INSTANCE.addJarFile(fp);
+        } else {
+            jarFile = new JarFile(fp, true);
+        }
         return ResourceLoaders.createJarResourceLoader(name, jarFile);
     }
 
--- a/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/BootstrapModuleFinder.java
+++ b/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/BootstrapModuleFinder.java
@@ -17,10 +17,7 @@ package org.wildfly.swarm.bootstrap.modules;
 
 import java.io.File;
 import java.io.IOException;
-import java.util.Enumeration;
 import java.util.HashSet;
-import java.util.Set;
-import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 
 import org.jboss.modules.DependencySpec;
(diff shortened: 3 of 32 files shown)


juangon:
@bobmcwhirter @Ladicek some things about this:

1-Don't know why uberjars doesn't stop gracefully, even when telling it to do so. This causes the file leaks on those places using uberjar at this point. Had to disable the java.io.tmpdir checks for these tests (5 of them).
2-Added a delete method so when maven plugin "stop" kicks in, it removes all the stuff created by Swarm in temp dir. So I think THIS time we let everything cleaned.

kenfinnigan:
retest this please

arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
```diff
@@ -387,6 +388,22 @@ private String gav(final String group, final String artifact, final String versi
...
+        if (file.isDirectory()) {
+            File[] files = file.listFiles();
+            if (files != null) {
+                for (File child : files) {
+                    deleteRecursively(child);
+                }
+            }
+        }
+
+        file.delete();
```
heiko-braun, line 404: According to what we found out together, I think Files.delete(Path) is more reliable or at least it has better error reporting. We may want to switch to that API here.

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/UberJarMavenResolver.java
```diff
@@ -63,14 +67,25 @@ public File resolveArtifact(ArtifactCoordinates coordinates, String packaging) t
...
             }
         }
 
         return resolved;
     }
 
+    @Override
+    public void close() throws IOException {
+        resolutionCache.forEach((a, f) -> {
+             f.delete();
```
heiko-braun, line 85: As mentioned above, Files.delete(Path) is recommended I think

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/TempFileManager.java
```diff
@@ -69,26 +66,26 @@ private void register(File file) {
...
+            File[] children = f.listFiles();
+            for (int i = 0; i < children.length; ++i) {
+                if (!deleteRecursively(children[i])) {
+                    return false;
                 }
             }
         }
 
-        file.delete();
+        return f.delete();
```
heiko-braun, line 88: Same here (See above)

meta/fraction-metadata/src/main/java/org/wildfly/swarm/fractions/scanner/JarScanner.java
```diff
@@ -36,14 +36,18 @@ public String extension() {
...
+            try (InputStream input = pathSource.getInputStream();
+                    FileOutputStream out = new FileOutputStream(jarFile)) {
+                copy(input, out);
+            }
 
-        handleFileAsZip.accept(jarFile);
-    }
+            handleFileAsZip.accept(jarFile);
+        } finally {
+            jarFile.delete();
```
heiko-braun, line 50: Same here

meta/fraction-metadata/src/test/java/org/wildfly/swarm/fractions/FractionUsageAnalyzerTest.java
```diff
@@ -39,14 +46,15 @@ public void testFractionMatching() throws Exception {
...
-        out.deleteOnExit();
 
         analyzer.source(out);
         assertThat(analyzer.detectNeededFractions()
                            .stream()
                            .filter(fd -> fd.getArtifactId().equals("jaxrs"))
                            .count())
                 .isEqualTo(1);
+
+        out.delete();
```
heiko-braun, line 57: Files.delete(Path) instead?

meta/fraction-metadata/src/test/java/org/wildfly/swarm/fractions/FractionUsageAnalyzerTest.java
```diff
@@ -74,13 +82,14 @@ public void testDetectEmptyWarAsUndertow() throws Exception {
...
-        out.deleteOnExit();
 
         analyzer.source(out);
         assertThat(analyzer.detectNeededFractions()
                            .stream()
                            .filter(fd -> fd.getArtifactId().equals("undertow"))
                            .count())
                 .isEqualTo(1);
+
+        out.delete();
```
heiko-braun, line 93: Files.delete(Path) instead?

plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StopMojo.java
```diff
@@ -59,8 +62,35 @@ public void executeSpecific() throws MojoExecutionException, MojoFailureExceptio
...
+        if (f.isDirectory()) {
+            File[] children = f.listFiles();
+            for (int i = 0; i < children.length; ++i) {
+                if (!deleteRecursively(children[i])) {
+                    return false;
+                }
+            }
+        }
+
+        return f.delete();
```
heiko-braun, line 92: Files.delete(Path) instead?

heiko-braun:
@juangon I think it looks good. I just left a remark in some places to switch to Files.delete(Path) opposed to using File.delete(). I believe that is the recommended way.

After digging some further, I think the benefit of the Files.delete() API is that it throws an exception when the deletion fails [1]. I think in the context of this issue, that might be reasonable.

[1] https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#delete-java.nio.file.Path-

juangon:
Ok @heiko-braun thanks for your review will do your suggestions. Tomorrow moving day so will do on Friday or even tomorrow if I find some time.

juangon:
Hmm failed in latest test, Microprofile TCK rest client. Will check on Friday.

juangon:
@Ladicek I would love hearing your suggestions also. Will do all changes tomorrow, together with the fix of the failing test. Thanks!

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/BootstrapModuleFinder.java
```diff
@@ -59,17 +56,23 @@ public void buildModule(ModuleSpec.Builder builder, ModuleLoader delegateLoader)
 
             ApplicationEnvironment env = ApplicationEnvironment.get();
 
+            PathFilter filter = new PathFilter() {
+                @Override
+               public boolean accept(String path) {
+                    return path.endsWith("module.xml");
```
Ladicek, line 62: Should probably be `/module.xml`?

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/JarFileManager.java
```diff
@@ -0,0 +1,57 @@
...
+        return jarFile;
+    }
+
+    public void close() throws IOException {
+        jarFileToClose.
+                forEach((f, j) -> {
+                        try {
+                            j.close();
+                        } catch (IOException e) {
+                        }
```
Ladicek, line 51: Shouldn't really swallow exceptions. All close operations should be attempted, the first exception thrown should be remembered, and rethrown at the end.

core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java
```diff
@@ -279,10 +284,40 @@ public void stop() throws Exception {
...
+            Class<?> shutdownHookHolder = null;
+            Class<?>[] declaredClasses = serviceContainer.getClass().getDeclaredClasses();
+            for (Class<?> clazz : declaredClasses) {
+                if (clazz.getName().contains("ShutdownHookHolder")) {
+                    shutdownHookHolder = clazz;
+                }
+            }
+
+            if (shutdownHookHolder != null) {
+                    Field containersSetField = shutdownHookHolder.getDeclaredField("containers");
```
Ladicek, line 303: Indentation :-)

core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java
```diff
@@ -279,10 +284,40 @@ public void stop() throws Exception {
...
+                    containersSetField.setAccessible(true);
+                    Set<?> set = (Set<?>)containersSetField.get(null);
+                    set.clear();
+            }
+        } catch (Throwable e) {
+            e.printStackTrace();
```
Ladicek, line 309: Please don't.

fractions/javaee/undertow/src/test/java/org/wildfly/swarm/undertow/runtime/CertInfoProducerTest.java
```diff
@@ -16,18 +16,36 @@
...
 public class CertInfoProducerTest {
 
+    @After
+    public void tearDown() {
+       try {
+            JarFileManager.INSTANCE.close();
+            MavenResolvers.close();
+            TempFileManager.INSTANCE.close();
+        } catch (IOException e) {
+            e.printStackTrace();
```
Ladicek, line 45: Please don't.

plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StopMojo.java
```diff
@@ -59,8 +62,35 @@ public void executeSpecific() throws MojoExecutionException, MojoFailureExceptio
         if (tmp != null && tmp.exists()) {
             tmp.delete();
         }
+
+        Path tmpDir = new File(System.getProperty("java.io.tmpdir")).toPath();
```
Ladicek, line 66: This looks a bit dangerous. Also there's `TempFileManager.TMPDIR_PROPERTY`, not sure how relevant it is here.

juangon: Those are different properties, and I think would only cover certain cases. I am opened if you have other better alternatives.

core/bootstrap/src/main/java/org/jboss/modules/maven/MavenArtifactUtil.java
```diff
@@ -265,7 +270,13 @@ public static ResourceLoader createMavenArtifactLoader(final String name) throws
     public static ResourceLoader createMavenArtifactLoader(final MavenResolver mavenResolver, final String name) throws IOException {
         File fp = mavenResolver.resolveJarArtifact(ArtifactCoordinates.fromString(name));
         if (fp == null) return null;
-        JarFile jarFile = new JarFile(fp, true);
+        Matcher matcher = tempFilePattern.matcher(fp.getName());
+        JarFile jarFile = null;
+        if (matcher.matches()) {
```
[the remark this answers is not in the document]
juangon: I didn't find any better solution in this case. Do you have any idea instead of checking a file pattern?

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/BootstrapModuleFinder.java
juangon: Ok, just in case someone wants to name something like blablahfoofoomodule.xml

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/JarFileManager.java
juangon: I should have done this. Being lost in all other problems. Thanks.

core/bootstrap/src/main/java/org/jboss/modules/maven/MavenArtifactUtil.java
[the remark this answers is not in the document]
Ladicek: Ah, I see, this is a forked class from JBoss Modules. Probably no better way, then.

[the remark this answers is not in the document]
juangon: Now you focus on this, I can improve the pattern by adding the prefix I added for all files created under Swarm.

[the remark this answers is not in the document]
Ladicek: Yea sounds like a good idea.

juangon:
Changes and suggestions here: https://github.com/wildfly-swarm/wildfly-swarm/pull/809

PR #802, opened 2018-02-22
SWARM-1846: Make MP-restclient usable from behind a firewall

wgresshoff:
Motivation
----------
Make MP REST client work behind a firewall.

Modifications
-------------
In BuilderImpl build(Class<T> aClass) first look if a proxy host is
defined before a client is build.

Result
------
RestEasy now uses a proxy if one is defined in the system properties.

- [x ] Have you created a [JIRA](https://issues.jboss.org/browse/SWARM) and used it in the commit message?
- [x ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/wildfly-swarm/wildfly-swarm/pulls) for the same issue?
- [x ] Have you built the project locally prior to submission with `mvn clean install`?

-----

### DIFF
--- a/client-apis/microprofile-restclient/src/main/java/org/wildfly/swarm/microprofile/restclient/BuilderImpl.java
+++ b/client-apis/microprofile-restclient/src/main/java/org/wildfly/swarm/microprofile/restclient/BuilderImpl.java
@@ -109,11 +109,12 @@ class BuilderImpl implements RestClientBuilder {
 
         final T actualClient;
 
-        if (System.getProperty("http.proxyHost") != null
-                && !noProxyHosts.contains(this.baseURI.getHost())) {
+        final String proxyHost = System.getProperty("http.proxyHost");
+
+        if (proxyHost != null && !noProxyHosts.contains(this.baseURI.getHost())) {
             // Use proxy, if defined
             actualClient = this.builderDelegate.defaultProxy(
-                    System.getProperty("http.proxyHost"),
+                    proxyHost,
                     Integer.parseInt(System.getProperty("http.proxyPort", "80")))
                     .build()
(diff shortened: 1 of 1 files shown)


heiko-braun:
Add to whitelist

heiko-braun (dismissed):

heiko-braun:
Thanks Werner

client-apis/microprofile-restclient/src/main/java/org/wildfly/swarm/microprofile/restclient/BuilderImpl.java
```diff
@@ -102,13 +104,33 @@ public RestClientBuilder baseUrl(URL url) {
...
-                .build();
+        List<String> noProxyHosts = Arrays.asList(
+                System.getProperty("http.nonProxyHosts", "localhost|127.*|[::1]").split("|"));
+
+        final T actualClient;
+
+        if (System.getProperty("http.proxyHost") != null
+                && !noProxyHosts.contains(this.baseURI.getHost())) {
+            // Use proxy, if defined
+            actualClient = this.builderDelegate.defaultProxy(
```
kenfinnigan: My only comment would be whether the System.getProperty() call result should be stored as a local variable as opposed to making two separate calls?

Don't recall if there's much overhead to System.getProperty() calls, just a thought

wgresshoff: Don't know either but changed the code accordingly. Just pushed it to the pull request.

kenfinnigan:
Thanks!

kenfinnigan (approved):

PR #803, opened 2018-02-22
SWARM-1794: Fix MP FT Timeout support.

mkouba:
Motivation
----------
Timeout does not work as expected - the execution thread is never
interrupted unless the operation is also marked as asynchronous.

Modifications
-------------
Always use THREAD isolation strategy for timeout operations. Also modify
the way the retry logic is implemented.

Result
------
Execution thread should be interrupted when timeout occurs.

### DIFF
--- a/testsuite/microprofile-tcks/fault-tolerance/pom.xml
+++ b/testsuite/microprofile-tcks/fault-tolerance/pom.xml
@@ -99,6 +99,11 @@
                      <name>java.util.logging.config.file</name>
                      <value>${project.build.testOutputDirectory}/logging.properties</value>
                   </property>
+                  <!-- Uncomment to debug arquillian test -->
+                  <!-- property>
+                     <name>swarm.debug.port</name>
+                     <value>8787</value>
+                  </property-->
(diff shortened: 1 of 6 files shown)


kenfinnigan:
retest this please

testsuite/microprofile-tcks/fault-tolerance/pom.xml
```diff
@@ -99,6 +99,11 @@
                      <name>java.util.logging.config.file</name>
                      <value>${project.build.testOutputDirectory}/logging.properties</value>
                   </property>
+                  <!-- Uncomment to debug arquillian test ->
```
kenfinnigan: Looks like this is what's causing the pom parsing to fail. Needs `-->` instead at the end

mkouba:
The failure on windows machine is `BulkheadSynchRetryTest` again. @antoinesd Have you discovered something?

antoinesd (approved):

antoinesd:
not yet

kenfinnigan (approved):

PR #806, opened 2018-02-22
Switch to 1.3.x of config-api in community

heiko-braun:

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -26,7 +26,7 @@
 
     <version.org.snakeyaml>1.17</version.org.snakeyaml>
 
-    <version.wildfly.config-api>1.2.0</version.wildfly.config-api>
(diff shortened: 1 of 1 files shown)


Ladicek (approved):

scottmarlow:
Do the other config-api's also need to be updated/released, more specifically the mongo.config-api, cassandra.config-api, neo4j.config-api, orientdb.config-api?

Ladicek:
I don't think the NoSQL config APIs need to be bumped, as they are not used in product (right?). What needs to be bumped is the Keycloak config API.

scottmarlow:
Correct, the NoSQL config APIs are not in product.  Thanks for clarifying!

heiko-braun:
@scottmarlow : @Ladicek is right. Only the ones shared with the prod build need to be bumped to avoid conflicts. 

And @Ladicek : The KC update is here https://github.com/wildfly-swarm/wildfly-swarm/pull/807

PR #807, opened 2018-02-22
Switch to 1.3.x of keycloak-config-api in community

heiko-braun:

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -28,7 +28,7 @@
 
     <version.wildfly.config-api>1.2.0</version.wildfly.config-api>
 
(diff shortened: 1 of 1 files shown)


heiko-braun:
retest this please

Ladicek:
Maybe Keycloak config API 1.3 should use the WildFly config API 1.3? Not sure how big of a deal that is.

heiko-braun:
@Ladicek I guess you talk about this: https://github.com/wildfly-swarm/keycloak-config-api/blob/master/pom.xml#L19

And yes, you are right I think. I don't know the impact, but it's better not take a risk. I'll spin another release

Ladicek (approved):

PR #810, opened 2018-02-23
SWARM-1693 - Microprofile + OpenAPI 1.0.1 spec implementation (fraction)

EricWittmann:

Note: `mvn clean install -DskipTests` works, but `mvn test` fails in the CDI test.  However, all of the OpenAPI tests pass, and the CDI test has never worked for me.

-----


kenfinnigan:
add to whitelist

mkouba:
retest this please

kenfinnigan:
test this please

mkouba:
@EricWittmann What do you mean by CDI test?

EricWittmann:
When I build (even from master) on my local workstation it fails on `Test Suite: CDI`.  So while working on this I've been doing `mvn clean install -DskipTests` and then running only the test suites relevant to my work.

Also a full build *without* running the tests takes around 40 minutes on my machine.  No idea how long it would take for me to do the full build without `-DskipTests`.

EricWittmann:
The next question is probably "why is Test Suite: CDI failing?".  That's fair.  If interested, here is the output:

https://gist.github.com/EricWittmann/e061b325f124f55faa3ea1e51a01958e

kenfinnigan:
retest this please

juangon:
retest this please

mkouba:
@juangon retest doesn't help here - we need to fix parent versions in few poms. I'm looking into this right now.

juangon:
Oh ok @mkouba

mkouba:
Failed tck tests (investigating):
> Test Result ("6 neúspěšných ")
    org.eclipse.microprofile.openapi.tck.OASConfigScanPackageTest.testScanPackage
    org.eclipse.microprofile.openapi.tck.OASConfigScanPackageTest.testScanPackage
    org.eclipse.microprofile.openapi.tck.OASConfigScanClassTest.testScanClass
    org.eclipse.microprofile.openapi.tck.OASConfigScanClassTest.testScanClass
    org.eclipse.microprofile.openapi.tck.OASConfigScanClassesTest.testScanClasses
    org.eclipse.microprofile.openapi.tck.OASConfigScanClassesTest.testScanClasses

EricWittmann:
Sorry for the parent versions - I fixed those locally but forgot to push them prior to submitting the PR. :(

Looking at the errors, it seems to be a ClassNotFound error...which is weird.  I'll poke around to see what I can find.

mkouba:
@EricWittmann It seems there is a regression caused by this commit https://github.com/wildfly-swarm/wildfly-swarm/pull/810/commits/766dbb95416fdead7b4dd6bd6e1dadca442e4155. `OpenApiDataObjectScanner` now attempts to load the classed when comparing two types of which at least one is not in Jandex (see also `org.wildfly.swarm.microprofile.openapi.runtime.util.TypeUtil.isA(IndexView, Type, Type)`). The problem is that during `OpenApiDeploymentProcessor.process()` the deployment does not exist yet and thus we get `ClassNotFoundException` if we try to load a class from the deployment (even if we try to use TCCL).

Furthermore, `OASConfigScanClassesTest` is using `mp.openapi.scan.classes` which effectively removes model classes (such as `org.eclipse.microprofile.openapi.apps.airlines.model.Flight`) from the Jadex index. I'm not sure whether this is OK or not.

A simple fix would be to include all deployment classes and all JDK classes used in `OpenApiDataObjectScanner` to the index.

EricWittmann:
Thanks for the analysis, @mkouba - I'm worried why this didn't fail for me locally when I ran the TCK test prior to submitting the PR.  Give me a little while to explore that, and also discuss the class loading problem with @msavy

EricWittmann:
So I've reproduced this issue locally, which is good but also fills me with self-loathing because it obviously means I didn't properly run the TCK prior to submitting the PR.  So... apologies for that.

Will work with Marc to resolve the problem.  Originally we had talked about never attempting any classloading, but instead using jandex for everything and just throwing up our hands for the cases where a class is encountered but doesn't exist in jandex.  I think that changed recently so that we could better support generating appropriate schemas for some standard Java classes like List and Map.

In any case, we need to better handle the situation where a class is encountered but is not in jandex *and* not available to the classloader.  This happens in the TCK thanks to the Scan* tests (which limit the information in jandex based on MPConfig settings).  But I can imagine it happening more broadly as well, so I think it needs to be handled in the code (rather than expand the jandex index).

msavy:
I thought incorrectly I handled this so my bad and will fix today.

msavy:
This is resolved and will push the small change shortly, just double and triple checking the tests 😄

kenfinnigan:
Looks like OpenAPI TCK passed, but the job failed because of:

`Failed pattern wfswarm\S+[0-9]{5,}.\S{5,} in directory /root/workspace/wildfly-swarm-pull-request-linux./root/workspace/wildfly-swarm-pull-request-linux/wfswarmarquillian9131380346002391809-swarm.jar size(13623159) too large. Max. is 0`

Looks like there's temp files left around. Are the test jars being exported perhaps?

msavy:
@kenfinnigan, are those enforcer name checks only on CI or with extra flags 🤔? My build passes locally and I can't find files with those name patterns in my build or temp.  Intriguing!

I'm using macOS.

kenfinnigan:
The build command is only `-B clean install -U -P microprofile-tck`, so doesn't look like it.

May need to get @juangon to take a look as he wrote the enforcer plugin

juangon:
So if it fails in Linux too it's a "good" fail. Maybe in those new tests server isn't shut down gracefully?

juangon:
What I don't understand is why @msavy can't reproduce locally in Linux too.

msavy:
I'm on macOS. Trying a Linux VM now.

msavy:
I see the failures now with refreshed snapshots but it seems to be on `testsuite-remoting`?

`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce (enforce) on project testsuite-remoting: Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed. -> [Help 1]`

Log:
https://drive.google.com/file/d/1BUI5nPOfg4ussic2gSu1I7-qC7w6wT3s/view?usp=sharing

 Maybe an updated snapshot somewhere causing issues, hmm? Strange...

juangon:
Strange indeed. I've been looking at test results previously and no failure was caused by this temp file checking, so I think it should be caused by this pull.

Additionally all temp files should be removed after every swarm:stop in Maven, so this temp files had been created by this test.

Are there any change you made that can affect testsuite-remoting?

msavy:
Perhaps an updated snapshot has changed behaviour between the previous builds and now? I notice that some commits relating to temp files have been made on master since we started this work (e.g. https://github.com/wildfly-swarm/wildfly-swarm/commit/846c1102088e367dc620e0f23c4ba7934c99c052). I might try applying it and seeing if the error goes away.

I can't see anywhere that we have touched testsuite-remoting (it's completely unrelated to this work).

Unfortunately top level builds take about 1h20m for me, so it's going to take a while to narrow it down!

BTW, I intermittently see testsuite-faulttolerance fail. Perhaps this is a known issue? https://drive.google.com/open?id=1J5YeliJezJTGIxIsQ5nYbB4iJ_RqKF5Y

juangon:
I think there is only 1 commit lately for the temp files (other ones were months ago). And those temp files are created using the new swarm prefix "wfswarm" so I think you should have all changes, unless something weird happened during your rebase...

About the other issue, yes, it's a known issue not related to temp files also I guess .

mkouba:
> BTW, I intermittently see testsuite-faulttolerance fail. 

@msavy I don't think this is a known issue. I'll look into this.

@juangon maybe OT - from my recent experience it seems that if you run `wildfly-swarm:run` and the deployment fails the `tmp` dir is not cleaned up properly and the swarm process hangs.

juangon:
@mkouba thanks, yes I think that could be one of the reasons. Files won't be deleted unless server is stopped.

msavy:
This branch does not contain 846c1102088e367dc620e0f23c4ba7934c99c052 if that might be relevant.

I've double-checked the code and we never touch testsuite-remoting.

Would you expect us to be able to trigger the enforcer checks from the testsuite's subdirectories or does it only work from top level builds?

When I execute the build and tests from `testsuite/microprofile-tcks/openapi` or `testsuite/testsuite-remoting` everything passes.

msavy:
@mkouba I've spent the morning and a bit of this afternoon trying to look at this:

AFAICT, it's not likely to do with our changes (the test that failed in CI is in another module we didn't touch). I've rebuilt a few times and the enforcer error only seems to trigger intermittently (successful example: https://drive.google.com/file/d/1X9rZeqVc1-BYC8CfbHkPVgoVkDcu-a_5/view?usp=sharing).

It seems to happen most often when you do a full top-level build with latest snapshots using `mvn clean install -U -P microprofile-tck` (i.e not from subdirectories). When running each testsuite submodule individually it always seem to pass without issue.

I think this behaviour perhaps supports your theory about the source of the problem?

(As noted in previous replies, 846c110 isn't on this PR, so might be worth adding that and trying again -- takes me over 2 hours to build on this machine though.)

We'll await your feedback. Still very much open minded!

mkouba:
@EricWittmann @msavy I think we should rebase this PR first. @juangon Is there a way to find out where those swarm uberjars come from, e.g. temporary archive artifacts from the workspace or something like that?

juangon:
@mkouba It seems those files are created here: https://github.com/wildfly-swarm/wildfly-swarm/blob/master/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java#L278.

BTW I think that is the only place I left .deleteOnExit. Maybe I should change that but that doesn't explain why it was working fine on CI and my env before this pull.

juangon:
@msavy I think I tested both mvn clean install from root, and for each testsuite directory getting consistent fails or successes.

mkouba:
@juangon Hm, shouldn't be the file created in temp dir? I mean the `java.io.tmpdir` system property seems to be overriden as the path reads e.g.:
> /root/workspace/wildfly-swarm-pull-request-linux./root/workspace/wildfly-swarm-pull-request-linux/wfswarmarquillian2812756886494495426-swarm.jar

In any case, `File.deleteOnExit()` does not seem to be reliable enough. We should probably delete all files explicitly?

juangon:
@mkouba it should as it uses File.createTempFile, which uses java.io.tmpdir AFAIK.

Yes I agree about deleteOnExit, in fact I removed everywhere, except here as I thought it wouldn't be necessary, as they are executable jars and not Swarm libraries. I will make a PR on Monday or this weekend.

Anyway, why it was working before this pull? :-)

EricWittmann:
I'm going to create a new PR that is rebased against the latest master.  I had trouble rebasing this one because git can be a POS.

EricWittmann:
Note:  I'll try to run everything on my centos build machine (this time) before submitting the PR.  So it may be awhile as I've never built Swarm on that machine before.

mkouba:
> Note: I'll try to run everything on my centos build machine...

It might as well explode before the build is finished ;-)

EricWittmann:
That's OK - the machine is in my basement so I won't be injured...  :)

EricWittmann:
BTW @juangon I wanted to add that I don't think there is much plausibility that this PR broke temp file cleanup.  Here is the list of files changed by this PR:

https://gist.github.com/EricWittmann/538f73f6b9f1dc6ae8162f59364282c1

So yeah, there's still the question of how the temp file thing is broken for this PR but not others.  Dunno.

EricWittmann:
OK my new PR branch is done.  `mvn clean install` seems to work OK.  I'll submit a new PR and we can see how it goes...

EricWittmann:
https://github.com/wildfly-swarm/wildfly-swarm/pull/818

juangon:
@EricWittmann @mkouba are the Microprofile changes the same as the https://github.com/wildfly-swarm/wildfly-swarm/pull/818 ?

In that case, as I couldn't reproduce the issue in local Windows, maybe we can try re-test this?

EricWittmann:
@juangon Yes the Microprofile changes are the same between this and #818 but the latter has been rebased on master.  So 818 has your temp file changes as well as an unknown number of other differences (whatever changes were committed to master between #810 and #818).

@kenfinnigan Closing this as out of date is fine with me.  At this point I don't really have any thoughts on why CI is failing or what the appropriate approach is to resolving the ongoing build issues. :(

juangon:
@EricWittmann  please can you rebase this one with latest master, then push --force?

EricWittmann:
@juangon rebasing this is a pain in the ass, but I think @msavy may already have something.  If not I'll rebase - it just may take a little while...

juangon:
@EricWittmann  why is a pain? It does conflict with something?

EricWittmann:
Mostly my lack of git rebase expertise I suspect. :)  But there are a few pom.xml conflicts that have to be resolved repeatedly.

juangon:
Ok if you want I can make a PR with my change commit reverted and then check CI.

EricWittmann:
Give me a few minutes and I"ll let you know. :)

juangon:
Ok thanks @EricWittmann

EricWittmann:
@juangon  -  OK I force pushed a rebased PR that excludes your latest temp-file commit.  Somehow the jandex version number got incremented from 2.0.3 to 2.0.4 and I missed that during my review of the rebase prior to force pushing it.  Not sure what that's about.  Hopefully that doesn't affect the testing.  I can revert it later if needed.

Otherwise I ***think*** everything was rebased properly.

mkouba:
@EricWittmann Just FYI jandex was downgraded to 2.0.3 (to align with version used in WildFly) and I think it should not affect your code:
https://github.com/wildfly-swarm/wildfly-swarm/commit/480c95a8dce46c04d6781027015f8fb368bd89a1

juangon:
This is funny. Now testsuite AJP is failing on Windows without any new change. :-(. @bobmcwhirter would be possible to clean windows temp file directory and re launch windows CI test?

kenfinnigan:
Closing as out of date

PR #813, opened 2018-02-27
SWARM-1833: Improve index available in DeploymentProcessor.

mkouba:
Motivation
----------
Currently, the index only contains classes of the top-level
ClassContainer, e.g. WEB-INF/classes for a WAR. Also we don't load/reuse
attached indexes.

Modifications
-------------
Modified DeploymentProducer.index() producer method.

Result
------
DeploymentProducer.index() returns a composite annotation index
(incl. dependencies, e.g. WEB-INF/lib) and attached indexes (located in
META-INF/jandex.idx) are reused.

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/runtime/cdi/configurable/DeploymentProducer.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/runtime/cdi/configurable/DeploymentProducer.java
@@ -16,19 +16,31 @@
 package org.wildfly.swarm.container.runtime.cdi.configurable;
 
 import java.io.IOException;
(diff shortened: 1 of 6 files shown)


mkouba:
Linux failure is:
> java.lang.AssertionError: <''> should be equal to :<'Management Interface'> ignoring case
	at org.wildfly.swarm.management.console.ManagementConsoleTest.testHomePageShouldHaveCorrectTitle(ManagementConsoleTest.java:47)

But the test passes locally.

kenfinnigan:
Weird, that one usually only fails on Windows!

kenfinnigan (approved):

PR #816, opened 2018-03-02
SWARM-1794: MP FT - remove CommandInterceptorTest.testTimeoutFallback().

mkouba:
- this test was replaced with
org.wildfly.swarm.microprofile.faulttolerance.deployment.timeout.TimeoutTest

### DIFF
--- a/fractions/microprofile/microprofile-fault-tolerance/src/test/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/basic/CommandInterceptorTest.java
+++ b/fractions/microprofile/microprofile-fault-tolerance/src/test/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/basic/CommandInterceptorTest.java
@@ -64,13 +64,6 @@ public class CommandInterceptorTest {
         assertTrue(MyFallbackHandler.DISPOSED.get());
     }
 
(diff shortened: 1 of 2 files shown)


kenfinnigan (approved):

kenfinnigan:
Passed on Linux but on Windows an FT TCK test failed.

Is it a problem or just an anomaly?

mkouba:
That's a known issue - `BulkheadAsynchRetryTest` fails intermittently, esp. on windows. Antoine was looking into this but so far we don't have a clue what's wrong. I'll create a jira issue and take it over.

antoinesd:
Unable to reproduce the error in my dev env (macos + JDK 1.8.0_152) after more than 30 attempts. I'll probably have to install a windows VM and study this issue in it.

mkouba:
@antoinesd I'd guess it's a similar problem as with `BulkheadSynchRetryTest`. Let me check this particular test as well.

mkouba:
Ok so looking at the test output and source it seems that `TestData.check()` is invoked before most of the async `BulkheadRapidRetry55MethodAsynchBean.test()` wrappers are executed. This could happen for many reasons (e.g. the thread pool init may take some time). I think the test should be fixed (and excluded for now). I'm going to create a separate FT issue.

PR #822, opened 2018-03-05
Use jandex 2.0.3

heiko-braun:


mkouba:
What's the reason for the downgrade?

heiko-braun:
Version alignment in prod. Do you know of a specific reason to use `2.0.4` instead?

mkouba:
No specific reason. I'm just wondering why 2.0.4 was used before because it seems that wildfly 11 is also using 2.0.3.

PR #825, opened 2018-03-06
[SWARM-1754] Warning only in case of the multiple factories

sberyozkin:
Removed the service loader resources given they refer to the default principal factory
Introduced an internal exception to help with deciding when to warn on the exception
Loading the Default factory directly if no custom factory has been found

### DIFF
--- a/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/principal/JWTCallerPrincipalFactory.java
+++ b/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/principal/JWTCallerPrincipalFactory.java
@@ -55,7 +55,7 @@ public abstract class JWTCallerPrincipalFactory {
                     newInstance = loadSpi(cl);
                 }
                 if (newInstance == null) {
-                    throw new IllegalStateException("No JWTCallerPrincipalFactory implementation found!");
+                    newInstance = new DefaultJWTCallerPrincipalFactory();
                 }
 
                 instance = newInstance;
@@ -86,17 +86,17 @@ public abstract class JWTCallerPrincipalFactory {
             try {
                 for (JWTCallerPrincipalFactory spi : sl) {
                     if (instance != null) {
-                        throw new IllegalStateException(
-                                "Multiple JWTCallerPrincipalFactory implementations found: "
-                                        + spi.getClass().getName() + " and "
-                                        + instance.getClass().getName());
+                        throw new MultipleFactoriesException(spi.getClass().getName());
                     } else {
                         log.debugf("sl=%s, loaded=%s", sl, spi);
                         instance = spi;
                     }
                 }
+            } catch (MultipleFactoriesException e) {
+                log.warn("Multiple JWTCallerPrincipalFactory implementations found: "
+                          + e.getSpiClassName() + " and " + instance.getClass().getName());
             } catch (Throwable e) {
-                log.warn("Failed to locate JWTCallerPrincipalFactory provider", e);
+                log.debugf("Failed to locate JWTCallerPrincipalFactory provider, e=%s", e);
             }
         }
         return instance;
@@ -119,4 +119,14 @@ public abstract class JWTCallerPrincipalFactory {
      * @throws ParseException on parse or verification failure.
      */
     public abstract JWTCallerPrincipal parse(String token, JWTAuthContextInfo authContextInfo) throws ParseException;
+
+    private static class MultipleFactoriesException extends Exception {
+        String spiClassName;
+        MultipleFactoriesException(String spiClassName) {
+            this.spiClassName = spiClassName;
+        }
+        String getSpiClassName() {
+            return spiClassName;
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


mkouba:
add to whitelist

mkouba:
test this please

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/principal/JWTCallerPrincipalFactory.java
```diff
@@ -86,17 +86,17 @@ private static JWTCallerPrincipalFactory loadSpi(ClassLoader cl) {
...
                         log.debugf("sl=%s, loaded=%s", sl, spi);
                         instance = spi;
                     }
                 }
+            } catch (MultipleFactoriesException e) {
+                log.warn("Multiple JWTCallerPrincipalFactory implementations found: "
+                          + e.getSpiClassName() + " and " + instance.getClass().getName());
             } catch (Throwable e) {
-                log.warn("Failed to locate JWTCallerPrincipalFactory provider", e);
+                log.debugf("Failed to locate JWTCallerPrincipalFactory provider, e=%s", e);
```
mkouba: I believe the `java.util.ServiceLoader` is implemented lazily. So if `ServiceLoader` fails to load a service provider (during the first iteration) the information will be lost unless you have the debug level enabled.

sberyozkin: Thanks, good catch; hmm, I wonder if I should update the PR to clone the iterator first, before entering the try section, otherwise, without rewriting the loading sequence, it is hard to find out in the catch if it was a cast issue related to the use of the parent loader or not; let me try...

sberyozkin:
I've updated the PR a bit, the cloning idea I proposed would not work either, but delaying the typed assignment will avoid a harmless cast exception while catching the real ones if any. Hopefully that is now safe enough.

heiko-braun:
test this please

heiko-braun:
add to whitelist

mkouba:
The MP FT TCK test which fails on windows should be excluded - see also https://github.com/wildfly-swarm/wildfly-swarm/pull/827.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/principal/JWTCallerPrincipalFactory.java
```diff
@@ -84,17 +84,19 @@ private static JWTCallerPrincipalFactory loadSpi(ClassLoader cl) {
...
+                        if (instance != null) {
+                            throw new MultipleFactoriesException(spi.getClass().getName());
+                        } else {
+                            log.debugf("sl=%s, loaded=%s", sl, spi);
+                            instance = (JWTCallerPrincipalFactory)spi;
+                        }
                     }
                 }
+            } catch (MultipleFactoriesException e) {
+                log.warn("Multiple JWTCallerPrincipalFactory implementations found: "
```
michalszynkiewicz: What do we benefit from throwing the exception and catching it instead of printing this warning directly?

sberyozkin: Sorry, missed this comment. I introduced it initially when trying to avoid the ambiguity at the catch level so that not to log harmless warnings which is less relevant now with the PR update (after Martin pointed out at the lazy ServiceLoader.iterator behavior).
During the 2nd update I was actually thinking about reverting the new exception class as well, but  decided that with the dedicated exception there will be a single warning only in case of the multiple factories issue as opposed to two if I revert it (within the for loop and then in the catch given that the for loop would have to be exited with the exception - where it would also be somewhat confusing because the 2nd warning would say a provider failed to be loaded which would not be true in case of the multiple factories issue). 
Does it sound reasonable ?

sberyozkin: The original code does not warn within the loop, it only throws ISE which is then logged. I can revert to that, though again a user would see something like "Failed to load the provider: multiple implementations found" which is somewhat contradictory. Either way, I can either revert the new class or keep it :-)

michalszynkiewicz: It can stay as it is, it just looks to me that the code below would have the same effect with a bit less code.
But maybe I'm missing sth :)
```
if (instance != null) {
   log.warn("Multiple JWTCallerPrincipalFactory implementations found: " + spi.getClass().getName() and ...")
   break;
}
```

sberyozkin: Sure, sounds good to me, 'break' works well. If Martin does not object then I'll remove the newly introduced exception and simplify. Thanks :-)

mkouba: +1

kenfinnigan (approved):

PR #828, opened 2018-03-07
SWARM-1326 Swarm brings weird Arquillian versions transitively

johnAls:
The different version seems like its picked up by wildfly-parent, the easy solution is make sure the arquillian-bom gets precedence. 

Or is this a dangerous path to go it may hard to overlook what other transitive dependencies that can be affected? 

-----

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -141,6 +141,14 @@
       </dependency>
 
       <dependency>
+        <groupId>org.jboss.arquillian</groupId>
+        <artifactId>arquillian-bom</artifactId>
+        <version>${version.org.arquillian}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+
+      <dependency>
         <groupId>org.wildfly</groupId>
         <artifactId>wildfly-parent</artifactId>
         <version>${version.wildfly}</version>
--- a/testsuite/microprofile-tcks/pom.xml
+++ b/testsuite/microprofile-tcks/pom.xml
@@ -38,6 +38,9 @@
     <properties>
         <!-- Do not upgrade until https://issues.jboss.org/browse/ARQ-2144 is fixed -->
         <version.testng>6.9.9</version.testng>
+        <!-- When upgrading testng, make sure to use swarm version of arquillian -->
+        <version.arquillian.testng>1.1.13.Final</version.arquillian.testng>
+        
         <!-- Overriding 2.19 due to https://issues.apache.org/jira/browse/SUREFIRE-1252 -->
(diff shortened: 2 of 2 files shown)


Ladicek:
I agree it's possible that some Arquillian transitive dependencies will override some WildFly transitive dependencies, but not sure how to find out. Doing a `mvn dependency:tree` diff would be hard, considering the number of Maven modules...

At the same time, I just tried my reproducer (https://github.com/Ladicek/swarm-arquillian-versions) with 2018.3.0 and the problem is still present:

```
!!! Arquillian Core API = 1.1.15.Final
!!! Arquillian Core SPI = 1.1.13.Final
!!! Arquillian Core Implementation Base = 1.1.13.Final
!!! Arquillian Container SPI = 1.1.13.Final
!!! Arquillian Container Implementation Base = 1.1.13.Final
!!! Arquillian Config API = 1.1.13.Final                                                                                                                                  
!!! Arquillian Config Implementation Base = 1.1.15.Final                                                                                                                  
!!! Arquillian Container Test API = 1.1.13.Final                                                                                                                          
!!! Arquillian Container Test SPI = 1.1.13.Final                                                                                                                          
!!! Arquillian Container Test Implementation Base = 1.1.13.Final                                                                                                          
!!! Arquillian Test API = 1.1.13.Final                                                                                                                                    
!!! Arquillian Test SPI = 1.1.13.Final                                                                                                                                    
!!! Arquillian Test Implementation Base = 1.1.13.Final
!!! Arquillian TestRunner JUnit Core = 1.1.13.Final
!!! Arquillian TestRunner JUnit Container = 1.1.13.Final
!!! Arquillian TestEnricher CDI = 1.1.13.Final
!!! Arquillian TestEnricher EJB = 1.1.13.Final
!!! Arquillian TestEnricher InitialContext = 1.1.13.Final
!!! Arquillian TestEnricher Resource = 1.1.13.Final
```

I'm obviously in favor of fixing this :-)

kenfinnigan:
Looks like this change breaks a lot of the MicroProfile TCKs, so that would need to be investigated further

johnAls:
In the arquillian-core 1.1.15.Final the requirement for TestNg changed from 6.9.10 as it was in 1.1.13.Final to 6.13. And in MicroProfile TCK we are using TestNg 6.9.9 due to ARQ-2144 so yes we get a NoSuchMethodError exception.

So the issue is pointing on a real problem we should assure that we use the same version of arquillian, so maybe a separate issue to fix microprofile tck to a higher version of arquillian-testng-container and at the moment let it use 1.1.3.Final

kenfinnigan (approved):

PR #833, opened 2018-03-09
SWARM-1886: Make use of repackaged war for wildfly-swarm:run.

mkouba:
Motivation
----------
Currently, StartMojo only specifies the swarm.app.name property and
repackaged war is ignored. Instead, Swarm creates a new war and adds
a bunch of project dependencies which are normally removed from the war.
See also DependenciesContainer.ALL_DEPENDENCIES_MARKER.

Modifications
-------------
Modified StartMojo to specify swarm.app.path property and make use
of the repackaged war.

Result
------
When executing app with Maven Plugin and wildfly-swarm.useUberJar=false
(default) the repackaged war is used and some weird classloading issues
should be gone.

-----

### DIFF
--- a/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StartMojo.java
+++ b/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StartMojo.java
@@ -200,7 +200,11 @@ public class StartMojo extends AbstractSwarmMojo {
             finalName = finalName + ".war";
         }
 
-        return executor(Paths.get(this.projectBuildDir, finalName), finalName, false);
+        Path warPath = Paths.get(this.projectBuildDir, finalName);
+        SwarmExecutor executor = executor(warPath, finalName, false);
+        // Specify swarm.app.path property so that repackaged war is used
+        executor.withProperty(BootstrapProperties.APP_PATH, warPath.toString());
+        return executor;
     }
 
 


mkouba:
BTW without this change the content of the `WEB-INF/lib` of the war deployed (see also SWARM-1886 reproducer) looks like this https://gist.github.com/mkouba/0e18a525393bc89c48c1858c5cabc7f9 (pretty scary I think)

famod:
It's been a while since last using Swarm (April 2017) but in our team we regularly used the plugin goal `run` with `useUberJar=false` bound to the very early maven phase `initialize` because it was much faster than packaging the actual war/jar.
It seems to me that with this change such a workflow would no be possible anymore?

PS: I do know that `useUberJar=false` is (still?) buggy, there a quite a few open tickets in JIRA, mostly older ones.
PPS: After rethinking this change, it might not be as problematic as I first thought: The `run` goal forks the maven lifecycle up to phase `package` so the properly repacked war _should_ be available and the fat jar has not yet been packaged, no?

kenfinnigan:
@mkouba, can you comment on @famod's question?

mkouba:
@kenfinnigan I'm not sure to be honest. @famod It would be great if you could test this PR with your setup. Anyway, I think that packaging the actual war/jar should not be a performance problem.

famod:
I won't be able to test this PR anytime soon due to lack of time (especially as the project in question is not under active development anymore and would need to be updated from 2017.3.1 first).
So I'd say just go ahead a merge the change. Once I get around to using Swarm again I'll come across this anyway and in the unlikely event of a problem regarding this change I will file an issue ticket.

kenfinnigan (approved):

PR #834, opened 2018-03-14
Manage xnio to 3.5.4.Final and javax.activation to 1.1.1

heiko-braun:

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -81,6 +81,8 @@
     <version.zipkin.reporter>0.6.1</version.zipkin.reporter>
     <version.zipkin>1.13.1</version.zipkin>
 
+    <version.xnio>3.5.4.Final</version.xnio>
+
     <!-- OpenTracing related versions -->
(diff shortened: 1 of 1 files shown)


build-parent/pom.xml
```diff
@@ -350,6 +354,25 @@
...
+        <version>${version.xnio}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>org.jboss.xnio</groupId>
+        <artifactId>xnio-nio</artifactId>
+        <version>${version.xnio}</version>
+      </dependency>
+
+      <dependency>
```
heiko-braun: ?

kenfinnigan: openshift-restclient-java dependency is already present, this is a duplication

kenfinnigan:
What's the reason that these need to have their versions managed?

We're simply pulling them from WF 11 directly.

If there's going to be lots of these managements necessary, they should be made to -prod only

PR #835, opened 2018-03-14
SWARM-1790: MicroProfile + JPA Hollow Jar

kenfinnigan:

-----

Motivation
----------
As MicroProfile doesn't offer any kind of persistence at present, it would be good to offer a Hollow jar that combines MicroProfile and JPA for that purpose.

Modifications
-------------
Removed `ManifestComparitor` as it was causing comparison inconsistencies and was not really serving a purpose.

Add `microprofile-jpa` hollow jar.

Modify `ApplicationModuleFinder` to allow adding arbitrary libraries to the `swarm.application` module by using the `swarm.classpath` system property. This makes it possible for Hollow Jars to specify a path for JDBC Drivers such that they can be auto discovered.

For example, this new Hollow Jar can be run with:

  java -Dswarm.classpath=./h2-1.4.196.jar -jar microprofile-jpa-hollow-swarm.jar myApp.war

and the H2 Driver will be auto discovered.

Result
------
Added MicroProfile + JPA Hollow Jar to build, and added support for -Dswarm.classpath to specify jars to be added to application module classloader.


juangon:
It's pretty interesting it passed temp files check even when adding a new JarFile(...) without using JarFileManager for closing it afterwards

docs/reference/configuration.adoc
```diff
@@ -67,6 +67,22 @@ To use a property on the command line, pass it as a command-line parameter to th
 $ java -Dswarm.bind.address=127.0.0.1 -jar myapp-swarm.jar
 ----
 
+[discrete]
```
tradej, line 70: There are several issues with the content, I'll send a pull request against `kenfinnigan:SWARM-1790` today (Mar 15), please do not merge yet if possible.

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/ApplicationModuleFinder.java
```diff
@@ -185,5 +191,25 @@ protected void addDependencies(ModuleSpec.Builder builder, ApplicationEnvironmen
                 });
     }
 
+    private void addJDBCDrivers(ModuleSpec.Builder builder) throws IOException {
```
Ladicek: I understand the primary motivation is JDBC drivers, but it's actually pretty universal. The name is slightly misleading :-)

kenfinnigan:
@juangon, it's because it's not a temp file.

All it's doing is loading up a JAR file that's already on the file system, not exploding out one into a temp location

docs/reference/configuration.adoc
kenfinnigan: Damn, thought I'd done well to document! Oh well, let me know where I messed up for next time

juangon:
Ok thanks @kenfinnigan . I think it was too much time looking for that String by me.

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/modules/ApplicationModuleFinder.java
kenfinnigan: That's when that was the original purpose, but it became more general purpose than JDBC. Will fix the name

Ladicek (approved):

PR #836, opened 2018-03-14
SWARM-1895: [Swagger] Align library & module dependencies.

roguexz:
Motivation
----------
The Swagger fraction is pulling in an older version of the Guava dependency (18.0) while the Maven project lists the provided dependency to "20.0". This causes weird issues when using the Gradle plugin for building a "bundled" uber jar file as the dependency is never satisfied. Instead of downgrading the Guava dependency or messing around with the swarmtool / gradle plugin, it makes more sense to add an explicit dependency on the Guava module instead of keeping track / updating the dependency version for each release.

Additionally, at some point in time, if we have the ability to import POM properties of a BOM, we can reuse the library versions easily by referring to the appropriate property names. At a minimum, by aligning the property names, it will help in ensuring that we are copying over the same values from the appropriate bom.

Modifications
-------------
* Refer to the Guava dependency module instead of managing the dependency on our own.
* Align the Jackson & Javassist library versions with those present in WF parent.
* Align the property name for library versions with those present in WF parent.

Result
------
Uber-bundled jars built using the Gradle plugin will no longer complain of missing Javassist, Guava & Jackson libraries.

-----

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,7 @@ catalog.xml
 #
 maven-ant-tasks.jar
 test-output
-transaction.log
+
 # vim files
 *.swp
 /.gitk-tmp.*
@@ -35,10 +35,6 @@ atlassian-ide-plugin.xml
 pom.xml.versionsBackup
 # hprof dumps
 /*.hprof
-# ignore 'randomly' strewn around logs
-server.log
-# ignore java crashes
-hs_err_pid*.log
 # H2 databases produced by tests
 *.h2.db
 *.trace.db
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -42,17 +42,17 @@
     <version.jboss-logmanager-ext>1.0.0.Alpha3</version.jboss-logmanager-ext>
     <!-- Weld SE version must be aligned with the version used in WildFly -->
     <version.weld-se>2.4.3.Final</version.weld-se>
+    <!-- Align dependency versions with those present in WildFly parent -->
+    <!-- Doing this explicitly since there is no easy way to import project properties from a pom file -->
+    <version.com.fasterxml.jackson>2.8.9</version.com.fasterxml.jackson>
 
     <version.org.objectweb.asm>6.0</version.org.objectweb.asm>
 
     <version.jolokia>1.3.4</version.jolokia>
     <version.io.swagger>1.5.16</version.io.swagger>
     <version.org.webjars.swagger-ui>3.2.2</version.org.webjars.swagger-ui>
-    <version.swagger.jackson>2.8.4</version.swagger.jackson>
     <version.swagger.reflections>0.9.10</version.swagger.reflections>
     <version.swagger.slf4j>1.6.3</version.swagger.slf4j>
-    <version.swagger.guava>18.0</version.swagger.guava>
-    <version.swagger.javassist>3.18.2-GA</version.swagger.javassist>
     <version.swagger.commons.lang>3.2.1</version.swagger.commons.lang>
 
     <version.hibernate-spatial>5.0.10.Final</version.hibernate-spatial>
--- a/fractions/swagger/pom.xml
+++ b/fractions/swagger/pom.xml
@@ -95,6 +95,19 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.wildfly</groupId>
+      <artifactId>wildfly-feature-pack</artifactId>
+      <type>zip</type>
+      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
     <!-- Meta SPI -->
     <dependency>
       <groupId>org.wildfly.swarm</groupId>
--- a/fractions/swagger/src/main/resources/modules/com/fasterxml/jackson/swagger/module.xml
+++ b/fractions/swagger/src/main/resources/modules/com/fasterxml/jackson/swagger/module.xml
@@ -1,18 +1,17 @@
 <module xmlns="urn:jboss:module:1.3" name="com.fasterxml.jackson" slot="swagger">
   <resources>
-    <artifact name="com.fasterxml.jackson.core:jackson-core:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.core:jackson-annotations:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.core:jackson-databind:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.datatype:jackson-datatype-joda:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.datatype:jackson-datatype-guava:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.dataformat:jackson-dataformat-xml:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.jaxrs:jackson-jaxrs-base:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider:${version.swagger.jackson}"/>
+    <artifact name="com.fasterxml.jackson.datatype:jackson-datatype-joda:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.datatype:jackson-datatype-guava:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.dataformat:jackson-dataformat-xml:${version.com.fasterxml.jackson}"/>
   </resources>
 
   <dependencies>
     <module name="javax.ws.rs.api"/>
     <module name="javax.api"/>
+    <module name="com.fasterxml.jackson.core.jackson-core" export="true"/>
+    <module name="com.fasterxml.jackson.core.jackson-databind" export="true"/>
+    <module name="com.fasterxml.jackson.core.jackson-annotations" export="true"/>
+    <module name="com.fasterxml.jackson.jaxrs.jackson-jaxrs-json-provider" export="true"/>
   </dependencies>
 </module>
--- a/fractions/swagger/src/main/resources/modules/io/swagger/main/module.xml
+++ b/fractions/swagger/src/main/resources/modules/io/swagger/main/module.xml
@@ -5,14 +5,14 @@
     <artifact name="io.swagger:swagger-models:${version.io.swagger}"/>
     <artifact name="io.swagger:swagger-annotations:${version.io.swagger}"/>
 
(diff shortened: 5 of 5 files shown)


.gitignore
```diff
@@ -36,9 +36,7 @@ pom.xml.versionsBackup
 # hprof dumps
 /*.hprof
 # ignore 'randomly' strewn around logs
-server.log
 # ignore java crashes
-hs_err_pid*.log
```
roguexz: There is already a *.log on line 57.

build-parent/pom.xml
```diff
@@ -42,17 +42,18 @@
     <version.jboss-logmanager-ext>1.0.0.Alpha3</version.jboss-logmanager-ext>
     <!-- Weld SE version must be aligned with the version used in WildFly -->
     <version.weld-se>2.4.3.Final</version.weld-se>
+    <!-- Align dependency versions with those present in WildFly parent -->
+    <!-- Doing this explicitly since there is no easy way to import project properties from a pom file -->
+    <version.com.fasterxml.jackson>2.8.9</version.com.fasterxml.jackson>
+    <version.org.javassist>3.20.0-GA</version.org.javassist>
```
roguexz: Using the same property name as is available in WF parent. This way, in case someone builds a Maven plugin that exports (filtered) properties from an imported BOM, we can remove these values as they will be inherited.

fractions/swagger/src/main/resources/modules/io/swagger/main/module.xml
```diff
@@ -5,14 +5,14 @@
...
-    <artifact name="com.google.guava:guava:${version.swagger.guava}"/>
     <artifact name="org.reflections:reflections:${version.swagger.reflections}"/>
-    <artifact name="org.javassist:javassist:${version.swagger.javassist}"/>
+    <artifact name="org.javassist:javassist:${version.org.javassist}"/>
     <artifact name="org.apache.commons:commons-lang3:${version.swagger.commons.lang}"/>
 
   </resources>
 
   <dependencies>
+    <module name="com.google.guava"/>
```
roguexz, line 14: I figured since there is a general desire to reduce the overall deliverable size, I could help in a small way.

.gitignore
Ladicek: Should probably also remove the comments `# ignore 'randomly' strewn around logs` and `# ignore java crashes`.

fractions/swagger/src/main/resources/modules/io/swagger/main/module.xml
kenfinnigan: org.javassist should be picked up by its module and not the artifact directly, the same as you've done for guava

fractions/swagger/src/main/resources/modules/com/fasterxml/jackson/swagger/module.xml
```diff
@@ -1,14 +1,14 @@
...
-    <artifact name="com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider:${version.swagger.jackson}"/>
+    <artifact name="com.fasterxml.jackson.core:jackson-core:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.core:jackson-annotations:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.core:jackson-databind:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.datatype:jackson-datatype-joda:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.datatype:jackson-datatype-guava:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.dataformat:jackson-dataformat-xml:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.jaxrs:jackson-jaxrs-base:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider:${version.com.fasterxml.jackson}"/>
```
kenfinnigan: If these are now aligned with WF provided versions, it should switch to module dependencies instead of artifacts

build-parent/pom.xml
kenfinnigan: If the below changes are made, these properties shouldn't need to be added

kenfinnigan:
With these changes, did you verify that a Swagger app still works?

I know in the past there were issues in that Swagger relied on an older version of Jackson, which is why the versions were managed that way

roguexz:
Looking at the review comments. Will update the PR and notify.

roguexz:
Quick question. When compiling on OS X, I keep hitting the following error:

    Failed pattern wfswarm\S+[0-9]{5,}.\S{5,} in directory .../roguexz/wildfly-swarm/testsuite
        /testsuite-cdi-ejb/target.../roguexz/wildfly-swarm/testsuite/testsuite-cdi-ejb/
        target/wfswarmCdiEjbTest4291612659053267952.jar size(128) too large. Max. is 0

When do such exceptions show up? Is this a problem with code not cleaning up after the test completes? Or is this an indication of an error?

bobmcwhirter:
It means there’s some lingering temp file not being cleaned up. This can happen if a build failed earlier. So mvn clean first.

roguexz:
Running the same command: `mvn clean install`.

If I navigate to the project and execute the same command again, it works. Strange! Looks like some kind of a timing related issue probably! Went through SWARM-1530 and looks like I am hitting it intermittently.

roguexz:
@bobmcwhirter Does Arquillian apply a timeout during teardown? I have been digging through the source and am failing to locate the right line of code that needs to be changed. Here is what happens,

1. There are multiple instances (usually 2) created for TempFileManager - one which is used before the server starts and one which is used inside the server.
2. The temp files which are managed by the TempFileManager instance that is created inside the deployment get cleaned up properly.
3. The temp files which are created outside (e.g., NestedJarResourceLoader, etc.) are not getting cleaned up properly because "something" is killing the thread pool.

Relevant execution reference: [RuntimeServer](https://github.com/wildfly-swarm/wildfly-swarm/blob/master/core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java#L315)

Depending on the system load, the following happens:

    *** Files to be cleared: 9
    Deleting file: .../wildfly-swarm/testsuite/testsuite-batch-jberet/target/wfswarmwildfly-swarm372552088680020708.d
    Deleting file: .../wildfly-swarm/testsuite/testsuite-batch-jberet/target/wfswarmmodule-jar5302142567147266322.jar_d
    Deleting file: .../wildfly-swarm/testsuite/testsuite-batch-jberet/target/wfswarmmodule-jar780217234412223336.jar_d
    Deleting file: .../wildfly-swarm/testsuite/testsuite-batch-jberet/target/wfswarmmodule-jar4155687540380722092.jar_d
    Deleting file: .../wildfly-swarm/testsuite/testsuite-batch-jberet/target/wfswarmBatchArquillianTest4465195220280811649.jar
    Deleting file: .../wildfly-swarm/testsuite/testsuite-batch-jberet/target/wfswarmmodule-jar125643956863438708.jar_d
    Deleting file: .../wildfly-swarm/testsuite/testsuite-batch-jberet/target/wfswarmswarm-config-6256933340879466625.xml
    
    Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 10.675 sec - in org.wildfly.swarm.batch.jberet.BatchArquillianTest

As you can see, the code was supposed to delete 9 files, but it only managed to get to 7. The thread-pool was killed in between.

So the question I have is .. is there a timeout being applied to the initial thread-pool / CDI container that hosts the RuntimeServer? If so, who is shutting it down? Is there a way to specify a timeout for these test case scenarios?

If I were to resume the build from that project, it goes through fine till it crashes at some other place. This is quite annoying.

**Environment Details**
Apache Maven 3.5.2 (138edd61fd100ec658bfa2d307c43b76940a5d7d; 2017-10-18T00:58:13-07:00)
Maven home: /usr/local/Cellar/maven/3.5.2/libexec
Java version: 1.8.0_162, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.13.3", arch: "x86_64", family: "mac"

roguexz:
I think I might have found a solution for that compilation failure issue. Testing it a couple of more times before raising a separate PR for it.

roguexz:
Raised PR #839 for fixing the local compilation problem.

build-parent/pom.xml
roguexz: The datatypes being used in the Swagger fraction are not yet available through WildFly 11. So, I shall make the required changes to the Javassist module and test it.

fractions/swagger/pom.xml
```diff
@@ -95,6 +95,19 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
```
roguexz, line 98: @kenfinnigan I am not sure if this is the right way to bring in the WF modules for module resolution or not, but I saw this in other places and used the same. Please advise if this is incorrect.

fractions/swagger/src/main/resources/modules/com/fasterxml/jackson/swagger/module.xml
```diff
@@ -1,18 +1,17 @@
...
-    <artifact name="com.fasterxml.jackson.datatype:jackson-datatype-guava:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.dataformat:jackson-dataformat-xml:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.jaxrs:jackson-jaxrs-base:${version.swagger.jackson}"/>
-    <artifact name="com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider:${version.swagger.jackson}"/>
+    <artifact name="com.fasterxml.jackson.datatype:jackson-datatype-joda:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.datatype:jackson-datatype-guava:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:${version.com.fasterxml.jackson}"/>
+    <artifact name="com.fasterxml.jackson.dataformat:jackson-dataformat-xml:${version.com.fasterxml.jackson}"/>
   </resources>
```
roguexz, line 7: As mentioned earlier, not all modules are available from WF 11. Looks like the datatype modules that are present in WF 11 point to a project that has been deprecated and needs to be fixed.

fractions/swagger/src/main/resources/modules/com/fasterxml/jackson/swagger/module.xml
```diff
@@ -1,18 +1,17 @@
...
 
   <dependencies>
     <module name="javax.ws.rs.api"/>
     <module name="javax.api"/>
+    <module name="com.fasterxml.jackson.core.jackson-core" export="true"/>
```
roguexz, line 12: Based on the limited reading, I noticed that I will need to export the APIs for the Swagger module to consume. Let me know if this is supposed to be done differently.

roguexz:
Raised PR #840 to help with generating local builds easily.

@kenfinnigan Local test suite is working after these changes. I have confirmed that my Gradle based project is also starting up fine and I am able to view the `swagger.json` file as expected.

kenfinnigan (approved):

PR #837, opened 2018-03-16
SWARM-1792: skipping cache activation for unavailable caches

michalszynkiewicz:

-----
Motivation
----------
To activate the appropriate caches but get rid of error message complaining about the failed services when some caches are not available.

Modifications
-------------
`InfinispanCustomizer` and `CacheActivator` where modified to skip activating caches which are not available (their fractions are not added).
Also, undertow cache was renamed to web and jpa to hibernate. I believe these are the proper names of the cache containers.

Result
------
The error message is no longer printed in the log, yet, when possible, the required caches are started

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java
@@ -20,12 +20,14 @@ import java.io.IOException;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.StreamSupport;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Any;
@@ -198,7 +200,9 @@ public class RuntimeServer implements Server {
 
         List<ServiceActivator> activators = new ArrayList<>();
 
-        this.serviceActivators.forEach(activators::add);
+        StreamSupport.stream(serviceActivators.spliterator(), false)
+                .filter(Objects::nonNull)
+                .forEach(activators::add);
 
         activators.add(new ContentRepositoryServiceActivator(this.contentRepository));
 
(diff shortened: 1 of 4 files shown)


fractions/wildfly/infinispan/src/main/java/org/wildfly/swarm/infinispan/InfinispanMessages.java
```diff
@@ -0,0 +1,38 @@
...
+ * Date: 3/15/18
+ */
+@MessageLogger(projectCode = "WFSISPN", length = 4)
+public interface InfinispanMessages extends BasicLogger {
+
+    InfinispanMessages MESSAGES = Logger.getMessageLogger(InfinispanMessages.class, "org.wildfly.swarm.infinispan");
+
+    @LogMessage(level = Logger.Level.INFO)
+    @Message(id = 1, value = "Skipping activating cache configuration service: %s. The service is missing.")
+    void skippingCacheActivation(String key);
```
Ladicek: The entire PR looks very good to me, the only thing I wonder about is this log message. Should it really be on INFO? Should it say "the service is missing"? (I'm thinking it should rather be DEBUG and read "the corresponding fraction is not present".)

michalszynkiewicz: done

kenfinnigan:
Another approach would be to not amend `CacheActivator` and simply not call new if it's unsatisfied. Might be a bit simpler

michalszynkiewicz:
@kenfinnigan updated, I also squashed the commits not to put stuff into `CacheActivator` history

core/container/src/main/java/org/wildfly/swarm/container/runtime/RuntimeServer.java
```diff
@@ -198,7 +200,9 @@ public ExtensibleConfigurationPersister createConfigurationPersister(ServerEnvir
 
         List<ServiceActivator> activators = new ArrayList<>();
 
-        this.serviceActivators.forEach(activators::add);
+        StreamSupport.stream(serviceActivators.spliterator(), false)
```
kenfinnigan, line 203: Any particular reason for this usage vs `this.serviceActivators.stream()` ?

michalszynkiewicz: `Instance::stream` is not available in CDI 1.2, it was added in 2.0 https://docs.jboss.org/cdi/api/2.0/javax/enterprise/inject/Instance.html#stream--

kenfinnigan: Ah, I think I got confused with the `List` just above that code with a similar name. Forgot it was dealing with `Instance`.

Ignore me

gastaldi:
I can confirm that this PR solves the SWARM-1792 issue. Good work @michalszynkiewicz ! :beers:

kenfinnigan (approved):

PR #839, opened 2018-03-17
Allow the test cases to cleanup after themselves.

roguexz:
Motivation
----------
I have been having a tough time trying to get the code to build on my local machine. I kept hitting the dreaded,

    Some files does not fullfill provided pattern rules:

I found a few issues while debugging the code,

* The thread pools might not be given enough time to clean up.
* The "recursive file delete" operation does not provide enough hints when things go wrong.
* When running with the UberjarSimpleContainer, the process shuts down rather quickly (for slow machines).

Modifications
-------------
1. Ensure that executor pools use the awaitTermination method where applicable.
2. Change the recursive delete operation to use FileWalker instead of the current implementation. This will ensure that if there is an IO exception, it gets printed to the console.
3. Give the Swarm process up to 2 minutes before forcibly killing it.

Result
------
Now the code should build fine on the local laptop (decent hardware specs). This should take care of issues similar to SWARM-1530.

-----

### DIFF
--- a/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
+++ b/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
@@ -355,7 +355,7 @@ public class UberjarSimpleContainer implements SimpleContainer {
 
     @Override
     public void stop() throws Exception {
-        this.process.stop();
+        this.process.stop(2, TimeUnit.MINUTES); // Because my laptop is slower than yours.
         TempFileManager.deleteRecursively(workingDirectory);
     }
 
--- a/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/TempFileManager.java
+++ b/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/TempFileManager.java
@@ -17,7 +17,11 @@ package org.wildfly.swarm.bootstrap.util;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Collections;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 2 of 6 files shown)


arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
```diff
@@ -355,7 +355,7 @@ private void registerContainerFactory(Archive<?> archive, Class<?> clazz) {
 
     @Override
     public void stop() throws Exception {
-        this.process.stop();
+        this.process.stop(2, TimeUnit.MINUTES); // Because my laptop is slower than yours.
```
roguexz, line 358: Technically, this is all that I needed to do .. but then .. I spent so much time cleaning up at other places, I figured it would help at some time.

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/TempFileManager.java
```diff
@@ -66,26 +70,43 @@ private void register(File file) {
...
     public static boolean deleteRecursively(File f) {
         if (!f.exists()) {
             return false;
         }
-        if (f.isDirectory()) {
-            File[] children = f.listFiles();
-            for (int i = 0; i < children.length; ++i) {
-                if (!deleteRecursively(children[i])) {
-                    return false;
+
```
roguexz, line 82: Switched the logic from a traditional recursion based search to a FileVisitor based approach.

plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StopMojo.java
```diff
@@ -68,30 +69,14 @@ public void executeSpecific() throws MojoExecutionException, MojoFailureExceptio
...
+                Matcher matcher = tempFilePattern.matcher(tmpFile.getName());
                 if (matcher.matches()) {
-                    deleteRecursively(tmpFile);
+                    TempFileManager.deleteRecursively(tmpFile);
                 }
             }
         }
     }
 
-    public boolean deleteRecursively(File f) {
```
roguexz, line 79: The Maven plugin has a dependency on the `bootstrap` module. So, we don't really need to duplicate this code.

kenfinnigan:
ok to test

roguexz:
Tests seem to have passed. Can we merge this please?

heiko-braun (approved):

heiko-braun:
lgtm, thanks for the fix!

juangon:
@roguexz thanks a lot for your contribution. I am eager to check if this fixes other specific problems we are having around this for some test cases.

roguexz:
@juangon Happy to help. I spent so much time debugging this problem, I might be able to help if you have a sample stack trace or behavior that you are noticing :)

juangon:
I am on vacation until Wednesday. Will try this then. Thanks again!

PR #840, opened 2018-03-18
Simplify the POM structures

roguexz:
Motivation
----------
I was trying to test out some changes locally. In order to avoid any locally cached files, I decided to tweak the version literal and build a complete new set of files. As I would've done earlier (for other projects), I executed the command "mvn versions:set". Turns out, not all projects are managed by the same parent, so I still had modules whose version numbers hadn't changed. This means, that anyone trying to execute a similar exercise will have to rely on a find-and-replace-all capability instead of using predictable tooling.

The changes made in this commit improve the maintainability of the POM files.

In order to try out changes locally, all one needs to do is,

    $> mvn versions:set -DnewVersion=X.Y.Z
    $> mvn clean install

    Consume your newly built versions in your project.

Modifications
-------------
1. Replaced explicit version declarations for modules (in the dependencyManagement section) with ${project.version}
2. When referencing tools / binaries built from this very same project, make use of ${project.version} instead of the version literal.
3. Restructured the parent projects for a few of the modules.
4. Moved plugin configurations to the respective parent-modules.

Result
------
There is no impact to the project, unless there is some tooling (not listed within the project) that relies on the current POM design.

-----


roguexz:
The module's are structured as follows now,

    org.wildfly.swarm:parent:9
        |- org.wildfly.swarm:wildfly-swarm:VERSION
            |
            |- org.wildfly.swarm:wildfly-swarm-enforcer-pattern-size:VERSION
            |
            |- org.wildfly.swarm:build-resources:VERSION
            |
            |- org.wildfly.swarm:howto-parent:VERSION
            |
            |- org.wildfly.swarm:testsuite-parent:VERSION
            |
            |- org.wildfly.swarm:build-parent:VERSION
                |
                |- All other projects come under this hierarchy.

I had to move the plugins to the respective parent's so that we can re-use them across the various modules. I tried moving `testsuite-parent` under `build-parent` ... but that didn't turn out very nice. I need to spend more time looking at it. Will fix it in a subsequent check-in.

In order to verify that things are working fine, I did the following,

    $> mvn versions:set -DnewVersion=2018.4.0-t2-SNAPSHOT
    $> mvn clean install
        // Build completed successfully without any issues.

Can you please review and let me know if these changes would cause any problems?

Having these changes will allow people to make changes locally, test quickly and contribute faster.

build-parent/pom.xml
```diff
@@ -112,6 +112,86 @@
 
   <build>
     <plugins>
+      <plugin>
```
roguexz, line 115: The `failsafe` & `surefire` plugin present in the root POM were being used by `build-parent` & `testsuite-parent`. Since I was not able to figure out (in my limited weekend time) the problem with why `testsuite-parent` cannot be brought under `build-parent`, I am duplicating the plugin configurations in these two files for now. Will take care of fixing it in a subsequent PR.

build-parent/pom.xml
```diff
@@ -129,6 +209,13 @@
     </plugins>
   </build>
 
+  <dependencies>
```
roguexz, line 212: Since I am bringing `build-resources` under the `root` pom, I had to move this dependency to the child parent poms.

docs/howto/autodetect-fractions/pom.xml
```diff
@@ -89,7 +89,6 @@
       </plugin>
       <plugin>
         <artifactId>maven-failsafe-plugin</artifactId>
-        <version>${version.maven-failsafe-plugin}</version>
```
roguexz, line 92: We can use the version coming in from WF Parent directly .. or if need be, we can manage it from root POM.

docs/howto/pom.xml
```diff
@@ -7,17 +7,22 @@
...
+  <parent>
+    <groupId>org.wildfly.swarm</groupId>
+    <artifactId>wildfly-swarm</artifactId>
+    <version>2018.4.0-SNAPSHOT</version>
+    <relativePath>../../</relativePath>
+  </parent>
+
   <groupId>org.wildfly.swarm.howto</groupId>
   <artifactId>howto-parent</artifactId>
-  <version>2018.4.0-SNAPSHOT</version>
```
roguexz, line 12: This was a redundant declaration (as it would get inherited from the parent).

pom.xml
```diff
@@ -82,131 +82,61 @@
   </properties>
 
   <build>
-    <plugins>
```
roguexz, line 85: Moved the explicit plugin definition to `pluginsManagement` so that they can be pulled in by the respective parent POMs (only if needed).

kenfinnigan:
I'm not sure this change is something we'd want to incorporate.

/docs/how-to is deliberately out of the build hierarchy to make each of the howto's easy to take and consume outside the WF Swarm build process.

roguexz:
@kenfinnigan I actually considered the scenario where the `docs` module could become an independent project (from a maintenance perspective). So I did compile the code with the following changes as well,

1. Remove the parent definition of `howto-parent`
2. Rebuild the project.

The above scenario also works. What caught me off-guard was that the [docs/pom.xml](https://github.com/wildfly-swarm/wildfly-swarm/blob/master/docs/pom.xml#L15) which pulls in `reference` & `howto-parent` is based off `build-parent`. I figured there was some refactoring required / pending in this space.

So, then I reasoned that if we are anyway building all of the modules (including docs) as part of the current contribution checklist (i.e., executing `mvn clean install` ), then shouldn't we make it simpler for the contributors to test their changes across all modules within the existing project? Hence the set of changes where the parent of `howto-parent` is set to `root`.

Do you still think I should close this PR? I am ok either ways. Just want to make sure that contributors can ensure that they are not breaking anything by, for example, changing the version and building locally to emulate a clean slate.

kenfinnigan:
I think at this point I'm not sure whether the benefits are worth the changes, and could likely make our productization process more complicated.

heiko-braun:
Please hold this for now. I need to dive into the changes to see what impact it might have on prod

roguexz:
Acknowledged. I just wanted to fix the Swagger related issue that I was facing .. then OCD kicked in 😄 .. one PR led to another and here we are! Will wait to hear your feedback.

roguexz:
@heiko-braun When you get some time, please have a look at this PR. I have been using this commit (quite frequently) for generating custom builds and verifying my local dev environment. If we do not wish to accept this commit, then we should probably have some `developer documentation` which talks about how one can generate a local build for testing.

heiko-braun:
@roguexz While I can see value in your changes, I have to agree with @kenfinnigan .  We have a parallel source tree used for productisation and that work is ongoing at the moment. Significant changes to the build structure, despite their benefit, make it harder for us to synchronise these two.  

Under that aspect I would not like to merge this. Let's wait for things to settle until we are in a better position to apply these changes to both code trees.

roguexz:
Sure thing @heiko-braun. I shall keep the PR alive as it helps with my testing. Let us look at it later.

roguexz:
Closing this PR. Will open a separate one with more targeted changes that don't raise the conflict identified in this PR's discussion.

PR #841, opened 2018-03-20
opentracing needs to be in product profile

heiko-braun:


kenfinnigan:
What's the requirement for cli, hibernate-search and infinispan to be included?

heiko-braun:
@kenfinnigan  It's derived from your fraction landscape spreadsheet. The are marked as supported.

kenfinnigan:
Ah.

I don't know how much John has reviewed that landscape in the past.

If we don't think those three are necessary, we don't have to productize them this release

Ladicek:
I think OpenTracing is a product requirement, no? I remember we debated including the Jaeger fraction too (I still think it's probably a good idea).

From the rest, I don't remember any discussion or requirement -- except maybe Infinispan, which might be required for the Caching mission?

heiko-braun:
@Ladicek yes, we keep OT and remove the others for now

kenfinnigan:
Ooh, hadn't thought of that.

@michalszynkiewicz, does Cache mission use Infinispan fraction?

kenfinnigan:
Per IRC discussion, Cache mission uses Hot Rod client not Infinispan fraction

Ladicek:
Ah OK, thanks!

kenfinnigan (approved):

PR #842, opened 2018-03-20
SWARM-1904: Gradle plugin does not package archiveTask

johnaohara:
I have changed the gradle plugin for derive the artifact name, group and version from the jarTask that is being wrapped, and not from the project name. This allows the archiveTask configuration option to function correctly


kenfinnigan:
ok to test

johnaohara:
Looks like the CI job failed as there was not enough space left on the disk?

kenfinnigan (approved):

PR #844, opened 2018-03-21
SWARM-1667: Swarm swagger module does not generate swagger.json

roguexz:
Motivation
----------
1. The spec states that if there is a custom application subclass that returns non-empty collections for either getClasses() or getSingletons() then the framework should only include those entities and nothing else.
2. The way the Swagger framework has been *leveraged* is by including its resources (ApiListingResource, SwaggerSerializers) to the list of classes being returned (either explicitly or implicitly).
3. The Swagger fraction adds these resources to the list of classes available to the archive so that an implicit scan will load them.

4. Incidentally, JAX-RS allows for CDI within the JAX-RS resources / application-sub-classes.
5. If CDI fraction has been enabled, then the framework introduces the application-sub-class as a *bean*.

If it is a bean .. then we can add an interceptor ... by plugging in as an extension and observing ProcessAnnotatedType events. We can potentially update the list of classes being returned by this bean (if either of the base target methods return a non-empty collection). And voila .. Swagger definitions start working!

Modifications
-------------
1. Implemented a CDI extension within the Swagger fraction.
2. Defined an InterceptorBinding & Interceptor.
3. The extension scans for classes which are sub-types of javax.ws.rs.core.Application
4. These AnnotatedTypes (associated with types identified above) are updated to include a custom InterceptorBinding.
5. The associated Interceptor checks if the application-sub-class returns resources / singletons explicitly or not. If it does, then the interceptor ensures that the Swagger resources are included.

Result
------
Updated the test cases to verify different scenarios. The build executes properly.

-----

### DIFF
--- a/fractions/swagger/src/main/java/org/wildfly/swarm/swagger/runtime/SwaggerArchivePreparer.java
+++ b/fractions/swagger/src/main/java/org/wildfly/swarm/swagger/runtime/SwaggerArchivePreparer.java
@@ -4,6 +4,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
+import javax.enterprise.inject.spi.Extension;
 import javax.inject.Inject;
 
 import org.jboss.shrinkwrap.api.Archive;
@@ -80,6 +81,10 @@ public class SwaggerArchivePreparer implements DeploymentProcessor {
 
             // Make the deployment a swagger archive
             SwaggerArchive swaggerArchive = deployment.as(SwaggerArchive.class);
+            // SWARM-1667: Add the custom CDI extension to the deployment to provide a workaround solution.
+            deployment.addModule("org.wildfly.swarm.swagger", "deployment");
+            deployment.addAsServiceProvider(Extension.class.getName(), "org.wildfly.swarm.swagger.deployment.SwaggerExtension");
+
 
             if (this.title != null) {
                 swaggerArchive.setTitle(this.title);
--- a/testsuite/testsuite-swagger-customized/src/test/java/com/mycorp/SwaggerConfiguredArquillianWebAppTest.java
+++ b/testsuite/testsuite-swagger-customized/src/test/java/com/mycorp/SwaggerConfiguredArquillianWebAppTest.java
@@ -1,22 +1,22 @@
-/**
- * Copyright 2015-2016 Red Hat, Inc, and individual contributors.
+/*
+ * Copyright 2015-2018 Red Hat, Inc, and individual contributors.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ *  http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
  */
+
 package com.mycorp;
 
 import java.io.BufferedReader;
-import java.io.File;
 import java.io.InputStreamReader;
 import java.net.URL;
 import java.net.URLConnection;
@@ -26,6 +26,7 @@ import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.wildfly.swarm.undertow.WARArchive;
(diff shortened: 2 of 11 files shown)


arquillian/arquillian/src/main/java/org/wildfly/swarm/arquillian/util/IOUtils.java
```diff
@@ -0,0 +1,60 @@
+/*
```
roguexz: Is this the right copyright to use? I found different versions in the code.

testsuite/testsuite-buildtool/multi-module/view/src/test/java/VerifyBuildArtefactIT.java
```diff
@@ -29,32 +26,9 @@ public static Archive createDeployment() throws Exception {
 
     @org.junit.Test
     public void testEndppoint() throws Exception {
-        String urlContents = getUrlContents("http://127.0.0.1:8080/");
+        String urlContents = IOUtils.getUrlContents("http://127.0.0.1:8080/");
         Assert.assertEquals("something\n", urlContents);
     }
 
-    private static String getUrlContents(String theUrl) {
```
roguexz: This method body was duplicated in multiple test classes. Moved the code in to a separate utility file under the Arquillian module and replace all occurrence that were identified by the IDE>

testsuite/testsuite-swagger-customized/src/test/java/com/mycorp/SwaggerConfiguredArquillianWebAppTest.java
```diff
@@ -0,0 +1,57 @@
...
+import static org.fest.assertions.Assertions.assertThat;
+
+/**
+ * Test case for verifying that the swagger resources are associated properly with WebArchive deployments.
+ */
+@RunWith(Arquillian.class)
+public class SwaggerConfiguredArquillianWebAppTest {
+
+    @Deployment
+    public static Archive createDeployment() throws Exception {
```
roguexz, line 43: This test case simulates the custom Application-sub-class scenario.

fractions/swagger/src/main/java/org/wildfly/swarm/swagger/runtime/SwaggerArchivePreparer.java
```diff
@@ -73,10 +89,25 @@ public void process() {
...
+            if (!Objects.equals(contextPath.get(), path)) {
+                contextPath.set(path);
+            }
+        }
+
         if (archive.getName().endsWith(".war")) {
             // Create a JAX-RS deployment archive
             WARArchive deployment = archive.as(WARArchive.class);
             deployment.addModule("io.swagger");
+            deployment.addAsServiceProvider(Extension.class, SwaggerExtension.class);
```
kenfinnigan: You don't need this line and the META-INF/services file for the Extension. One or the other will suffice.

Does this actually work? I think `SwaggerExtension` needs to be in `.deployment` package so it's available within user deployment.

kenfinnigan:
So does this change mean the Swagger fraction doesn't work without CDI present?

fractions/swagger/src/main/java/org/wildfly/swarm/swagger/runtime/SwaggerArchivePreparer.java
roguexz: Ah! That might have been the reason it did not get picked up. I had to add it explicitly for it to work. I will run another test after the 1471 PR has been approved.

roguexz:
The Swagger faction itself will work with or without CDI.

**Scenario:** Developer has defined a custom application class that returns specific types / singletons.
Because of the specification, we are unable to add the Swagger resources (unless we find a way to plug into the RESTEasy framework to make them default). In this scenario, if the developer wishes to see the Swagger resources, then we are going to ask them to include the CDI fraction as well. Otherwise, they will not see the Swagger definitions (unless they add it manually).

roguexz:
Splitting up this PR in to smaller pieces. Please hold on before merging this.

fractions/swagger/src/main/java/org/wildfly/swarm/swagger/runtime/SwaggerArchivePreparer.java
roguexz: @kenfinnigan How does the `.deployment` package work? Is it added **automatically** to the archive? Do I need to add it as a module (with slot=deployment)?

I am trying to modify this PR and running in to issue. The [create-a-fraction](https://github.com/wildfly-swarm/wildfly-swarm/blob/master/docs/howto/create-a-fraction/index.adoc) doc does not provide enough clarity on this behavior. What I did learn is that classes defined in `.deployment` are not available to `.runtime`. Are there any samples that I can see to get some inspiration?

bobmcwhirter: It is not added automatically, but a jboss-module module with the slot of `:deployment` is created (and is excluded from the normal API or runtime class-space). A fraction can then do a `@DeploymentModule` to add the `:deployment` module to a deployment. Not necessarily very clear or straight-forward, but that's how it works. :)

roguexz: @bobmcwhirter Thank you for the explanation. It makes a lot more sense now. Is there a way to define an module as an optional dependency in the `module.conf` file? This solve relies on the availability of CDI. But having CDI is not mandatory. So, can I define an optional dependency on `org.wildfly.swarm.cdi` in the `module.conf` file so that I don't run in to linkage issues?

roguexz: Or rather .. put another way. If I define a dependency in `module.conf`, does that mean that dependency is exported to the archive? Or are they unrelated?

bobmcwhirter: Yep, just prefix the line with a literal `*`.

e.g.

https://github.com/wildfly-swarm/wildfly-swarm/blob/master/fractions/jolokia/module.conf#L4

Jolokia optionally has a dependency on keycloak, if it's around.

bobmcwhirter: No, it will *not* be exported. It'll just be available to things within the `:runtime` and such modules so they themselves can link against it.

If you want a module exported (or imported? depending on how you look at it) into the deployment, then once against just `@DeploymentModule` it, which will add it to the deployment. But alas `@DeploymentModule` doesn't support optionals.

You *could* add a deployment processor, which should be able to `addModule` on the deployment archive, which ultimately allows full control over adding modules, as it's just scribbling the `jboss-deployment-structure.xml`, which is what `@DeploymentModule` does also, ultimately.

https://github.com/wildfly-swarm/wildfly-swarm/blob/master/core/spi/src/main/java/org/wildfly/swarm/spi/api/JBossDeploymentStructureContainer.java#L39

https://github.com/wildfly-swarm/wildfly-swarm/blob/master/core/spi/src/main/java/org/wildfly/swarm/spi/api/Module.java

roguexz:
@bobmcwhirter @kenfinnigan Thank you for your inputs. I have restructured the code, split it in to 3 separate requests that are easy to review and verified that the build is working fine locally.  Also put in a test case for verifying the same.

Please review and provide me your comments.

kenfinnigan:
So is this PR still valid or should it be closed?

ie. are the changes now in different PRs?

roguexz:
@kenfinnigan This is a valid PR. I split my changes associated with the following two items in to separate PRs.

1. Refactoring duplicate code in to a common utility method ( #849 ).
2. Aligning the context-root of the Swagger bean with the actual JAX-RS context root ( #850 )

The text for #850 needs to change both in the commit as well as in the bug. Once you guys merge any of the pending PRs, I will rebase the other PRs and make sure that are no conflicts.

kenfinnigan:
There now appears to be a conflict, could you rebase and resolve it? Thanks

roguexz:
Ack. Working on it.

roguexz:
Rebased and verified the local build. Ready for review.

roguexz:
Retest this please

roguexz:
Finally! Successful builds - ✅ 
Please review / merge.

kenfinnigan (approved):

PR #845, opened 2018-03-21
SWARM-1906: JAX-RS Fraction does not introspect jar files inside war files when searching for the presence of ApplicationPath

roguexz:
Motivation
----------
I noticed that there were multiple JAX-RS applications deployed for my WAR file. As I debugged further, I realized that the issue was because the JAR file (inside the WAR file) was being viewed as a ByteArrayAsset and was being skipped over. With the recent fixes to IndexView representation, the fix was actually quite simple.

Modifications
-------------
1. During the deployment process, query the IndexView to see if there are **any** classes which contain the @ApplicationPath annotation.
2. If present, do not generate the default JAX-RS Application class.

Result
------
1. Verified a locally patched build and ensured that things are working as desired.
2. Ensured that the build has no errors.

-----

### DIFF
--- a/fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/runtime/DefaultApplicationDeploymentProcessor.java
+++ b/fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/runtime/DefaultApplicationDeploymentProcessor.java
@@ -20,8 +20,11 @@ import java.io.InputStream;
 import java.util.Map;
 
 import javax.inject.Inject;
+import javax.ws.rs.ApplicationPath;
 import javax.ws.rs.core.Application;
 
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.IndexView;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.ArchivePath;
 import org.jboss.shrinkwrap.api.ArchivePaths;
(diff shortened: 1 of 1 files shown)


heiko-braun:
add to whitelist

kenfinnigan:
As mentioned on the issue itself, this isn't the fix we want to implement.

What we want to do is defined in SWARM-1471, which is to make the default Application generation optional. With that change this one is not necessary

roguexz:
@kenfinnigan I see that you are recommending the reorganization of the logic for creating the default JAX-RS application class. I can definitely close this PR in favor of the other solution. The concern that I see with SWARM-1471 is that we may end up breaking away from the spec.

If there are no definitions, then the framework should automatically create a default Application instance.

So I am not sure if that would be the right way to achieve it. What do you think?

kenfinnigan:
What we did with WF Swarm by generating that class actually goes against the Java EE spec and caused much confusion.

That's why we need to simplify and only generate the Application class when the developer sets a property to request it

roguexz:
@kenfinnigan Ah! Understood! So basically, WF Swarm provides the ability to declaratively define the JAX-RS application root. That is neat!

I shall close this PR and raise a separate one. This behavior has confused me as well 😄.

roguexz:
@kenfinnigan Quick question ... The fix for 1471 is a breaking change. Do you still want to proceed with it? Or is it used rarely that the impact would be rather low?

kenfinnigan:
To fix it we're going to have to break the behavior at some point

roguexz:
Closing this PR. Will raise one that fixes the issue identified in SWARM-1471.

PR #847, opened 2018-03-22
SWARM-1471: Do not auto-generate the JAX-RS application if the application-path has not been defined in YAML.

roguexz:
Motivation
----------
Currently WF Swarm auto-generates a JAX-RS application class which leads to an inconsistent behavior (does not comply with the JAX-RS spec). Developers tend to get confused with the presence of multiple Application references being printed in the console. The intention of this auto-generation was to help developers easily define the JAX-RS application-path declaratively without needing to define a superfluous empty class. As discussed in the Jira issue, we should be generating the class only if the declarative parameter is defined.

Modifications
-------------
1. Add the Application subclass only if no sub-class of Application is present **and** the YAML file defines the "application-path" entry.
2. Updated the deployment definitions for various test cases that were relying on this implicit behavior.

Result
------
Verified the behavior (across many, many test cases).

-----


fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/runtime/DefaultApplicationDeploymentProcessor.java
```diff
@@ -76,7 +76,9 @@ public void process() throws Exception {
             return;
         }
 
-        addGeneratedApplication(archive.as(JAXRSArchive.class));
+        if (applicationPath.isExplicit()) {
+            addGeneratedApplication(archive.as(JAXRSArchive.class));
+        }
     }
```
roguexz, line 83: This is the only change that was required. All the other changes pertain to the test cases.

bobmcwhirter:
Haven’t reviewed but I think this is a good change. We just need to manage
expectations since it would be breaking.

roguexz:
In addition to the release notes, the JAX-RS framework would automatically print a message that there are no Application subclasses or Servlet mappings present.

It might be a manageable scenario.

fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/runtime/DefaultApplicationDeploymentProcessor.java
```diff
@@ -76,7 +76,9 @@ public void process() throws Exception {
             return;
         }
 
-        addGeneratedApplication(archive.as(JAXRSArchive.class));
+        if (applicationPath.isExplicit()) {
```
Ladicek, line 80: Should probably change a bit the `@AttributeDocumentation` of `applicationPath`, to reflect the change.

Ladicek:
I myself would create the `Application` classes inside the tests, instead of adding `project-defaults.yml`, but I don't really mind :-) I mean, I don't really mind in the tests, but maybe in the howtos, we should really prefer creating the `Application` class, as that's the common JAX-RS way. WDYT?

kenfinnigan:
I agree it's fine to use the yml version for tests, but the howto's should create the `Application` class directly

docs/howto/create-a-hollow-jar/src/main/resources/project-defaults.yaml
```diff
@@ -0,0 +1,5 @@
+swarm:
```
kenfinnigan: Should this contain the port setting from the removed project-defaults.yml?

roguexz: Not really. The file was (for some reason) not being picked up. The test case was still pointing to 8080.

fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/runtime/DefaultApplicationDeploymentProcessor.java
roguexz: Let me take a stab at providing some useful documentation for the annotation. Will need a customer focussed doc-review of that line though.

roguexz (commented):
I shall create Application classes in the how-to projects as well and remove the `application-path` entries (and the whole file if applicable).

fractions/javaee/jaxrs/src/main/java/org/wildfly/swarm/jaxrs/runtime/DefaultApplicationDeploymentProcessor.java
Ladicek: It's a single sentence, not an overly big deal. My suggestion: `If set, WildFly Swarm will automatically generate a JAX-RS Application class and use this value as the @ApplicationPath`

roguexz: Since I haven't looked at the existing documentation that much, I wasn't aware of the tone / technical details that are included in the documentation. Will use the above sentence and update.

roguexz:
Incorporate the changes. Executed a build from the `howto` folder, which went through fine.
Can you please review and provide your comments?

kenfinnigan (approved):

roguexz:
A little less than 3 hours for completion of the builds. I wonder if we can optimize that! Anyways, ready for merging.

roguexz:
@kenfinnigan @bobmcwhirter Is this PR good to be merged? Do you have any additional comments about it? Or should we put it on hold?

PR #849, opened 2018-03-23
Code cleanup. Refactor duplicate code in to a utility.

roguexz:
Motivation
----------
While working on a solution for SWARM-1667, I noticed (thanks to the IDE) that there was code duplication within the test cases which were trying to read the contents of the deployed archive.

Modifications
-------------
* Moved all such occurrences to the Arquillian module and consumed the utility method in existing test cases.

Result
------
Verified that the build is going through properly.

-----

### DIFF
--- a/testsuite/testsuite-buildtool/multi-module/pom.xml
+++ b/testsuite/testsuite-buildtool/multi-module/pom.xml
@@ -72,13 +72,6 @@
       <groupId>org.wildfly.swarm</groupId>
       <artifactId>jaxrs</artifactId>
     </dependency>
-
-    <dependency>
-      <groupId>org.wildfly.swarm</groupId>
-      <artifactId>arquillian</artifactId>
-      <scope>test</scope>
-    </dependency>
-
   </dependencies>
 
 
--- a/testsuite/testsuite-buildtool/multi-module/view/pom.xml
+++ b/testsuite/testsuite-buildtool/multi-module/view/pom.xml
@@ -51,7 +51,11 @@
       <groupId>org.wildfly.swarm.testsuite</groupId>
       <artifactId>multi-module-api</artifactId>
     </dependency>
-
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>2.4</version>
+    </dependency>
   </dependencies>
 
 
--- a/testsuite/testsuite-buildtool/multi-module/view/src/test/java/VerifyBuildArtefactIT.java
+++ b/testsuite/testsuite-buildtool/multi-module/view/src/test/java/VerifyBuildArtefactIT.java
@@ -1,9 +1,8 @@
-import java.io.BufferedReader;
 import java.io.File;
-import java.io.InputStreamReader;
 import java.net.URL;
-import java.net.URLConnection;
+import java.nio.charset.Charset;
 
+import org.apache.commons.io.IOUtils;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.junit.Arquillian;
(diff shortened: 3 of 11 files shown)


Ladicek:
I'm not sure if WildFly Swarm is the right project to carry this kind of utilities (but I'm not the one to decide either!). The tests should just use `org.apache.httpcomponents:fluent-hc`, where doing this is as simple as `String response = Request.Get("http://localhost:8080/").execute().returnContent().asString();`.

roguexz:
Agreed. It looks like there was a lot of copy-paste within a few of the test suites. So, instead of adding a new module dependency (not sure if it would be new), I am just moving the code around so that it can be replaced at a later point.

kenfinnigan:
I would tend to agree, if there's a way to do this without adding utility classes that would be my preference.

roguexz:
Sure thing. Let me see if I can find something in the existing dependency tree.

roguexz:
Looks like Commons IOUtils is available. I should be able to replace all calls with the [toString](https://commons.apache.org/proper/commons-io/javadocs/api-2.4/org/apache/commons/io/IOUtils.html#toString(java.net.URL,%20java.nio.charset.Charset)) method invocation. Will rebase and update.

roguexz:
Done.

roguexz:
Retest this please

kenfinnigan:
This needs rebasing

roguexz:
Rebased. Verified local build.

roguexz (commented):
@kenfinnigan Would you like me to make any further changes before merging this?

roguexz:
There problem was that though the version is being manged, it was not being
made available inside the failsafe deployment. I noticed there were few
other instances in the test suite projects that seemed to be pulling it
explicitly.

Would you happen to know anything about why the Arquillian deployment might
not have the dependency when running inside the failsafe plugin?

kenfinnigan:
@Ladicek probably knows for sure, but I think it's because the project is executed as a Maven Plugin test

roguexz:
@Ladicek 
This is the error that I get if I don't include the dependency explicitly,

    Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 8.934 sec <<< FAILURE! - in VerifyBuildArtefactIT
    testEndppoint(VerifyBuildArtefactIT)  Time elapsed: 0.215 sec  <<< ERROR!
    java.lang.NoClassDefFoundError: org/apache/commons/io/IOUtils
        at VerifyBuildArtefactIT.testEndppoint(VerifyBuildArtefactIT.java:31)
    Caused by: java.lang.ClassNotFoundException: org.apache.commons.io.IOUtils from [Module "deployment.endpoint.war" from Service Module Loader]
        at VerifyBuildArtefactIT.testEndppoint(VerifyBuildArtefactIT.java:31)

Can you confirm the changes that I made (for including the commons-io jar directly in to the POM) are the correct ones?

kenfinnigan (approved):

Ladicek:
I'm not quite sure what's the actual question, but given that the PR was merged and it doesn't touch any of the Maven Plugin test suite files, I hope I don't really have to answer :-)

roguexz:
@Ladicek I think the question still exists. If you look at the [pom.xml file in multi-module view](https://github.com/wildfly-swarm/wildfly-swarm/blob/master/testsuite/testsuite-buildtool/multi-module/view/pom.xml#L55). If you delete the `commons-io` dependency, the code does compile, but the test case fails. Not entirely sure why some of the dependencies are not being bundled. If you do have some insights, we can potentially fix a hidden issue, or have a point about documentation.

Ladicek:
Ah! Sorry, I have no idea about `testsuite-buildtool`.

PR #851, opened 2018-03-26
SWARM-1891: OpenApiHttpHandler - support graceful startup.

mkouba:
Motivation
----------
OpenApiHttpHandler currently attempts to load the OpenApi model in
static init block. This results in ExceptionInInitializerError if
Undertow instantiates the handler before the app is deployed completely.

Modifications
-------------
1. Skip execution until the model is available.
2. Cache serialized OpenApi models lazily.

Result
------
Graceful startup is supported, no ExceptionInInitializerError is thrown.

### DIFF
--- a/fractions/microprofile/microprofile-openapi/src/main/java/org/wildfly/swarm/microprofile/openapi/runtime/OpenApiHttpHandler.java
+++ b/fractions/microprofile/microprofile-openapi/src/main/java/org/wildfly/swarm/microprofile/openapi/runtime/OpenApiHttpHandler.java
@@ -18,8 +18,9 @@ package org.wildfly.swarm.microprofile.openapi.runtime;
 
 import java.io.IOException;
 import java.util.Deque;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
-import org.eclipse.microprofile.openapi.models.OpenAPI;
 import org.wildfly.swarm.microprofile.openapi.api.OpenApiDocument;
 import org.wildfly.swarm.microprofile.openapi.runtime.io.OpenApiSerializer;
 import org.wildfly.swarm.microprofile.openapi.runtime.io.OpenApiSerializer.Format;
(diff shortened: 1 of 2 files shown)


mkouba:
retest this please

kenfinnigan:
retest this please

fractions/microprofile/microprofile-openapi/src/main/java/org/wildfly/swarm/microprofile/openapi/runtime/OpenApiHttpHandler.java
```diff
@@ -39,31 +40,21 @@
...
+        this.cachedModels = new ConcurrentHashMap<>();
     }
 
     /**
      * @see io.undertow.server.HttpHandler#handleRequest(io.undertow.server.HttpServerExchange)
      */
     @Override
     public void handleRequest(HttpServerExchange exchange) throws Exception {
-        if (OAI.equalsIgnoreCase(exchange.getRequestPath())) {
+        if (OAI.equalsIgnoreCase(exchange.getRequestPath()) && OpenApiDocument.INSTANCE.isSet()) {
```
kenfinnigan, line 57: Is there a situation that could occur where the model wouldn't be set on `OpenApiDocument` and this is called?

mkouba: Yes. This issue is all about graceful startup, i.e. processing http requests *before* the app is deployed.

mkouba:
Windows machine fails again with:
>java.lang.AssertionError: <''> should be equal to :<'Management Interface'> ignoring case
	at org.wildfly.swarm.management.console.ManagementConsoleTest.testHomePageShouldHaveCorrectTitle(ManagementConsoleTest.java:47)

fractions/microprofile/microprofile-openapi/src/main/java/org/wildfly/swarm/microprofile/openapi/runtime/OpenApiHttpHandler.java
kenfinnigan: Is it about processing http requests, or just that Undertow initialized the class?

I would expect that Undertow initializing the class, and thus the static method being called, doesn't mean it's handling requests, merely that Undertow is being setup ready to take requests.

That's why I wasn't understanding at the point we're accepting a request, that the model wouldn't be set on `OpenApiDocument`?

mkouba: > Is it about processing http requests, or just that Undertow initialized the class?

I have to find out.

mkouba: @kenfinnigan ~~You were right, it seems the handler is not used until the app is deployed. I will remove the needless condition~~. Actually no, if I try to GET `/openapi`, e.g. `watch -n1 curl http://localhost:8080/openapi` the handler is used once Undertow is up and **even before** the app is deployed. So I think the additional condition is correct.

kenfinnigan: Ok, in that case, should a message be logged that it's not being processed for that reason?

Otherwise it just passes through and does nothing when it was expected it would

mkouba: The client receives HTTP status 404 (until the app is deployed) which is imho OK.

kenfinnigan (approved):

PR #856, opened 2018-04-02
[SWARM-1924] Creating a keycloak fraction test

sberyozkin:


testsuite/testsuite-keycloak/src/test/java/org/wildfly/swarm/keycloak/KeycloakArquillianTest.java
```diff
@@ -15,40 +15,122 @@
...
+        JAXRSArchive deployment = ShrinkWrap.create(JAXRSArchive.class);
+        deployment.addResource(SecuredApplication.class);
+        deployment.addResource(SecuredResource.class);
+        deployment.addAsResource("wildfly-swarm-keycloak-example-realm.json");
+        deployment.addAsResource("keycloak.json");
         return deployment;
     }
 
     @CreateSwarm
     public static Swarm newContainer() throws Exception {
```
kenfinnigan, line 58: Is there a way to do this without `@CreateSwarm`?

testsuite/testsuite-keycloak/src/main/java/org/wildfly/swarm/keycloak/SecuredApplication.java
```diff
@@ -0,0 +1,11 @@
+package org.wildfly.swarm.keycloak;
+
+import javax.ws.rs.ApplicationPath;
+import javax.ws.rs.core.Application;
+
+@ApplicationPath("/")
+public class SecuredApplication extends Application {
+
+    public SecuredApplication() {
+    }
```
Ladicek: Ctor unnecessary :-)

testsuite/testsuite-keycloak/src/test/java/org/wildfly/swarm/keycloak/KeycloakArquillianTest.java
```diff
@@ -15,40 +15,122 @@
...
+
     }
 
+    private String getAccessTokenFromResponse(String response) {
+        String tokenStart = response.substring("{\"access_token\":\"".length());
+        return tokenStart.substring(0, tokenStart.indexOf("\""));
+    }
+
+    private static String getWarName() throws Exception {
+        // Arquillian creates a war using a "wfswarm" + UUID + ".war" format.
```
Ladicek: I'm not sure if it would work with Swarm Arquillian adapter, but you can provide an archive name in the `ShrinkWrap.create()` method call.

sberyozkin: No, that does not work, it still uses UUID, I tried it few times earlier

testsuite/testsuite-keycloak/src/test/java/org/wildfly/swarm/keycloak/KeycloakArquillianTest.java
sberyozkin: I don't know if it is possible to avoid creating Swarm explicitly :-) given that it has to be aware of the custom configuration resource and a  '-s' command line argument as opposed to a system property is used to let Swarm know about its location

testsuite/testsuite-keycloak/src/main/java/org/wildfly/swarm/keycloak/SecuredApplication.java
sberyozkin: Done :-)

testsuite/testsuite-keycloak/src/test/java/org/wildfly/swarm/keycloak/KeycloakArquillianTest.java
kenfinnigan: Adding a `project-defaults.yml` inside the Archive should mean it's picked up, so it should be possible to add into `createDeployment()`

sberyozkin: But the war name is dynamic, and as far as I understand the war name is not known yet at the deployment creation time

kenfinnigan: It doesn't work if you call `Shrinkwrap.create(JAXRSArchive.class, "myName.war")`?

sberyozkin: I can definitely use Secured instead of generating a custom project config ? It is much simpler (though I guess testing the config approach is preferred)

sberyozkin: Let me double check again :-), I don't think it works...

testsuite/testsuite-keycloak/src/test/java/org/wildfly/swarm/keycloak/KeycloakArquillianTest.java
Ladicek: Ah OK, sorry. We should probably fix it :-D

testsuite/testsuite-keycloak/src/test/java/org/wildfly/swarm/keycloak/KeycloakArquillianTest.java
sberyozkin: No, does not work, I spent quite a lot of time with it last week, FYI, after specifying 'example-keycloak' as a war name, I get

"target/wfswarmexample-keycloak233151353883937125.war/WEB-INF/"

kenfinnigan: ok, ignore me then

testsuite/testsuite-keycloak/src/test/java/org/wildfly/swarm/keycloak/KeycloakArquillianTest.java
sberyozkin: Hi Ladislav, np at all. Interesting, is it in our control then :-) ? Yes, see some Arquillian related code...I can definitely open an issue to track it ?

sberyozkin:
Windows failure is well captured at
https://bugs.openjdk.java.net/browse/JDK-8197918
Let me try with URI first before trying to strip off the leading slash if it is Windows

kenfinnigan:
restest this please

sberyozkin:
Looks like stripping the forward slash worked on Win :-).
I have a pending commit updating SecuredArchiveArchive but I'll postpone it now till the next dev version gets opened; this test will be simplified then then as well...

Ladicek (approved):
LGTM with one question.

testsuite/testsuite-keycloak/src/test/java/org/wildfly/swarm/keycloak/KeycloakArquillianTest.java
```diff
@@ -15,40 +15,127 @@
...
+    private static String stripLeadingSlashIfWin(String path) {
+        return System.getProperty("os.name").toLowerCase().indexOf("win") >= 0 && path.startsWith("/") 
+            ? path.substring(1) : path;
     }
 
     @Test
-    @RunAsClient
-    public void testNothing() throws InterruptedException {
-        // TODO figure out how to confirm keycloak is working.
+    public void testResourceIsSecured() throws Exception {
```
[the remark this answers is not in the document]
sberyozkin: In this particular case I dropped @RunAsClient as I did not have any opinion about it but technically the client which needs to get something out if KC can be run standalone or in the container scope. I agree though it is cleaner to run it independently...I'm going to update the PR for SWARM-1921 which builds on top of this test. Let me know please if you'd like this PR be updated instead...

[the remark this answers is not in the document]
Ladicek: I think we can update the test to `@RunAsClient` in the future PR, sure.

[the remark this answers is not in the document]
sberyozkin: OK, thanks, I've confirmed it works independently as well, will update the follow up PR asap

PR #857, opened 2018-04-02
[SWARM-1769] Adding a basic keycloak-server test

sberyozkin:


testsuite/testsuite-keycloak-server/src/test/java/org/wildfly/swarm/keycloak/server/KeycloakServerArquillianTest.java
```diff
@@ -32,16 +37,21 @@
...
 
     @Deployment(testable = false)
-    public static Archive createDeployment() {
+    public static Archive<?> createDeployment() {
         JARArchive deployment = ShrinkWrap.create(JARArchive.class);
         deployment.add(EmptyAsset.INSTANCE, "nothing");
         return deployment;
     }
 
+    @CreateSwarm
```
kenfinnigan: This particular method shouldn't be necessary, as that should be the default behavior of the container

sberyozkin: That is done... The actual endpoint address is used to access the welcome page, so getting 200 on that should be sufficient to confirm the server is up; pinging other endpoints require a more involved set up (which is what the keycloak test now does)

sberyozkin:
Linux one seems to be hanging here...

sberyozkin:
retest this please

kenfinnigan:
retest this please

Ladicek (approved):

PR #862, opened 2018-04-04
SWARM-1929: MP FT - circuit breaker never opens if there is a fallback.

mkouba:
Motivation
----------
There is a bug in MP FT implementation.

Modifications
-------------
Fix HystrixCommandInterceptor and DefaultCommand to support this basic
scenario. Also add basic support for CircuitBreaker.failOn() which was
so far completely ignored. We should align the functionality once the
spec issue is resolved.

Result
------
The bug should be fixed now.

### DIFF
--- a/fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/DefaultCommand.java
+++ b/fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/DefaultCommand.java
@@ -16,7 +16,11 @@
 
 package org.wildfly.swarm.microprofile.faulttolerance.deployment;
 
+import static org.wildfly.swarm.microprofile.faulttolerance.deployment.config.CircuitBreakerConfig.FAIL_ON;
+
 import java.util.function.Supplier;
+
+import org.wildfly.swarm.microprofile.faulttolerance.deployment.config.FaultToleranceOperation;
 
 import com.netflix.hystrix.HystrixCommand;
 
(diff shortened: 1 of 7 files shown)


fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/HystrixCommandInterceptor.java
```diff
@@ -137,9 +137,22 @@ public Object interceptCommand(InvocationContext ic) throws Exception {
                     res = command.execute();
                 }
                 if (syncCircuitBreaker != null) {
-                    syncCircuitBreaker.executionSucceeded();
+                    // HystrixRuntimeException is only thrown if command fails and does not have a fallback
+                    if (command.hasFailure()) {
```
mkouba: Hm, I still need to check this one...

fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/DefaultCommand.java
```diff
@@ -30,26 +34,56 @@
...
+        this.failure = null;
+        this.operation = operation;
     }
 
     @Override
     protected Object run() throws Exception {
-        return ctx.proceed();
+        try {
+            return ctx.proceed();
+        } catch (Exception e) {
```
Ladicek: `failOn` defaults to `Throwable`, so maybe you need to catch that here?

mkouba: Good point, although I don't expect people using `Error`s here ;-)

kenfinnigan:
retest this please

antoinesd (approved):
LGTM

antoinesd:
FT TCK doesn’t have test for this. Will add one.

Ladicek:
Thanks @mkouba!

PR #864, opened 2018-04-05
[SWARM-1921] Making KC SecurityContext availability not depending on the use of Secured

sberyozkin:


fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/KeycloakFraction.java
```diff
@@ -27,7 +28,10 @@
  */
 @WildFlyExtension(module = "org.keycloak.keycloak-adapter-subsystem")
 @WildFlySubsystem("keycloak")
-@DeploymentModule(name = "org.keycloak.keycloak-core")
+@DeploymentModules({
+  @DeploymentModule(name = "org.keycloak.keycloak-core"),
+  @DeploymentModule(name = "org.wildfly.swarm.keycloak", slot = "deployment", export = true, metaInf = DeploymentModule.MetaInfDisposition.IMPORT)
+})
```
Ladicek: Isn't the `@DeploymentModule` annotation `@Repeatable`?

testsuite/testsuite-keycloak/src/main/java/org/wildfly/swarm/keycloak/SecuredResource.java
```diff
@@ -4,12 +4,16 @@
...
+
 @Path("/secured")
 public class SecuredResource {
 
     @GET
     @Produces("text/plain")
     public String get() {
-        return "This resource is secured";
+        // This can also be done in the ContainerResponseFilter
+        return "Hi " + KeycloakSecurityContextAssociation.get().getToken().getPreferredUsername()
```
Ladicek, line 16: Is there another way to get the login info? I think `HttpServletRequest.getRemoteUser()` should work here, no?

sberyozkin: Hi Ladislav, I'm not sure; in this case the remote user is technically a client which invokes on behalf of user1 unless KC impersonates the client as a user1 (but given that it a non-code flow KC SecurityContext's IdToken is also null, which would otherwise point to a live user). 
The test simply asserts that the KC SC is available in the context, but I've been planning to enhance a bit the examples/security/keycloak later on, and may be use a filter to provide a custom JAX-RS SecurityContext and get the service code use the injected SecurityContext to get the info.

fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/KeycloakFraction.java
sberyozkin: Thanks, did not know that, updating...

testsuite/testsuite-keycloak/src/main/java/org/wildfly/swarm/keycloak/SecuredResource.java
sberyozkin: Just tried out of interest, both HttpServletRequest.getRemoteUser and (JAXRS)SecurityContext.getUserPrincipal().getName() return UUID, which matches the access token's subject property. I'm not sure if that points to the actual 'user1' or not in the access token (as opposed to the id token), can try to check with the demo later on by starting a standalone KC. And with the demo we can get the actual 'user1' returned as a principal name :-)

sberyozkin:
retest this please

testsuite/testsuite-keycloak/src/main/java/org/wildfly/swarm/keycloak/SecuredResource.java
Ladicek: Ah OK. My only concern here is using a Swarm proprietary API, but probably not a big deal.

sberyozkin: Sure, see what you mean. Once we get MP-JWT connected with KC tokens then the portability will be back :-). Unfortunately, until then, if the service needs to do some application decisions based on the token content, it would be difficult to avoid using the proprietary API,  but for simple cases like getting the real user name we can 'hide' in the filters :-)

sberyozkin: Re that UUID, it does correspond to the "user1", I did not guess initially it could be seen in the test/demo realm json :-)

sberyozkin:
Windows tests look good

Ladicek (approved):

PR #869, opened 2018-04-09
SWARM-1947: GradleFileSystemLayout does not support customization of the build root

roguexz:
Motivation
----------
If a developer is using Gradle as the build tool and decides to customize the "build" directory of the project, then Arquillian tests start failing with the message:

    WFSWARM0012: Cannot identify FileSystemLayout for given path:

This is because the `GradleFileSystemLayout` looks for content in the default folder locations and is not customizable. It would be nice if the value can be customized and provided as system properties during the test execution stage.

Modifications
-------------
In GradleFileSystemLayout,

* Look for two system properties: `src.dir` & `build.dir`.
* If the properties have not been defined, then default them.
* Use the resolved values for locating additional resources.
* Collapsed the logic of Gradle4FileSystemLayout in to this class.

Additionally, updated the dependency adapter for the upcoming Gradle BOM support.

Result
------

-----


roguexz:
Additionally, can someone please point in the direction of updating the docs for Gradle tooling?

kenfinnigan:
retest this please

roguexz:
@kenfinnigan How does the arquillian daemon startup? I am looking at revamping some of the Gradle tooling that we in the project and I am trying to verify multiple test cases. I have the following questions,

1. When we say add all dependencies, we should ensure that the exported libraries are not included in the WEB-INF/lib folder. Is that correct?
2. How is the Arquillian daemon started? How is this library included?

roguexz:
For the linux build, is the tmp directory unique for each build (like being part of the workspace folder)? Or are they sharing the same location across all builds? If they are shared across all builds, then that might explain the enforcer rules failure. Not entirely sure, but just a guess.

roguexz:
retest this please

roguexz:
@kenfinnigan The build is green. We can merge it any time.

roguexz:
@kenfinnigan Can you please merge this PR? I am reworking the Gradle plugin and would prefer to have as little changes as possible.

heiko-braun (approved):

PR #870, opened 2018-04-09
SWARM-1944: MP FT - support @Asynchronous @Retry.

mkouba:
Motivation
----------
There is a bug in MP FT implementation.

Modifications
-------------
Fixed the relevant parts of the code.

Result
------
@Retry with @Asynchronous (and with fallback) works  if an exception is
thrown from within a business method.


heiko-braun (approved):

Ladicek:
LGTM post merge :-)

Ladicek:
Though I guess we could have moved the retry logic into `DefaultCommand` entirely -- we now have it split into 2 places, one for sync and one for async case.

mkouba:
Hm, for sync we create a separate command for each invocation/retry which is imho more correct. For async we only have one command and invoke/retry within its `run()`. It could be feasible though. I came to the project later when basic design decision were already made...

Ladicek:
Got it.

PR #874, opened 2018-04-10
SWARM-1782: Allow finalname of uberjar to be customized

leonardsson:
Allow finalname to be set in maven builds.

For gradle builds it will use the old behaviour, so the hardcoded "-swarm.jar " is moved from BuildTool.java to PackageTask.java.

If no name is set, it will default to the old value (finalName + (this.hollow ? "-hollow" : "") + "-swarm.jar").

- [ x] Have you created a [JIRA](https://issues.jboss.org/browse/SWARM) and used it in the commit message?
- [x ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/wildfly-swarm/wildfly-swarm/pulls) for the same issue?

-----

### DIFF
--- a/plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/PackageTask.java
+++ b/plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/PackageTask.java
@@ -273,7 +273,7 @@ public class PackageTask extends DefaultTask {
 
     @OutputFile
     private File getOutputFile() {
-        return BuildTool.getOutputFile(getBaseName(), getOutputDirectory());
+        return BuildTool.getOutputFile(getBaseName() + "-swarm.jar", getOutputDirectory());
     }
 
     private String getBaseName() {
--- a/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
+++ b/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
@@ -69,6 +69,9 @@ public class PackageMojo extends AbstractSwarmMojo {
     @Parameter(alias = "hollow", defaultValue = "false", property = "swarm.hollow")
     protected boolean hollow;
 
+    @Parameter(property = "finalName")
+    public String finalName;
+
     /**
      * Flag to skip packaging entirely.
      */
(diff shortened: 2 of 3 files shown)


plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
```diff
@@ -200,8 +203,12 @@ public void error(String msg, Throwable t) {
                 .forEach(tool::additionalModule);
 
         try {
-            File jar = tool.build(finalName + (this.hollow ? "-hollow" : ""), Paths.get(this.projectBuildDir));
-
+            String jarFinalName = finalName + (this.hollow ? "-hollow" : "") + "-swarm";
```
kenfinnigan, line 206: What's produced if `finalName` isn't set?

plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/PackageTask.java
```diff
@@ -273,7 +273,7 @@ private File getPropertiesFile() {
 
     @OutputFile
     private File getOutputFile() {
-        return BuildTool.getOutputFile(getBaseName(), getOutputDirectory());
+        return BuildTool.getOutputFile(getBaseName() + "-swarm.jar", getOutputDirectory());
```
kenfinnigan, line 276: How does Gradle get the `-swarm.jar` ending with this change?

The logic below is in the Maven plugin only

plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
leonardsson: final String finalName = this.project.getBuild().getFinalName();
so the default finalName of maven. If that's not set, it will default to ${project.artifactId}-${project.version} afaik.

plugins/gradle/src/main/java/org/wildfly/swarm/plugin/gradle/PackageTask.java
leonardsson: I don't know if i misunderstand you here, but you're looking at this change. BuildTool.getOutputFile is called from PackageTask.getOutputFile in the wildfly-swarm-gradle-plugin. Since I moved the name generation to one place for maven builds (PackageMojo.java) and thus removed the +"-swarm.jar" from the BuildTool.getOutputFile I also had to add the +"-swarm.jar" in the PackageTask.getOutputFile to keep the behaviour consistent with the previous implementation.

kenfinnigan: Sorry, was getting confused between removals and additions!

kenfinnigan:
ok to test

kenfinnigan (approved):

PR #876, opened 2018-04-10
SWARM-1953: MP JWT - resource without @RolesAllowed is secured.

mkouba:
Motivation
----------
Currently, if there is no @RolesAllowed declared on a JAX-RS resource,
no resource method is secured. This is rather a corner case but still a
valid use case.

Modifications
-------------
Fix MPJWTAuthExtensionArchivePreparer to also handle JAX-RS resource
classes which only declare @DenyAll/@PermitAll.

Result
------
JAX-RS resource without @RolesAllowed is secured (if needed).

### DIFF
--- a/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
+++ b/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
@@ -23,6 +23,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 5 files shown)


heiko-braun:
@Ladicek Would you like to review this one?

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
```diff
@@ -280,10 +283,13 @@ private SecurityConstraint createSecurityConstraint(WebXmlAsset webXml, String u
...
+        if (subpath.isEmpty()) {
+            // Remove the trailing slash
+            uriPath = fullAppPath.substring(0, fullAppPath.length() - 1);
+        } else if (subpath.charAt(0) == '/') {
+            uriPath = fullAppPath + subpath.substring(1);
         } else {
-            uriPath = fullAppPath.toString() + subpath;
+            uriPath = fullAppPath + subpath;
         }
         // If this uri includes a path param, truncate and add a wildcard
```
Ladicek, line 294: Off topic: this line makes me think about a bunch of interesting test cases...

Ladicek (approved):

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
Ladicek: Filed https://issues.jboss.org/browse/SWARM-1972

PR #877, opened 2018-04-10
SWARM-1964: MP-JWT should support public keys from JWKS URI

danpfe:
Motivation
----------
It is common that token issuers provide a convenient JWKS URI to distribute public keys (for signature-verification). Wildfly Swarm's MP-JWT fraction should be able to take advantage of these JWKS URIs rather than hard-code a public key in the configuration. The major benefits are that several issuers are supported (through "kid"-matching) and that expiring/changing public keys do not require modifications to an application's configuration.

Modifications
-------------
Several classes in the MP-JWT fraction (configuration-related classes, signature-validation of JWT) and an Arquillian-based testsuite to verify different configuration combinations and key-provider-mechanisms.

Result
------
The MP-JWT fraction can no be configured through MP-Config and Fraction-config to support a JWKS address for public keys ('jwksUri' and 'jwksRefreshInterval' have been added). Several signer keys are now supported (through "kid"-matching) and it's no longer mandatory to provide a static public key just as long as a JWKS address is provided. If both JWKS URIs and a static key is present, the latter is used.

Signed-off-by: Daniel Pfeifer <daniel@pfeifer.io>

-----


starksm64 (dismissed):
This looks good to me

Ladicek:
Sounds like SWARM-1950, too

EDIT: after looking at the content of this PR, I retract this comment. This is in the same area as SWARM-1950, but it's different. IMHO, SWARM-1950 (ability to configure a file path to a PEM-formatted public key) would still be a nice complement to this (ability to configure a JWKS URL).

kenfinnigan:
add to whitelist

danpfe:
@Ladicek SWARM-1950 and my pull request touch the same general area but don't solve the same problem (as you already concluded). I haven't seen a pull request for SWARM-1950 but after having tinkered around with the code for this fraction for a while I believe this could be added fairly easily without stepping on either proposal's toes.

Ladicek:
Thanks @danpfe, and agree. We'd just have to define what happens if a user (by accident) configures more than 1 of these properties (as they are mutually exclusive).

Actually that reminds me -- you already add a property that is mutually exclusive to an existing one (the PEM key literal). How do you handle that situation?

danpfe:
@Ladicek If a developer uses the MP-JWT Fraction and sets both a JWKS and a PEM, the PEM will take precedence over the JWKS URI. It's also documented accordingly so that there is no confusion in that regard.

Ladicek:
Ah OK, good, thanks!

Ladicek:
Just remembered: you should probably rebase and align the config keys (I mean the ones under `swarm.microprofile.*` that are used in the Swarm YAML config format). See https://issues.jboss.org/browse/SWARM-1963

danpfe:
@Ladicek Sure, can do. Will take the opportunity to modify the documentation string as it sounds a bit weird ("used if signer key isn't defined") and add a log-warning during server boot when both jwksUri and signerPubKey is set ( after all, who reads the documentation..... ;-) ).

Ladicek:
Thanks @danpfe, sounds good!

danpfe:
@Ladicek @starksm64 Hello guys, merged upstream into the branch and modified my code accordingly.

Ladicek:
I would expect an actual rebase instead of merge, but not sure if we have a "policy" or "guidelines" about this.

danpfe:
@Ladicek Hm, yeah. Guess so. Can rebase instead and force push again.

danpfe:
@Ladicek Now it should be fine.

Ladicek (dismissed):
There's some code duplication in the tests, suboptimal HTTP request handling (`fluent-hc` for the win!), it also implements JWT almost from scratch (in my standalone test suite, I used `jjwt`), but overall looks good to me.

I'd appreciate another pair of eyes, though!

sberyozkin:
The changes look good to me, and I'm waiting for this PR to go in so that I could extend the tests (as opposed to adding something to TCK) :-). I did notice too though that KeyTool and especially JwkTool can probably be optimized, though I'm not aware 100% right now the existing libs offer what is required for the tests

danpfe:
@Ladicek @sberyozkin Agree partially, but in this case I opted for the minimum dependency route - I didn't want to use things that weren't there to begin with.

About the duplication of the HTTP code in the test-methods: Agree with that, but at the time it was intentional because I am planning to contribute some more code to the fraction which is not strictly part of the JWKS-feature. If you'd prefer a small utility method for this I can quickly push it as well.

Ladicek:
@danpfe I understand why you didn't want to bring JWT-related dependencies, but we use the Fluent HTTP Client `org.apache.httpcomponents:fluent-hc` in a bunch of other test suite modules, and it really is much easier to read/write than `HttpURLConnection` :-)

sberyozkin:
@danpfe, Ladislav may also like JAX-RS 2.0 Client API :-)

Ladicek:
As a matter of fact, I hate JAX-RS Client, mostly because the `Client` isn't `AutoCloseable`, but also the interface is a lot painful to use compared to `fluent-hc` :-)

EDIT: but obviously anything is better than `HttpURLConnection`! :-)

danpfe:
I hear you guys but when it comes to that (the RESTEasy-based ClientBuilder), there are other fixes coming your way which are beyond the scope of this JWKS Pull Request :)

Ladicek:
I was just speaking about usage of HTTP client in the test :-)

But as I said, this PR is fine from my perspective.

sberyozkin:
Hi @danpfe, 
Just come across a somewhat similar situation to do with the kid driven key selection.
The way it works in your patch is that the server is no longer in control of which public key to use to validate - the kid is coming from the incoming JWT. So there is a minor attack surface there. So adding another property, 'kid' for example, will let protect against it if ever needed - this property will ensure that the resolved key has the 'kid' which the server expects. It would be an optional property, not required to be used by default, but can be used to restrict when required. The rotation can still work too. Right now, MP JWT has the implicit protection by providing a public key (inlined or soon - via the file).

sberyozkin:
Having said that, the actual attack is probably too theoretical (the attacker puts a kid pointing to the expired key which is still in the cache, and breaking the RS256 along the way :-) ), so I guess it should be safe enough to ignore

starksm64:
@danpfe, the behavior of the PEM key taking precedence over the JWKS key is the opposite of what we are proposing in the next rev of the MP-JWT spec. The reason being that JWKS provides more flexibility, and can still be used to embed PEM encoded keys using custom URL protocol handlers as described here: 
https://groups.google.com/forum/#!topic/microprofile/w3_yxbpzYTI

starksm64:
@sberyozkin, that the JWT kid is used as a selector from the JWKS is what we are expecting in the next rev of the MP-JWT spec. If an application is using multiple tokens with different signer keys, there has to be a way to select the public key from claims from the JWT, and the kid is the most common way this is done. The server configuration is still in control of the allowed keys since the JWKS is coming from the configured JWKS URI.

sberyozkin:
Hi @starksm64, thanks, sure the kid driving the selection from the JWK set is proven in itself. My only minor concern is about the case where a set contains several public keys while all the tokens are signed with a single private key hence the theoretical attack possibility probably exists. What I suggested would not go against the idea of using JWKS URI, it was only about having a config option at some point where a server can restrict kid to a known value...Even if you and others agree it may make sense, I do not propose to hold off this PR but rather follow later on with a minor update... Thanks

heiko-braun:
@danpfe It looks like another PR introduced changes to the same files. Could you rebase this?

sberyozkin:
Just to follow up on my earlier note, basically I got confused, in the lib I'm using it is possible to set a 'kid' property, the key 'alias', and if it is set then I believe it will be used before the 'kid' header will be tried - this was done for the only reason - to support the cases where the JOSE headers have no 'kid' property (as opposed to what I imagined was protecting against the attack).
I'm pretty sure I saw an OIDC RP test requirement at one stage where, if no 'kid' was provided then the key was still expected to be found if the set had the single key only. So the property I'm referring to might help in cases where a set has more than one key while the token has itself no 'kid' - whether MP-JWT will support such cases I'm not sure.
Daniel, All - sorry for the noise :-)

danpfe:
Sorry for being a bit late to the party, was busy with other stuff. Will rebase.

danpfe:
Rebased and resolved conflict. Enjoy :)

danpfe:
@starksm64 Sorry, didn't see your message. I could inverse this (i.e. JWKS taking precedence) if you feel we want to align towards the next iteration of MP-JWT.

danpfe:
@starksm64 Check out https://github.com/danpfe/wildfly-swarm/tree/inverseprecedence ... it's basically the same code but with the JWKS URI taking precedence. Could be merged into this PR without issues.

kenfinnigan:
retest this please

danpfe:
@kenfinnigan Hi, I did test and now reran everything again. Is there a specific problem that you noticed that I should check?

@starksm64 Did you check the proposal where I give the JWKS URI precedence?

kenfinnigan:
There wasn't, it was just that CI had died so we didn't have a valid CI job result to compare with.

Unfortunately it now looks like there's a conflict in the testsuite-microprofile-jwt/pom.xml that needs to be fixed

starksm64:
@danpfe, after the last MP-JWT call, we have changed things up so that the immediate value type config does take precedence, so the original ordering is fine. No reason to hold this up as it can be realigned with MP-JWT when 1.1 comes out.

danpfe:
@kenfinnigan Interesting that that file changed, it's completely new from my pull request... Hmm...

danpfe:
@kenfinnigan alright, see it now, someone else managed to get a new testsuite for MP-JWT merged... will see if stuff still works.

danpfe:
@starksm64 Alright, skipping that bit
@kenfinnigan All merged, builds and tests fine so let's hope that this was the golden push =)

sberyozkin:
@danpfe, originally there was a differently named test suite in the other PR, and then yet another PR followed - in both cases we tried to make sure it would align well with your PR to make it easier for you to merge.

sberyozkin:
Hi @danpfe, FYI, a minor alignment to the tests is on the way:
https://github.com/wildfly-swarm/wildfly-swarm/pull/905

danpfe:
@sberyozkin alright, got it

PR #879, opened 2018-04-11
Simplify POM files

roguexz:
Motivation
----------
When testing out changes locally, I tend to use `mvn versions:set -DnewVersion=xxx` a lot. As a habit, I tend to look at the changed files. In the current form, there are a lot of changes that show up in two POM files in particular - the root POM and the Netflix POM.

Modifications
-------------
* Use ${project.version} instead of an explicit version detail.
* Set `wildfly-swarm` as the parent of `build-resources` & `maven-enforcer-pattern-size`.
* Moved the enforcer plugin configuration from root to one level below. This is needed to avoid circular dependencies during POM evaluation.

Result
------

-----

### DIFF
--- a/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
+++ b/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/UberjarSimpleContainer.java
@@ -355,7 +355,7 @@ public class UberjarSimpleContainer implements SimpleContainer {
 
     @Override
     public void stop() throws Exception {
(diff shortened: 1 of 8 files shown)


roguexz:
@heiko-braun I pulled in only the required changes which do the following now.

1. `mvn versions:set -DnewVersion=xxx` will update the version of all projects except the docs.
2. I moved the build-resources & enforcer project under the main root project so that there are no unexpected surprises.

Would this structure be more amenable to your existing product implementation? If it doesn't work out, not a problem, I shall use it for my local testing alone then.

kenfinnigan:
As this is quite out of date with the current code I'm going to close it.

However, if you want to propose similar updates in an updated PR we can then take a look

PR #885, opened 2018-04-16
[SWARM-1951] Making the use of roles properties file optional

sberyozkin:


heiko-braun:
@Ladicek Is this the thing you wanted in ENTSWM too?

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/jaas/JWTLoginModule.java
```diff
@@ -98,4 +100,15 @@ protected JWTCallerPrincipal validate(JWTCredential jwtCredential) throws ParseE
...
     }
+
+    @Override
+    protected Group[] getRoleSets() throws LoginException {
+        if (options.containsKey("rolesProperties")) {
+            return super.getRoleSets();
+        } else {
+            Principal group =  subject.getPrincipals().stream()
+                .filter(p -> p instanceof Group && "Roles".equals(p.getName())).findFirst().get();
+            return new Group[] {(Group)group};
```
Ladicek, line 111: Is this exactly the same as what would happen if I specified an empty file in `rolesProperties`?

sberyozkin: JWTLoginModule sets the group in commit(). The proposed code will return this group bypassing the super.getRolesSet() only if no properties file (empty or not) is set. If it is an empty file, then RoleMappingLoginModule itself should be OK with that (processRoles()) though I guess the RBAC won't be enforced, it is a diff issue though

Ladicek: I want to be very cautious, as this is security-related.

Basically, if we want to make the `rolesProperties` option optional, then when it's missing, it needs to behave exactly like if it was set to an empty file.

Ladicek:
There's an ENTSWM for this, yes.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/jaas/JWTLoginModule.java
sberyozkin: I run the TCK with empty roles properties file and it worked, so in both cases (with the empty properties and the proposed code) 
https://github.com/MicroProfileJWT/microprofile-jwt-auth/blob/master/tck/src/test/resources/Token1.json#L14
is used as the roles, so I was wrong with my prev comment.
As far as the test is concerned, I can at least duplicate the current TCK and test without the properties, but I wonder if a better option exists :-)

heiko-braun:
test this please

heiko-braun:
@Ladicek Approved?

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/jaas/JWTLoginModule.java
Ladicek: Just looked at `RoleMappingLoginModule` and indeed this implementation looks like it will do exactly what would happen if I configured an empty file for `rolesProperties`. With one exception: if `subject.getPrincipals()` doesn't contain a `Group` named `Roles`, the code in `RoleMappingLoginModule` would return `null`, while this code will throw `NoSuchElementException`. It's probably unlikely, but...

Ladicek:
@heiko-braun Sort of, but I'd really like if we could get @starksm64 to review all our MP JWT JIRAs and approve on our intended course of action.

starksm64:
I'll take a look at this today.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/jaas/JWTLoginModule.java
sberyozkin: Hi Ladislav, JWT login module guarantees Group Roles exists since the "groups" is currently a required element so it adds it in its commit

Ladicek: Ah, it seems `getRoleSets` is guaranteed to be called only in `super.commit`. OK then, thanks for your patience! :-)

Ladicek:
The implementation looks fine to me and I'd just like an approval on the entire idea :-)

heiko-braun (approved):

heiko-braun:
@starksm64 Do you need more time for the review or are we good to go?

sberyozkin:
Hi All, I can try to add a test confirming that both not referencing the properties file or making it empty works as long as the JWT groups property contains the expected roles, once #890 goes in :-) 
IMHO, letting the users drop the properties from the SWARM config is completely safe from the spec compliance point of view, it is all about making it easier for the users when possible, as Ladislav noticed...

Ladicek:
I agree with @sberyozkin. I'm just extra-cautious, because security :-)

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/jaas/JWTLoginModule.java
starksm64: I think this is ok.

PR #886, opened 2018-04-17
SWARM-1981: Include infinispan in ‘full’

heiko-braun:


fractions/wildfly/infinispan/pom.xml
```diff
@@ -25,7 +25,8 @@
   <properties>
     <swarm.fraction.stability>stable</swarm.fraction.stability>
     <swarm.fraction.tags>Data</swarm.fraction.tags>
-  </properties>
+    <swarm.fraction.internal>true</swarm.fraction.internal>
```
Ladicek: We don't mark JGroups as internal in community. We probably should mark Infinispan as internal either.

Ladicek (approved):

PR #887, opened 2018-04-17
SWARM-1945: MP FT - refactor internals to support async circuit breaker.

mkouba:
Motivation
----------
Refactor the internals to support async circuit breaker. This will pose
various challenges related to unclear spec wording around async retry logic
and also related to the nature of Hystrix.

Modifications
-------------
Do not use HystrixCommand#queue() for async execution. Instead, to
workaround various problems of Asynchronous Retry combination,
create a "composite" command which wraps any async execution, i.e. executes
nested commands synchronously.

Note that if SynchronousCircuitBreaker is used (default) it is not possible
to track the execution inside a Hystrix command because a TimeoutException should be
always counted as a failure, even if the command execution completes normally.

Result
------
Async circuit breaker is supported. MP FT 1.0 TCK should pass with known
exludes (each exclude refers to a spec issue).

NOTES
------

@Ladicek @antoinesd  @heiko-braun @michalszynkiewicz @sberyozkin Guys, I really need your review here. It's a non-trival change and to be honest, I'm not very proud of this piece of code. I had to go back and forth many times during the last couple of days and this is probably the best version I came up with. Any tips for improvements would be appreciated.

Few 1.0 TCK tests fail but all excludes should have reasonable justification.

The most problematic are of course `@Asynchronous` + `@Retry` + `someOf(@Timeout,@Bulkhead,@CircuitBreaker)` combinations.

### DIFF
--- a/fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/HystrixCommandInterceptor.java
+++ b/fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/HystrixCommandInterceptor.java
@@ -27,6 +27,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 import javax.annotation.Priority;
@@ -40,9 +41,11 @@ import javax.interceptor.InvocationContext;
 
 import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.faulttolerance.Asynchronous;
 import org.eclipse.microprofile.faulttolerance.CircuitBreaker;
 import org.eclipse.microprofile.faulttolerance.Fallback;
 import org.eclipse.microprofile.faulttolerance.FallbackHandler;
+import org.eclipse.microprofile.faulttolerance.Retry;
 import org.eclipse.microprofile.faulttolerance.exceptions.BulkheadException;
 import org.eclipse.microprofile.faulttolerance.exceptions.CircuitBreakerOpenException;
 import org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceException;
@@ -56,6 +59,7 @@ import org.wildfly.swarm.microprofile.faulttolerance.deployment.config.FaultTole
 import org.wildfly.swarm.microprofile.faulttolerance.deployment.config.TimeoutConfig;
 
 import com.netflix.hystrix.HystrixCircuitBreaker;
+import com.netflix.hystrix.HystrixCommand;
 import com.netflix.hystrix.HystrixCommand.Setter;
 import com.netflix.hystrix.HystrixCommandGroupKey;
 import com.netflix.hystrix.HystrixCommandKey;
@@ -63,9 +67,21 @@ import com.netflix.hystrix.HystrixCommandProperties;
 import com.netflix.hystrix.HystrixThreadPoolKey;
 import com.netflix.hystrix.HystrixThreadPoolProperties;
 import com.netflix.hystrix.exception.HystrixRuntimeException;
+import com.netflix.hystrix.exception.HystrixRuntimeException.FailureType;
 
 /**
+ * <h2>Implementation notes:</h2>
+ * <p>
+ * If {@link SynchronousCircuitBreaker} is used it is not possible to track the execution inside a Hystrix command because a {@link TimeoutException} should be
+ * always counted as a failure, even if the command execution completes normally.
+ * </p>
+ * <p>
+ * We never use {@link HystrixCommand#queue()} for async execution. Mostly to workaround various problems of {@link Asynchronous} {@link Retry} combination. Instead, we
+ * create a composite command and inside its run() method we execute commands synchronously.
+ * </p>
+ *
  * @author Antoine Sabot-Durand
+ * @author Martin Kouba
  */
 @Interceptor
 @HystrixCommandBinding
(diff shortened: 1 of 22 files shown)


fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/RetryContext.java
```diff
@@ -51,23 +51,23 @@ RetryConfig getConfig() {
...
         // Decrement the retry count for this attempt
         remainingAttempts.decrementAndGet();
         // Check the exception type
         if (shouldRetryOn(throwable, System.nanoTime())) {
             delayIfNeeded();
-            return true;
+            return null;
         } else {
             if (throwable instanceof Error) {
                 throw (Error) throwable;
```
Ladicek: Should `return`?

fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/RetryContext.java
```diff
@@ -103,11 +103,16 @@ private boolean retryOn(Throwable throwable) {
...
-            long jitter = (long) (Math.random() * ((jitterBase * 2) + 1)) - jitterBase; // random number between -jitter and +jitter
-            TimeUnit.MILLISECONDS.sleep(delay + Duration.of(jitter, config.getJitterDelayUnit()).toMillis());
+    void delayIfNeeded() {
+        try {
+            if (delay > 0) {
+                long jitterBase = config.getJitter();
+                long jitter = (long) (Math.random() * ((jitterBase * 2) + 1)) - jitterBase; // random number between -jitter and +jitter
+                TimeUnit.MILLISECONDS.sleep(delay + Duration.of(jitter, config.getJitterDelayUnit()).toMillis());
+            }
+        } catch (InterruptedException e) {
```
Ladicek: Any reason why can't rethrow like we used to?

fractions/microprofile/microprofile-fault-tolerance/src/test/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/SimpleCommandTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.Test;
+
+public class SimpleCommandTest {
+
+    @Test
+    public void testGetCommandKey() throws NoSuchMethodException, SecurityException {
+        assertEquals("org_wildfly_swarm_microprofile_faulttolerance_deployment_SimpleCommandTest#testGetCommandKey()",
+                SimpleCommand.getCommandKey(SimpleCommandTest.class.getMethod("testGetCommandKey")));
+        assertEquals("org_wildfly_swarm_microprofile_faulttolerance_deployment_SimpleCommandTest$Foo#ping(java.lang.String,java.lang.Integer)",
+                SimpleCommand.getCommandKey(Foo.class.getMethod("ping", String.class, Integer.class)));
```
Ladicek, line 31: Test for method that has generic parameters (e.g. `List<String>`)?

fractions/microprofile/microprofile-fault-tolerance/src/test/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/async/circuitbreaker/AsynchronousCircuitBreakerTest.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+        // Circuit should be open now
+        try {
+            helloService.hello(Result.SUCCESS).get();
+            Assert.fail();
+        } catch (ExecutionException expected) {
+            assertTrue(expected.getCause() instanceof CircuitBreakerOpenException);
+        }
+        assertEquals(AsyncHelloService.THRESHOLD, COUNTER.get());
+    }
```
Ladicek: At this point, should add something like

```java
await().atMost(5, TimeUnit.SECOND).untilAsserted(() -> {
  assertEquals("Hello", helloService.hello(Result.SUCCESS).get());
});
```

to check that the CB eventually closes again.

fractions/microprofile/microprofile-fault-tolerance/src/test/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/bulkhead/BulkheadTest.java
```diff
@@ -38,16 +47,25 @@ public static JavaArchive createTestArchive() {
...
-                pingService.ping();
-            } catch (BulkheadException e) {
-                Assert.assertEquals("BulkheadException thrown for " + i, loop, i);
-                throw e;
-            }
+    @Test
+    public void testWaitingQueue(PingService pingService) throws InterruptedException, ExecutionException {
+        int loop = QUEUE_SIZE * 2;
+        CountDownLatch startLatch = new CountDownLatch(loop);
+        CountDownLatch endLatch = new CountDownLatch(1);
```
Ladicek, line 54: Nice, much better than what I have :-)

Ladicek (dismissed):
I'm not an expert on our MP FT implementation, but I can understand it enough to know that this PR uses the general approach I'd expect it to use, so +1 from me.

fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/RetryContext.java
mkouba: Yes, it should return. We've changed the method return type (see also javadoc ;-) and only throw `Error`s.

fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/RetryContext.java
mkouba: Hm, this method is used in `nextRetry()` so we need to try-catch but I think we could return the exception instead of creating ISE. Let me try this.

fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/RetryContext.java
Ladicek: No, I understand that the exceptions are to be `return`-ed now, was just wondering if this particular one should really be `throw`-n. It seems it should.

mkouba:
It seems there are many intermittent failures on windows machines, all probably related to https://github.com/eclipse/microprofile-fault-tolerance/issues/227. We can either exclude all tests from `BulkheadAsynchRetryTest` and `BulkheadSynchRetryTest` or maybe we could introduce a special exclude list for windows machines (activated by a profile)? Thoughts, ideas?

heiko-braun:
@mkouba I'd say we waive it through. @Ladicek any objections?

Ladicek:
Isn't https://github.com/eclipse/microprofile-fault-tolerance/issues/227 fixed in MP FT TCK? Can we adopt the fix?

If not, I'm fine with waiving for now, and if the failures repeat, we can exclude later.

mkouba:
@Ladicek It should be fixed but I don't think there will be TCK 1.0.1. @antoinesd Do you have more info?

fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/HystrixCommandInterceptor.java
```diff
@@ -113,111 +129,122 @@ public Object interceptCommand(InvocationContext ic) throws Exception {
...
-                syncCircuitBreaker = getSynchronousCircuitBreaker(metadata.commandKey, metadata.operation.getCircuitBreaker());
+    private static Object executeCommand(Function<Supplier<Object>, SimpleCommand> commandFactory, RetryContext retryContext, CommandMetadata metadata,
+            ExecutionContextWithInvocationContext ctx, SynchronousCircuitBreaker syncCircuitBreaker) throws Exception {
+        while (true) {
+            if (retryContext != null) {
+                LOGGER.debugf("Executing %s with %s", metadata.operation, retryContext);
             }
             Supplier<Object> fallback = null;
-            if (retryContext == null || retryContext.isLastAttempt() || (metadata.operation.isAsync() && metadata.operation.hasRetry())) {
+            if (retryContext == null || retryContext.isLastAttempt()) {
```
michalszynkiewicz, line 155: what does `retryContext == null` mean in here?

mkouba: No `@Retry` annotation used -> no retry policy.

heiko-braun (approved):
We ignore the intermittent windows failures

PR #888, opened 2018-04-19
SWARM-1917: Mail - allow to modify defaults through project-defaults.yml

mkouba:
Motivation
----------
It is not possible to modify attributes of the default mail session.

Modifications
-------------
Fix MailFraction so that it does not attempt to create a new mail
session when applying config values from project-defaults.yml.

Result
------
It should be possible to modify attributes of the default mail session.


Ladicek (approved):
Didn't try :-), but code LGTM.

PR #889, opened 2018-04-19
SWARM-1954: MP JWT - allow to skip unconstrained resource method.

mkouba:
Motivation
----------
If there is no class-level security annotation an unconstrained resource
method has always an implicit DenyAll. We would like to have this
configurable.

Modifications
-------------
Add
swarm.microprofile.jwt.default-missing-method-permissions-deny-access
config property (true by default). If set to false then if there are no
class-level security annotations an unconstrained resource method is
ignored, i.e. no security containts are implied.

Result
------
It is now possible to configure the default behavior.


Ladicek (approved):

heiko-braun:
test this please

PR #895, opened 2018-04-22
SWARM-1990: [Gradle] Package task is not generating artifact with the right name

roguexz:
Motivation
----------
The fix for SWARM-1782 has partly broken the Gradle tool. The final artifact file does not end with "-swarm.jar" any more. The package task should pass the final artifact name to the build tool.

Modifications
-------------
In PackageTask, ensure that we are passing the final artifact name to the build tool.

Additionally, I have included a rather simplistic attempt at solving the problem of bundledDependencies not working for projects built using Gradle. The problem is as follows,

Let's say you have two libraries ( A & B ) and both of them have a dependency on a library C but need two different versions. i.e.,

    A pulls C:1.0.0
    B pulls C:1.0.1

When the final deliverable is being built, the tool will ask the Gradle resolver to resolve artifacts for A, B, C(1.0.0) & C(1.0.1). Gradle will do its magic and skip the resolution of C-1.0.0 (since we have requested for a higher version). As you can see, the module system will start failing with this style of resolution.

While I work on providing a better fix (a much more wholistic solution), this should take care of most of the common scenarios.

Result
------

-----


roguexz:
@kenfinnigan We will need this fix before the next release goes out since the Gradle tooling is broken in master.

kenfinnigan:
retest this please

kenfinnigan (approved):

PR #896, opened 2018-04-23
[SWARM-1951] Adding the tests with the empty and missing properties

sberyozkin:


sberyozkin:
TestApplication is aligned with Daniel's test

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedEmptyPropsTest.java
```diff
@@ -0,0 +1,115 @@
...
+    @Deployment
+    public static Archive<?> createDeployment() throws Exception {
+        JAXRSArchive deployment = ShrinkWrap.create(JAXRSArchive.class);
+        deployment.addResource(RolesEndpointClassLevel.class);
+        deployment.addResource(TestApplication.class);
+        deployment.addAllDependencies();
+        deployment.addAsResource("emptyRoles.properties");
+        deployment.addAsResource("project-defaults.yml");
+        URL url = RolesAllowedEmptyPropsTest.class.getResource("/publicKey.pem");
+        deployment.addAsManifestResource(new UrlAsset(url), "/MP-JWT-SIGNER");
```
Ladicek: Could use `ClassLoaderAsset`?

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedEmptyPropsTest.java
```diff
@@ -0,0 +1,115 @@
...
+        return deployment;
+    }
+
+    @RunAsClient
+    @Test
+    public void testRolesAllowed() throws Exception {
+        String uri = "http://localhost:8080/mpjwt/rolesClass";
+        WebTarget target = ClientBuilder.newClient().target(uri).queryParam("input", "hello");
```
Ladicek: Yea, this is why I hate JAX-RS Client: it doesn't implement `AutoCloseable`, so noone bothers to properly close it. Especially in a test.

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedEmptyPropsTest.java
```diff
@@ -0,0 +1,115 @@
...
+    private static String createToken(String groupName) throws Exception {
+        String value = "{"
+                + "\"iss\": \"https://server.example.com\","
+                + "\"sub\": \"24400320\","
+                + "\"exp\": " + System.currentTimeMillis() + 36000 + ","
+                + "\"upn\": \"jdoe@example.com\","
+                + "\"groups\": ["
+                + " \"" + groupName + "\"" 
+                + " ]"
+                + "}";
```
Ladicek: Doesn't jose4j have a builder to construct the token in a type-safe fashion?

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedEmptyPropsTest.java
sberyozkin: What is wrong with URL asset ? And ".addAsResource" is used in most of the tests - it is very simple.

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedNoPropsTest.java
```diff
@@ -0,0 +1,121 @@
...
+        deployment.addResource(RolesEndpointClassLevel.class);
+        deployment.addResource(TestApplication.class);
+        deployment.addAllDependencies();
+        deployment.addAsResource("project-no-roles-props.yml");
+        URL url = RolesAllowedNoPropsTest.class.getResource("/publicKey.pem");
+        deployment.addAsManifestResource(new UrlAsset(url), "/MP-JWT-SIGNER");
+        return deployment;
+    }
+    @CreateSwarm
+    public static Swarm newContainer() throws Exception {
```
Ladicek: Could remove the entire method if `project-no-roles-props.yml` was added to the archive under a name of `project-defaults.yml`?

Ladicek:
Just minor comments, general approach LGTM.

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedEmptyPropsTest.java
Ladicek: Nothing wrong, would just save 1 line of code :-)

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedEmptyPropsTest.java
sberyozkin: What exactly is the problem here from the test's point of view ? Fluent HC does not really scale. Just because it auto closes does not make a preferred candidate IMHO. It will break as soon as one needs to introduce a filter or deal with more than a String. JAX-RS 2.0 client API is rich. I agree tat it is not perfect :-), I did have a lot of unhappiness about some parts of it some time ago but it is very capable nonetheless. FYI, Response InputStream is expected to be auto-closed by the spec for the typed reads

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedEmptyPropsTest.java
sberyozkin: Let me try...

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedNoPropsTest.java
sberyozkin: That would step over the default config which lists the empty properties ?

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedEmptyPropsTest.java
sberyozkin: Got it :-). I'm just copying bits and pieces I'm finding so not yet aware, thanks

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedNoPropsTest.java
sberyozkin: I see, I did not understand it first.
However, I tried 'addAsResource("project-defaults.yml", "project-no-roles-props.yml")' and removed CreateSwarm and that failed the test, I guess this is because the 2nd test uses its own "project-defaults.yml"

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/RolesAllowedEmptyPropsTest.java
sberyozkin: May be RestEasy specific Client implementation allows explicitly closing or even Auto closing, I don't know. JAX-RS Response itself has close() but I'm not sure how it is implemented in RestEasy - at te very least InputStream must be closed. And this InputStream must be closed if there was something like ".get(String.class)" - though in CXF it is only auto-closed if it was pre-configured to do so - otherwise there were side-effects with lazy reading DOM/XPath and similar types

sberyozkin:
Should be possible to deal with the connections for RestEasy by getting into HttpClient:
http://www.baeldung.com/resteasy-client-tutorial
May be some default handling is also done when the Client goes out of scope, hopefully that addresses your concerns a bit :-)

Update: no, RestEasy client is not auto-closeable. In general though it is hard to decide when it is safe to auto-close at the generic client implementation level which supports returning arbitrary types (via the custom MBReaders). I recall at some point we had an internal cache of Client instances weakly collected and being close()-ed - the clients dealing with DOM4J started failing. Ex: the function returns DOM4J and the Client goes out of scope inside that function, closing the streams if auto-closed, with the DOM tree still not being read, similarly for InputStream...

You are right though that for simple cases having to call close() is not looking nice :-)

Ladicek (dismissed):

Ladicek:
Much nicer, thanks! :-)

BTW, when I complain that the JAX-RS Client isn't `AutoCloseable`, I literally mean that the interface doesn't `extends AutoCloseable`. IMHO, try-with-resources is one of the best additions to Java ever.

sberyozkin:
Will update to the fluent client now since it was first on the master :-), and the tests are simple enough

Ladicek:
I certainly don't want you to rewrite this X times just because I don't like JAX-RS Client :-) IMHO, what we have now is perfectly fine.

sberyozkin:
Sorry, noticed it just now. Np at all, having both Fluent HC and JAX-RS in one test suite would be not too good :-). I was happy enough to do a bit of Fluent HC learning curve :-)

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/keys/ClasspathResourceTest.java
```diff
@@ -43,18 +39,12 @@
...
-        deployment.addClass(ResourceLoadingServlet.class);
-        deployment.addAsResource(pubKeyFile, "/public-key.pem");
+    public static Archive<?> createDeployment() throws Exception {
 
-        /*URL url = Thread.currentThread().getContextClassLoader().getResource("cp-resource.yaml");
-        assertThat(url).isNotNull();
-        File projectDefaults = new File(url.toURI());*/
+        JAXRSArchive deployment = ShrinkWrap.create(JAXRSArchive.class, "pubkey.war");
+        deployment.addResource(TestApplication.class);
+        deployment.addResource(KeyLoadingService.class);
```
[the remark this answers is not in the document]
sberyozkin: Np. I thought JAXRSArchive was somehow helping in glueing JAX-RS resources :-)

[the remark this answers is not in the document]
Ladicek: Personally, when I use `@Deployment` instead of `@DefaultDeployment`, I default to the standard ShrinkWrap `JavaArchive` or `WebArchive`, and only resort to the WildFly Swarm archive classes if I need to do something specific.

Ladicek (approved):

PR #900, opened 2018-04-25
SWARM-1993: support user classes as config sources for MP Config

michalszynkiewicz:
Motivation
----------
Before the changes, the only possible value for `swarm.microprofile.config.config-sources.KEY.attribute-class` property was:
```
 swarm.microprofile.config.config-sources.KEY.attribute-class:
    module: some.module
    name: my.package.MyClass
```

This change bring the possibility to use classes by class name, e.g:
```
swarm:
   microprofile:
      config:
         config-sources:
            default:
               attribute-class:
                com.test.microprofile.config.TestConfigSource
```

Modifications
-------------
MicroProfileConfigDeploymentProcessor has been added.
The processor checks if the value defined `swarm.microprofile.config.config-sources.KEY.attribute-class`
is a `String` and if it is, assumes it is a name of a class and adds a microprofile config service definition pointing to it.

Result
------
The YAML configuration for attribute-class works as described in the Motivation section

-----

### DIFF
--- a/fractions/microprofile/microprofile-config/src/main/java/org/wildfly/swarm/microprofile/config/fraction/runtime/MicroProfileConfigDeploymentProcessor.java
+++ b/fractions/microprofile/microprofile-config/src/main/java/org/wildfly/swarm/microprofile/config/fraction/runtime/MicroProfileConfigDeploymentProcessor.java
@@ -1,0 +1,101 @@
+/**
+ * Copyright 2015-2017 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.microprofile.config.fraction.runtime;
+
+import org.jboss.logging.Logger;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.Node;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.impl.base.path.BasicPath;
+import org.wildfly.swarm.spi.api.DeploymentProcessor;
+import org.wildfly.swarm.spi.api.config.CompositeKey;
+import org.wildfly.swarm.spi.api.config.ConfigView;
+import org.wildfly.swarm.spi.runtime.annotations.DeploymentScoped;
+
+import javax.inject.Inject;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ * <br>
+ * Date: 4/24/18
+ */
+@DeploymentScoped
+public class MicroProfileConfigDeploymentProcessor implements DeploymentProcessor {
+    private static final String CONFIG_SOURCE_SERVICES = "META-INF/services/org.eclipse.microprofile.config.spi.ConfigSource";
+    private static Logger log = Logger.getLogger(MicroProfileConfigDeploymentProcessor.class);
+
+    private final Archive<?> archive;
+    private final ConfigView configView;
+
+    @Inject
+    public MicroProfileConfigDeploymentProcessor(Archive<?> archive, ConfigView configView) {
+        this.archive = archive;
+        this.configView = configView;
+    }
+
+    @Override
+    public void process() throws Exception {
+        CompositeKey key = new CompositeKey("swarm", "microprofile", "config", "config-sources");
+
+        List<String> configSourceClasses = readConfigSourceClasses(key);
+
+        defineServices(configSourceClasses);
+
+        log.debugf("Configured MicroProfile config source classes %s", configSourceClasses);
+    }
+
+    private void defineServices(List<String> configSourceClasses) {
+        StringBuilder servicesContent = new StringBuilder();
+        servicesContent.append(readCurrentServices());
+        configSourceClasses
+                .forEach(source -> servicesContent.append(source).append("\n"));
+
+        archive.add(new StringAsset(servicesContent.toString()), new BasicPath(CONFIG_SOURCE_SERVICES));
+    }
+
+    private String readCurrentServices()  {
+        Node node = archive.get(CONFIG_SOURCE_SERVICES);
+        if (node != null) {
+            try (InputStream stream = node.getAsset().openStream();
+                 InputStreamReader streamReader = new InputStreamReader(stream);
+                 BufferedReader reader = new BufferedReader(streamReader)
+            ) {
+                return reader.lines()
+                        .filter(line -> !line.isEmpty())
+                        .collect(Collectors.joining("\n"));
+            } catch (IOException e) {
+                throw new IllegalStateException("Unable to parse " + CONFIG_SOURCE_SERVICES + " file", e);
+            }
+        }
+        return "";
+    }
+
+    private List<String> readConfigSourceClasses(CompositeKey key) {
+        return configView.simpleSubkeys(key)
+                    .stream()
+                    .map(subKey -> key.append(subKey).append("attribute-class"))
+                    .map(configView::valueOf)
+                    .filter(String.class::isInstance)
+                    .map(String.class::cast)
+                    .collect(Collectors.toList());
+    }
+}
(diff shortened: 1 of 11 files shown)


mkouba (changes requested):
Looks good but we should probably add a simple test?

fractions/microprofile/microprofile-config/src/main/java/org/wildfly/swarm/microprofile/config/fraction/runtime/MicroProfileConfigDeploymentProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+ * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ * <br>
+ * Date: 4/24/18
+ */
+@DeploymentScoped
+public class MicroProfileConfigDeploymentProcessor implements DeploymentProcessor {
+    private static final String CONFIG_SOURCE_SERVICES = "META-INF/services/org.eclipse.microprofile.config.spi.ConfigSource";
+    private static Logger log = Logger.getLogger(MicroProfileConfigDeploymentProcessor.class);
+
+    private final Archive<?> archive;
```
mkouba: I think you have to use `Archive` raw type here because of `org.wildfly.swarm.container.runtime.cdi.configurable.DeploymentProducer.archive()`...

michalszynkiewicz:
Also, something is wrong, the MP test suite fails with:
```
Caused by: org.jboss.weld.exceptions.DeploymentException: WELD-001408: Unsatisfied dependencies for type Archive<?> with qualifiers @Default
  at injection point [BackedAnnotatedParameter] Parameter 1 of [BackedAnnotatedConstructor] @Inject public org.wildfly.swarm.microprofile.config.fraction.runtime.MicroProfileConfigDeploymentProcessor(Archive<?>, ConfigView)
  at org.wildfly.swarm.microprofile.config.fraction.runtime.MicroProfileConfigDeploymentProcessor.<init>(MicroProfileConfigDeploymentProcessor.java:0)
```

fractions/microprofile/microprofile-config/src/main/java/org/wildfly/swarm/microprofile/config/fraction/runtime/MicroProfileConfigDeploymentProcessor.java
michalszynkiewicz: Will fix, I didn't expect it may make a difference...

mkouba: Yeah, CDI typesafe resolution + generics = expect troubles ;-)

See also: http://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#assignable_parameters

fractions/microprofile/microprofile-config/src/main/java/org/wildfly/swarm/microprofile/config/fraction/runtime/MicroProfileConfigDeploymentProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+        log.debugf("Configured MicroProfile config source classes %s", configSourceClasses);
+    }
+
+    private void defineServices(List<String> configSourceClasses) {
+        StringBuilder servicesContent = new StringBuilder();
+        servicesContent.append(readCurrentServices());
+        configSourceClasses
+                .forEach(source -> servicesContent.append(source).append("\n"));
+
+        archive.add(new StringAsset(servicesContent.toString()), new BasicPath(CONFIG_SOURCE_SERVICES));
```
Ladicek, line 73: What happens if the `CONFIG_SOURCE_SERVICES` file is already present?

Ladicek:
Also I think we actually don't really need this -- the big problem is that our documentation is misleading (wrong, in fact).

fractions/microprofile/microprofile-config/src/main/java/org/wildfly/swarm/microprofile/config/fraction/runtime/MicroProfileConfigDeploymentProcessor.java
michalszynkiewicz: `readCurrentServices` will read the ones in the file, the replacing file will contain them + the ones defined in YAML

michalszynkiewicz:
@juangon I used classes from your test app to create a test, hope you don't mind

michalszynkiewicz:
@Ladicek from user's perspective, providing a class in the yaml config is easier than creating a service, IMO. 
There's this ugly, meaningless key in the YAML (`default` in https://github.com/wildfly-swarm/wildfly-swarm/pull/900/files#diff-0ef26422feae1f53388ea68875ab1243R5), but otherwise it's not that bad ;)
And providing a class is much easier than providing a module and a class

michalszynkiewicz:
Linux failure: org.eclipse.microprofile.fault.tolerance.tck.bulkhead.BulkheadAsynchTest.testBulkheadClassAsynchronous10

heiko-braun:
@mkouba  is that one of the  intermittent failures in mp ft?

mkouba:
@Ladicek @michalszynkiewicz I agree with Michal that it might be easier to modify yaml instead of adding a service provider. On the other hand, 1. once we fix [SWARM-1998](https://issues.jboss.org/browse/SWARM-1998) the docs will be wrong again, 2. using service provider would make the app portable. So I think we should probably either create a new config property with proper description just for this specific case or instruct users to use a service provider instead of yaml.

@heiko-braun Probably yes. It seems that most Bulkhead tests are affected.

michalszynkiewicz:
should we close this PR then?

mkouba:
@michalszynkiewicz I'd wait a moment so that others can add their comment. @wildfly-swarm/core WDYT?

testsuite/testsuite-microprofile-config/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.arquillian.junit</groupId>
+      <artifactId>arquillian-junit-container</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.rest-assured</groupId>
+      <artifactId>rest-assured</artifactId>
```
Ladicek, line 49: `<scope>test</scope>`?

testsuite/testsuite-microprofile-config/src/main/java/org/wildfly/swarm/microprofile/config/test/HelloServlet.java
```diff
@@ -0,0 +1,74 @@
...
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+        resp.getWriter().println("{");
+        resp.getWriter().println("\"timeout\": " + appTimeout + ",");
+        resp.getWriter().println("\"optionalTimeoutFromConfig\": " +
+                config.getOptionalValue("app.timeout", Integer.class).orElse(-1) + ",");
+        resp.getWriter().println("\"propFromConfigSource\": \"" + propFromConfigSource + COMMA);
+        resp.getWriter().println("\"propFromConfigSourceFromService\": \"" + propFromConfigSourceFromService + COMMA);
+        resp.getWriter().println("\"propFromConfigSourceFromYaml\": \"" + propFromConfigSourceFromYaml + COMMA);
+        resp.getWriter().println("\"missingProperty\": \"" + missingProperty + "\"");
+        resp.getWriter().println("}");
```
Ladicek, line 71: JSON-P? :-)

Ladicek:
Adding a `ServiceLoader` file can be achieved by an annotation processor (I've seen a lot of people doing that). @mkouba has a very good point, too -- relying on Swarm YAML config makes the app non-portable. (It's technically not using MP Config as it is supposed to be used.)

Frankly, I can only imagine _one_ usecase for the WildFly MP Config subsystem's ability to define a config source by referring to a class/module pair: some other module in the app server defines a config source, and you want to enable it. All other usecases (defining a config source in the deployment, or in a dependency of the deployment) should IMHO be satisfied primarily by the `ServiceLoader` mechanism.

EDIT: CC @jmesnil -- Jeff, could you please explain what's the intention behind this attribute: https://github.com/wildfly-extras/wildfly-microprofile-config/blob/1.x/extension/src/main/java/org/wildfly/extension/microprofile/config/ConfigSourceDefinition.java#L72-L83 ?

testsuite/testsuite-microprofile-config/pom.xml
Ladicek: Ah sorry, didn't notice!

jmesnil:
This feature can be used to provide `ConfigSource` backed by other extensions or subsystems.
An example would be a DB ConfigSource that would use a JDBC DataSource from the `datasources` subsystem.
It was not meant to  be used for "user-provided" ConfigSource that are bundled in his application. This use case is covered by the ServiceLoader.

I see 2 issues with this PR:
* the feature is non-portable compared to using the ServiceLoader.
* it is not a good idea for Swarm to "override" the semantic of the `attribute-class` from the `microprofile-config` subsystem. The definition of this attribute is a tuple `(class-name, module)`.
This PR also allows that to be a `String` and change the whole runtime behaviour of the attribute. If this PR would be merged, I'd recommend to use another attribute (specific to Swarm, not backed by the `microprofile-config` subsystem).

Ladicek:
Thanks @jmesnil, that's what I thought. For me, the outcome is that we sorely need to improve our fraction reference docs.

mkouba:
Ok, so after reading previous comments I think we should close this PR and the corresponding issue and instead try to fix https://issues.jboss.org/browse/THORN-1998. @michalszynkiewicz Do you agree?

Ladicek:
Agree this should be closed.

michalszynkiewicz:
yes, done

PR #901, opened 2018-04-25
SWARM-1077: Make it possible to exclude auto-detected fractions.

mkouba:
Motivation
----------
Currently, there is no way to exclude auto-detected fractions.

Modifications
-------------
wildfly-swarm-plugin - enhance the functionality of the "fractions"
parameter.

Result
------
If a fraction value starts with the character ! a corresponding auto-detected
fraction is not installed (unless it is a dependency of any other
fraction).

**NOTE:** This is rather a low priority issue. Also I'm not quite sure whether the solution is correct and if the functionality is worth including.

### DIFF
--- a/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StartMojo.java
+++ b/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StartMojo.java
@@ -245,7 +245,7 @@ public class StartMojo extends AbstractSwarmMojo {
                 .map(d -> String.format("%s:%s", d.getGroupId(), d.getArtifactId()))
                 .collect(Collectors.toSet());
 
-        final Set<FractionDescriptor> fractions;
+        final Set<FractionDescriptor> detectedFractions;
         final FractionUsageAnalyzer analyzer = new FractionUsageAnalyzer(FractionList.get()).source(source);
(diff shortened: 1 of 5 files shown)


michalszynkiewicz:
I don't have enough proficiency in this area to approve, but I like this change :)

plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StartMojo.java
```diff
@@ -265,9 +265,15 @@ protected SwarmExecutor executor(final Path appPath, final String name,
                 .sorted()
                 .collect(Collectors.toList())));
 
-        fractions.addAll(this.additionalFractions.stream()
-                                 .map(f -> FractionDescriptor.fromGav(FractionList.get(), f))
-                                 .collect(Collectors.toSet()));
+        this.fractions.forEach(f -> {
```
kenfinnigan, line 268: I think it's worth renaming the local `fractions` variable to prevent confusion between it and `this.fractions`

mkouba: +1

kenfinnigan (approved):

PR #904, opened 2018-04-26
SWARM-1473: Back WARArchive by exploded temp dir.

mkouba:
Motivation
----------
Deployment processors can analyze/modify a repackaged WAR through
ShrinkWrap API. However, we currently use ZipImporter which loads
the whole archive in memory. Moreover, the WAR is currently imported
twice (see also StaticContentContainer).

As a result, an app with large repackaged WAR (libs cannot be replaced
by module deps) can run out of memory during bootstrap. E.g. cca 20MB
repackaged WAR + Xmx64m = OutOfMemoryError.

Modifications
-------------
Make use of exploded application artifact (tmp dir) and use ExplodedImporter, i.e.
the ShrinkWrap archive is backed by a directory structure.

Also skip the superfluous import in StaticContentContainer if the
archive was created by DefaultWarDeploymentFactory.

Result
------
Even an app with large repackaged WAR can be deployed. Also the overall memory
load during bootsrtap should be lower.

### DIFF
--- a/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/TempFileManager.java
+++ b/core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/TempFileManager.java
@@ -25,6 +25,7 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Collections;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * @author Bob McWhirter
@@ -64,6 +65,14 @@ public class TempFileManager {
         tmp.delete();
         register(tmp);
         return tmp;
+    }
+
+    public File getExplodedApplicationArtifact() {
+        return explodedApplicationArtifact.get();
+    }
+
+    public void setExplodedApplicationArtifact(File explodedApplicationArtifact) {
+        this.explodedApplicationArtifact.set(explodedApplicationArtifact);
     }
 
     private void register(File file) {
@@ -109,7 +118,9 @@ public class TempFileManager {
         return true;
     }
 
-    private Set<File> registered = Collections.newSetFromMap(new ConcurrentHashMap<>());
+    private final Set<File> registered = Collections.newSetFromMap(new ConcurrentHashMap<>());
+
+    private final AtomicReference<File> explodedApplicationArtifact = new AtomicReference<>();
 
     private File tmpDir;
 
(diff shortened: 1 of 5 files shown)


mkouba:
Windows build failures:
```
Test Result (2 failures / +1) 

org.eclipse.microprofile.fault.tolerance.tck.bulkhead.BulkheadAsynchTest.testBulkheadClassAsynchronousDefault

org.eclipse.microprofile.fault.tolerance.tck.bulkhead.BulkheadAsynchTest.testBulkheadMethodAsynchronous10
```

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/TempFileManager.java
```diff
@@ -109,7 +118,9 @@ public FileVisitResult postVisitDirectory(Path dir, IOException e)
         return true;
     }
 
-    private Set<File> registered = Collections.newSetFromMap(new ConcurrentHashMap<>());
+    private final Set<File> registered = Collections.newSetFromMap(new ConcurrentHashMap<>());
+
+    private final AtomicReference<File> explodedApplicationArtifact = new AtomicReference<>();
```
Ladicek, line 123: Not quite sure why this is a `AtomicReference` -- looking at what you do with it, making it `volatile` should be enough.

Ladicek:
I'm not sure about all our `DefaultDeploymentFactory` impls, but would the same work also for `DefaultJarDeploymentFactory`? That's probably the 2nd most often used deployment we have.

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/TempFileManager.java
mkouba: Right, `volatile` should be enough. I'm just used to use `AtomicReference` wrapper in similar scenarios as the "contract/API" is imho more clear and "it does not hurt" ;-). But if you feel like it's inappropriate here we can use `volatile` instead...

mkouba:
AFAIK `DefaultJarDeploymentFactory`'s deployment does not bundle any additional libraries. So it should not be a big deal compared to WARs.

Ladicek:
Good point re the JARs.

The PR looks good to me, but I don't feel totally confident in this area :-)

core/bootstrap/src/main/java/org/wildfly/swarm/bootstrap/util/TempFileManager.java
Ladicek: It's not inappropriate, it's just my habits are the other way around :-)

mkouba:
@Ladicek I'm not confident here either but thanks for review!

@juangon @bobmcwhirter Your opinion?

juangon:
@mkouba don't know how "heavy" is using AtomicXX comparing to volatile, but I tend to agree on setting variables with the simplest type that makes it working fine, if possible. BUT it's true setting as Atomic makes it more explicit. So, both seems fine? :-)

mkouba:
@wildfly-swarm/core If there are no objections I'm going to merge this PR today.

Ladicek (approved):

PR #907, opened 2018-04-30
SWARM-2009: Add fix for handling deployment of artifacts with a dot i…

fransflippo:
…n their artifactId.

Motivation
----------
Fix an exception (WFSWARM0009: Unable to determine version number from GAV) when an attempt
is made to deploy an artifact with a dot in its artifactId.

Modifications
-------------
Added a unit test for ArtifactDeployer (ArtifactDeployerTest) to demonstrate the issue. Fixed
the issue itself in ArtifactDeployer by using lastIndexOf('.') instead of split('\\.').
Other (minor) changes to ArtifactDeployer:
- Inverted the 'if' that checks for presence of a colon (:) in the deploy spec and added a TODO
  to consider if we should throw an exception when a colon does not exist (currently this
  case is silently ignored)
- If the artifact has no type specified (e.g. 'groupId:artifact' instead of 'groupId:artifact.jar'
  then 'jar' is now picked as artifact type. It made sense to add this given the changes in the
  code

Result
------
Changes are fully backwards compatible but fix the deployment of artifacts with a dot in their name
(most notably the IBM WebSphere MQ resource adapter whose artifactId is wmq.jmsra).
If the deploy spec does not specify a type, the deployment will now succeed if the type is 'jar'.
Previously this would have resulted in an ArrayIndexOutOfBoundsException as the 'parts' array would
have had only one element in this case.

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/runtime/ArtifactDeployer.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/runtime/ArtifactDeployer.java
@@ -44,16 +44,25 @@ public class ArtifactDeployer {
 
         for (SimpleKey subkey : subkeys) {
             String spec = subkey.name();
-            if (spec.contains(":")) {
-                String[] parts = spec.split(":");
-                String groupId = parts[0];
-                parts = parts[1].split("\\.");
-                String artifactId = parts[0];
-                String packaging = parts[1];
-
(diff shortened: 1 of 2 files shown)


fransflippo:
`mvn clean install` on full project fails: several tests in MicroProfile OpenAPI failing.
`mvn clean install` on just `core/container` passes.

kenfinnigan:
ok to test

kenfinnigan (approved):

PR #908, opened 2018-04-30
[SWARM-2003] Support for loading the custom classpath keycloak json resources

sberyozkin:


fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/runtime/SecuredArchivePreparer.java
```diff
@@ -84,21 +88,25 @@ private InputStream getKeycloakJsonFromCustomPath() {
...
 
             String appArtifact = System.getProperty(BootstrapProperties.APP_ARTIFACT);
 
             if (appArtifact != null) {
                 try (InputStream in = ClassLoader.getSystemClassLoader().getResourceAsStream("_bootstrap/" + appArtifact)) {
-                    Archive tmpArchive = ShrinkWrap.create(JARArchive.class);
+                    Archive<?> tmpArchive = ShrinkWrap.create(JARArchive.class);
                     tmpArchive.as(ZipImporter.class).importFrom(in);
-                    Node jsonNode = tmpArchive.get("keycloak.json");
+                    Node jsonNode = tmpArchive.get(resourceName);
```
Ladicek, line 103: This has nothing to do with this PR, but I've asked myself once or twice: could `keycloak.json` possibly contain private data? Because putting `keycloak.json` to the root directory of the WAR means it's accessible to the public, right?

sberyozkin: That said, if a browser client needs to be bootstrapped then the json has to be public. So I guess it may be needed assuming it is HTTPS (possibly 2-way TLS)

Ladicek: Totally agree that if we need to enable the Keycloak client in the browser, the JSON needs to be exposed -- but that would be a different JSON than the one used for the server? I'd guess.

Ladicek: Actually the secured booster show exactly this: there's a `webapp/kc.json` for the browser, and a wholly different `webapp/WEB-INF/keycloak.json` for the server. And `kc.json` isn't called `keycloak.json`, because if it would, Swarm would use it for configuring the server. Which we don't want.

sberyozkin: Just to copy what we chatted about on IRC: for a public client case it might be the same keycloak.json. (I guess especially if a link to the remote auth url is used, thus sharing no sensitive info). Perhaps removing this code makes sense anyway, but it would likely be safer to be done after the release...

kenfinnigan (approved):

fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/runtime/SecuredArchivePreparer.java
Ladicek: Yea as I said on IRC, this isn't / wasn't a request for change in this PR.

PR #910, opened 2018-05-02
SWARM-2015: old version of org.wildfly.swarm:config-api-runtime gets pulled in transitively in certain situations

Ladicek:
Motivation
----------
All the Config API projects depend on the WildFly Config API Runtime,
and they all refer to an old version. At the same time, it can easily
happen that the application `pom.xml` has a dependency on something
that brings MP Config / Keycloak / NoSQL / Teiid Config API earlier
than a dependency on something that would bring the core WildFly
Config API. Together, these two conditions lead to an old version
of WildFly Config API Runtime winning over the correct one, and that
results in runtime failures.

Modifications
-------------
The WildFly Swarm BOM now manages version of the WildFly Config API
Runtime to the correct version.

Result
------
WildFly Config API Runtime is always present in the correct version,
which prevents runtime failures.

-----

### DIFF
--- a/boms/src/main/resources/bom-template.xml
+++ b/boms/src/main/resources/bom-template.xml
@@ -50,6 +50,7 @@
     <version.shrinkwrap>${version.org.jboss.shrinkwrap}</version.shrinkwrap>
     <version.shrinkwrap-descriptors>${version.org.jboss.shrinkwrap.descriptors}</version.shrinkwrap-descriptors>
     <version.wildfly-swarm>${project.version}</version.wildfly-swarm>
+    <version.wildfly-config-api-runtime>${version.wildfly.config-api}</version.wildfly-config-api-runtime>
   </properties>
 
   <dependencyManagement>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -33,6 +33,7 @@
 
     <version.certified-community>2017.10.1</version.certified-community>
     <version.wildfly.swarm>${project.version}</version.wildfly.swarm>
+    <version.wildfly.config-api>1.3.1</version.wildfly.config-api>
```
michalszynkiewicz: was it failing with version in build-parent?

Ladicek: Actually that's a good point, the inheritance hierarchy is a bit more complex than I thought. I'll check again.

Ladicek: Thanks @michalszynkiewicz, indeed it worked fine even without moving the property. Nice :-)

sberyozkin:
I've confirmed the fix works at least for the management example (updated to add a keycloak fraction)

Ladicek:
Would be great if someone could review and hopefully merge :-)

bobmcwhirter (approved):

PR #916, opened 2018-05-08
[SWARM-2005] KC Multitenancy support

sberyozkin:
**Motivation**
At the moment it is not possible to configure a Thorntail instance to get KC validating the clients coming from the different realms - which is what can be expected in the production. This can be required not only for the complex deployments but even for simple cases, for example, the public browser and command line clients are different types of clients as far as their security characteristics are concerned and it can make sense to get them scoped by the different realms.   

The simplest way to support the dynamic loading of the KC adapter configurations is to have the path-based selection. Minimum effort should be required from the users. More involved selection strategies might be supported in the future. 

**Modifications**

* Introduced a "swarm.keycloak.multitenancy.paths" map property, the key is the relative path, the value - the custom KC json path; the query or headers or even claims based selection may be supported going forward
* Added a new KeycloakAdapterConfigResolver, it is statically initialized with the map of the deployments and added as a context parameter to the web.xml. 
* Even though SecurityArchivePreparer was updated recently to support "classpath:" URIs, it was further updated to try to get the resources from the classpath even without the "classpath:" to make the "swarm.keycloak.multitenancy" config look nicer
* Added a simple KeycloakMultitenancyTest - it is very similar to the original KeycloakArquillianTest but it resolves a KC config dynamically using a single entry deployments map  

**Result**

Users will now be able to start experimenting with the multi-tenancy support by doing a very simple Swarm/Thorntail configuration...
Experimenting with securing the management/health and the user resources with KC at the same time will also be possible


fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/deployment/KeycloakAdapterConfigResolver.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Map;
+
+import org.keycloak.adapters.KeycloakConfigResolver;
+import org.keycloak.adapters.KeycloakDeployment;
+import org.keycloak.adapters.OIDCHttpFacade;
+
+public class KeycloakAdapterConfigResolver implements KeycloakConfigResolver {
+
+    private static Map<String, KeycloakDeployment> deployments;
+    public static void setDeployments(Map<String, KeycloakDeployment> map) {
```
Ladicek: Can it be updated concurrently with processing? If so, the field should be `volatile`. In fact it should probably be `volatile` anyway, to ensure safe publication. Or maybe just make the `setDeployments` method `synchronized`, as we don't need to do a volatile read in the `resolve` method.

fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/runtime/SecuredArchivePreparer.java
```diff
@@ -143,10 +165,60 @@ private Asset createAsset(InputStream in) throws IOException {
...
+                    deployments.put(entry.getKey(), KeycloakDeploymentBuilder.build(is));
+                }
+            }
+            try {
+                // Statically initialize KeycloakAdapterConfigResolver
+                Module module = Module.getBootModuleLoader().loadModule("org.wildfly.swarm.keycloak:deployment");
+                final String resolverClassName = "org.wildfly.swarm.keycloak.deployment.KeycloakAdapterConfigResolver";
+                Class<?> resolverClass = module.getClassLoader().loadClass(resolverClassName);
+                Method setDeployments = resolverClass.getDeclaredMethod("setDeployments", Map.class);
+                setDeployments.invoke(null, deployments);
```
Ladicek: Whoah does it look weird. This so looks like it should be done elsewhere, but I really don't know where.

fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/deployment/KeycloakAdapterConfigResolver.java
sberyozkin: No it can't. A class name only is expected by KC. Check their multitenancy demo

fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/runtime/SecuredArchivePreparer.java
sberyozkin: This is how it is done in Swarm today

fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/deployment/KeycloakAdapterConfigResolver.java
sberyozkin: Besides, doing it concurrently would be a bad idea given the info is available at the start up

Ladicek: I guess there's a barrier somewhere on that path then :-)

fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/runtime/SecuredArchivePreparer.java
```diff
@@ -143,10 +165,61 @@ private Asset createAsset(InputStream in) throws IOException {
...
+    }
+
     private final Archive<?> archive;
 
-    @AttributeDocumentation("Path to keycloak.json configuration")
+    @AttributeDocumentation("Path to Keycloak adapter configuration")
     @Configurable("swarm.keycloak.json.path")
     String keycloakJsonPath;
 
+    @AttributeDocumentation("Keycloak multitenancy paths configuration")
```
Ladicek: Not very descriptive, also not sure how extensible it is, but whatevz :-)

sberyozkin: Now it should be more extensible :-). Originally I had it under 'swarm.keycloak.multitenancy' - but then thought that if if were to support the query or header or claim based dynamic config resolution then it would not fir all under 'swarm.keycloak.multitenancy'. Hence it is now ''swarm.keycloak.multitenancy.paths' for the path-based resolution, later if needed we can have "swarm.keycloak.multitenancy.queries", ".headers", etc, example: "headers: headerA" where the value of the headerA will be used to select a deployment from the 'headersDeployments' etc. Similarly for the claim-based resolution...

Let me try to update the doc a bit

Ladicek: Yea that's what I imagine, but no idea how these multiple config items will compose :-) But that's a bit premature I guess.

sberyozkin: They can probably be composed like this: check the path based deployments, next - queries, next headers. It is quite likely that for the cases where the dynamic resolution will require some more involved checks, say select a given deployment only if a given path and a given header and may be two JWT claims have some values, all at the same time, the users will just do it themselves.
The goal for this minor enhancement is to support few common mainstream cases and let users do it with just some basic config...
It would be good if we start getting some enhancement requests around it later on :-)

Ladicek: Ah that makes sense. Thanks!

Ladicek:
LGTM, with my favorite pet peeve about JAX-RS Client: the client instances aren't being closed :-)

sberyozkin:
Ladislav, refactored a bit the test code to close the clients, it made it look cleaner IMHO, thanks. Also opened RESTEASY-1898 as a result, so your feedback re the JAX-RS Client is very welcomed :-)

sberyozkin:
Windows build keeps failing due to the lack of space on the disk

Ladicek (approved):

Ladicek:
Thanks @sberyozkin!

sberyozkin:
retest this please

sberyozkin:
Windows build has failed for the fault tolerance, the test added with this PR passed:
https://ci.wildfly-swarm.io/job/wildfly-swarm-pull-request-windows/368/org.wildfly.swarm.testsuite$testsuite-keycloak/testReport/org.wildfly.swarm.keycloak/KeycloakMultitenancyTest/

So it is now all green as far as this PR is concerned and it has been approved, so merging it next

PR #918, opened 2018-05-09
[SWARM-1942] Ignoring the nexus staging plugin

sberyozkin:

### DIFF
--- a/testsuite/microprofile-tcks/pom.xml
+++ b/testsuite/microprofile-tcks/pom.xml
@@ -70,6 +70,13 @@
            </configuration>
          </plugin>
          <plugin>
+           <groupId>org.sonatype.plugins</groupId>
+           <artifactId>nexus-staging-maven-plugin</artifactId>
+           <configuration>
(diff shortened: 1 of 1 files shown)


testsuite/microprofile-tcks/pom.xml
```diff
@@ -69,6 +69,13 @@
              <skip>true</skip>
            </configuration>
          </plugin>
+         <plugin>
+           <groupId>org.sonatype.plugins</groupId>
+           <artifactId>nexus-staging-maven-plugin</artifactId>
+           <configuration>
+             <skip>true</skip>
+           </configuration>
+         </plugin>
```
Ladicek, line 78: Should it be in the entire testsuite parent too? (I know this one doesn't inherit from testsuite parent.)

Ladicek: Ah now I see the JIRA. Nevermind :-)

Ladicek (approved):

PR #923, opened 2018-05-16
THORN-2030: MP Rest Client - CDI interceptors support.

mkouba:
Motivation
----------
It would be useful to support CDI interceptors on MP rest client
proxies. The primary use case is to support MP FT annotations on rest
client interfaces.

Modifications
-------------
ProxyInvocationHandler invokes interceptor chains (if present). Every
proxy client now implements RestClientProxy so that it is possible to
access/close the unerlying client instance. It is also possible to
configure some of the ResteasyClientBuilder delegate properties using
RestClientBuilder#property().

Result
------
CDI interceptors can be bound to a rest client proxy using interceptor
bindings. MP FT annotations are supported and tested, except for
@Asynchronous - resteasy client proxy framework does not support async
return types.

### DIFF
--- a/client-apis/microprofile-restclient/src/main/java/org/wildfly/swarm/microprofile/restclient/InvocationContextImpl.java
+++ b/client-apis/microprofile-restclient/src/main/java/org/wildfly/swarm/microprofile/restclient/InvocationContextImpl.java
@@ -1,0 +1,168 @@
+/**
+ * Copyright 2018 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.microprofile.restclient;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.enterprise.inject.spi.InterceptionType;
+import javax.enterprise.inject.spi.Interceptor;
+import javax.interceptor.InvocationContext;
+import javax.ws.rs.client.ResponseProcessingException;
+
+/**
+ *
+ * @author Martin Kouba
+ */
+class InvocationContextImpl implements InvocationContext {
+
+    private final Object target;
+
+    private final Method method;
+
+    private Object[] args;
+
+    private int position;
+
+    private final Map<String, Object> contextData;
+
+    private final List<InterceptorInvocation> chain;
+
+    /**
+     * @param target
+     * @param method
+     * @param args
+     * @param chain
+     */
+    InvocationContextImpl(Object target, Method method, Object[] args, List<InterceptorInvocation> chain) {
+        this.target = target;
+        this.method = method;
+        this.args = args;
+        this.contextData = new HashMap<>();
+        this.position = 0;
+        this.chain = chain;
+    }
+
+    boolean hasNextInterceptor() {
+        return position < chain.size();
+    }
+
+    protected Object invokeNext() throws Exception {
+        int oldPosition = position;
+        try {
+            return chain.get(position++).invoke(this);
+        } finally {
+            position = oldPosition;
+        }
+    }
+
+    protected Object interceptorChainCompleted() throws Exception {
+        try {
+            return method.invoke(target, args);
+        } catch (InvocationTargetException e) {
+            if (e.getCause() instanceof ResponseProcessingException) {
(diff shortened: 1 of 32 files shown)


client-apis/microprofile-restclient/src/main/java/org/wildfly/swarm/microprofile/restclient/InvocationContextImpl.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    @Override
+    public Map<String, Object> getContextData() {
+        return contextData;
+    }
+
+    @Override
+    public Object getTimer() {
+        return null;
```
Ladicek, line 146: Maybe throw `IllegalStateException`?

client-apis/microprofile-restclient/src/main/java/org/wildfly/swarm/microprofile/restclient/InvocationContextImpl.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    static class InterceptorInvocation {
+
+        @SuppressWarnings("rawtypes")
+        private final Interceptor interceptor;
+
+        private final Object interceptorInstance;
```
Ladicek, line 154: Is this an _intercepted_ instance?

testsuite/testsuite-microprofile-restclient/src/test/java/org/wildfly/swarm/microprofile/restclient/ft/HelloClient.java
```diff
@@ -0,0 +1,69 @@
...
+    @GET
+    @Path("/hello")
+    String helloFallback();
+
+    @Fallback(fallbackMethod = "fallback")
+    @GET
+    @Path("/hello")
+    String helloFallbackDefaultMethod();
+
+    @GET // resteasy client proxies do not ignore default methods, see also RESTEASY-798 fixed in 3.5.1.Final
```
Ladicek, line 46: Wouldn't work without this annotation?

client-apis/microprofile-restclient/src/main/java/org/wildfly/swarm/microprofile/restclient/InvocationContextImpl.java
mkouba: I don't think so, `getTimer()` method's javadoc: _"returns the timer object, or null if the invocation did not apply to a timeout method"_.

Ladicek: Ah OK, sorry -- as I said, I don't really know anything about interceptors :-)

Ladicek:
Two bigger questions:

1. This doesn't _only_ add support for MP FT, but any CDI interceptor in general, right? (Or maybe just around-invoke interceptors?)

2. Is there a way to, I don't know, log a warning or something if CDI isn't present?

client-apis/microprofile-restclient/src/main/java/org/wildfly/swarm/microprofile/restclient/InvocationContextImpl.java
mkouba: No. It's really an *interceptor* instance (e.g. instance of [HystrixCommandInterceptor](https://github.com/wildfly-swarm/wildfly-swarm/blob/master/fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/deployment/HystrixCommandInterceptor.java)). Whereas the `Interceptor` represents interceptor metadata.

testsuite/testsuite-microprofile-restclient/src/test/java/org/wildfly/swarm/microprofile/restclient/ft/HelloClient.java
mkouba: Exactly. Fixed in RESTEASY-798.

mkouba:
@Ladicek 

> 1. This doesn't only add support for MP FT, but any CDI interceptor in general, right? (Or maybe just around-invoke interceptors?)

Yes, around-invoke interceptors ("Business Method Interceptors" in the spec language) are supported atm. I don't think that lifecycle and timeout methods make much sense in the context of rest client proxies.

> 2. Is there a way to, I don't know, log a warning or something if CDI isn't present?

Good point!

kenfinnigan:
Not sure I see the need to log anything with CDI not present, as I don't believe that would ever happen.

CDI is a dependency to Rest Client, so should always be there. Is that not the case?

mkouba:
> CDI is a dependency to Rest Client, so should always be there. Is that not the case?

@kenfinnigan  Yep, CDI classes will always be there but the CDI container may not be started (in theory, e.g. if your app does not contain `beans.xml` or an implicit bean archive and resteasy-cdi module will not be present - I believe this one turns JAX-RS resources into CDI beans). I think a simple warning wouldn't hurt.

testsuite/testsuite-microprofile-restclient/src/test/java/org/wildfly/swarm/microprofile/restclient/ft/HelloClient.java
Ladicek: Ah I see, we don't have the fix yet! Thanks, makes sense.

antoinesd (dismissed):

mkouba:
retest this please

antoinesd (approved):

PR #924, opened 2018-05-16
[THORN-2034] Introduce Jose Fraction

sberyozkin:
**Motivation**
See https://issues.jboss.org/browse/THORN-2034

**Modifications**
New "jose" fraction, testsuite-jose (testing the currently default Apache CXF implementation) and testsuite-jose-jose4j (testing Jose4j based one) are added. Apache CXF is 3.2.5-SNAPSHOT (to be released end of May - early June). The default implementation can be easily switched (as shown in the testsuite-jose-jose4j) as required. The feature has the experimental tag.

**Result**

Users will apply JOSE with ease to their application service and filter code. It is not tied to JAX-RS contexts and can be used anywhere the data integrity/confidentiality are important


sberyozkin:
Tests run without RunAsClient. The goal is to test a non-Inject lookup as well - given that Inject works in ApplicationScoped beans only while the Jose fraction is expected to work in different scopes

boms/src/main/resources/bom-template.xml
```diff
@@ -60,11 +59,6 @@
         <artifactId>spi</artifactId>
         <version>${project.version}</version>
       </dependency>
-      <dependency>
-        <groupId>org.wildfly.swarm</groupId>
-        <artifactId>config-api-runtime</artifactId>
-        <version>${version.wildfly-config-api-runtime}</version>
-      </dependency>
```
Ladicek: Whoops!

fractions/jose/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+
+  <name>JOSE Signing and Encryption Support</name>
+  <description>Supports JWS signing and JWE encryption</description>
+
+  <packaging>jar</packaging>
+
+  <properties>
+    <swarm.fraction.stability>experimental</swarm.fraction.stability>
+    <swarm.fraction.tags>Signing,Encryption</swarm.fraction.tags>
+    <version.cxf>3.2.5-SNAPSHOT</version.cxf>
```
Ladicek: Shouldn't be merged with a SNAPSHOT dependency I think.

fractions/jose/src/main/java/org/wildfly/swarm/jose/DecryptionOutput.java
```diff
@@ -0,0 +1,37 @@
...
+    private Map<String, Object> metadata;
+    private String data;
+    public DecryptionOutput(String data) {
+        this(Collections.emptyMap(), data);
+    }
+    public DecryptionOutput(Map<String, Object> metadata, String data) {
+        this.metadata = metadata;
+        this.data = data;
+    }
+    public Map<String, Object> getMetadata() {
```
Ladicek: Is the `Map` that callers can access through this method expected to be modified? If so, probably shouldn't use `Collections.emptyMap()` in the 1st constructor.

fractions/jose/src/main/java/org/wildfly/swarm/jose/EncryptionInput.java
```diff
@@ -0,0 +1,37 @@
...
+    private Map<String, Object> metadata;
+    private String data;
+    public EncryptionInput(String data) {
+        this(Collections.emptyMap(), data);
+    }
+    public EncryptionInput(Map<String, Object> metadata, String data) {
+        this.metadata = metadata;
+        this.data = data;
+    }
+    public Map<String, Object> getMetadata() {
```
Ladicek: Same here: is the `Map` that callers can access through this method expected to be modified? If so, probably shouldn't use `Collections.emptyMap()` in the 1st constructor.

fractions/jose/src/main/java/org/wildfly/swarm/jose/Jose.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    String verify(String jws) throws JoseException;
+
+    /**
+     * Verify the signed data in the JWS compact or JSON (optional) format.
+     * @param jws the JWS sequence.
+     * @return verified data and metadata
+     * @throws JoseException
+     */
+    VerificationOutput verification(String jws) throws JoseException;
```
Ladicek, line 53: Should be called `verify` for consistency?

fractions/jose/src/main/java/org/wildfly/swarm/jose/Jose.java
```diff
@@ -0,0 +1,103 @@
...
+    String verifyDetached(String jws, String detachedData) throws JoseException;
+
+    /**
+     * Verify the signed and detached data in the JWS compact or JSON (optional) format.
+     * @param jws the JWS sequence.
+     * @param detachedData the signed and detached data.
+     * @return verified data and metadata
+     * @throws JoseException
+     */
+    VerificationOutput verificationDetached(String jws, String detachedData) throws JoseException;
```
Ladicek, line 71: Same here, should be called `verifyDetached`?

fractions/jose/src/main/java/org/wildfly/swarm/jose/Jose.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    String decrypt(String jwe) throws JoseException;
+
+    /**
+     * Decrypt the encrypted data in the JWE compact or JSON (optional) format.
+     * @param jwe the JWE sequence.
+     * @return decrypted data and verified metadata
+     * @throws JoseException
+     */
+    DecryptionOutput decryption(String jwe) throws JoseException;
```
Ladicek, line 103: Same here, should be called `decrypt`?

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    public String contentEncryptionAlgorithm() {
+        return this.contentEncryptionAlgorithm.get();
+    }
+    /**
+     * Keystore type.
+     */
+    @Configurable("swarm.jose.keystore.type")
+    @AttributeDocumentation("Keystore type")
```
Ladicek: What are the possible values?

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+    private Defaultable<String> keystoreType = string(DEFAULT_KEYSTORE_TYPE);
+
+    /**
+     * Path to the keystore.
+     */
+    @Configurable("swarm.jose.keystore.path")
+    @AttributeDocumentation("Path to the keystore")
```
Ladicek: Is it only a filesystem path, or can I use a classpath resource too?

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+     */
+    @Configurable("swarm.jose.keystore.password")
+    @AttributeDocumentation("Password to the keystore")
+    private Defaultable<String> keystorePassword = string(DEFAULT_KEYSTORE_PASSWORD);
+
+    /**
+     * Signature algorithm.
+     */
+    @Configurable("swarm.jose.signature.algorithm")
+    @AttributeDocumentation("Signature algorithm")
```
Ladicek: What are the possible values?

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+    private Defaultable<String> signatureAlgorithm = string(DEFAULT_SIGNATURE_ALGORITHM);
+
+    /**
+     * Signature Format.
+     */
+    @Configurable("swarm.jose.signature.format")
+    @AttributeDocumentation("Compact or JSON JWS format, support for JSON is optional")
```
Ladicek: What are the possible values?

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+     */
+    @Configurable("swarm.jose.signature.key.alias")
+    @AttributeDocumentation("Alias to the signature key entry in the keystore")
+    private Defaultable<String> signatureKeyAlias = string(DEFAULT_KEY_ALIAS);
+
+    /**
+     * Encryption Format.
+     */
+    @Configurable("swarm.jose.encryption.format")
+    @AttributeDocumentation("Compact or JSON JWE format, support for JSON is optional")
```
Ladicek: What are the possible values?

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+    private Defaultable<String> encryptionFormat = string(DEFAULT_JOSE_FORMAT.name());
+
+    /**
+     * Key Encryption algorithm.
+     */
+    @Configurable("swarm.jose.encryption.keyAlgorithm")
+    @AttributeDocumentation("Key Encryption algorithm")
```
Ladicek: What are the possible values?

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+    private Defaultable<String> keyEncryptionAlgorithm = string(DEFAULT_KEY_ENCRYPTION_ALGORITHM);
+
+    /**
+     * Content Encryption algorithm.
+     */
+    @Configurable("swarm.jose.encryption.contentAlgorithm")
+    @AttributeDocumentation("Content Encryption algorithm")
```
Ladicek: What are the possible values?

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseFormat.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.wildfly.swarm.jose;
+
+public enum JoseFormat {
+    /**
+     * JOSE Compact representation
+     */
+    COMPACT,
+    /**
+     * JOSE JOSE representation: supporting this format is optional
```
Ladicek: One of the `JOSE` should probably be `JSON`?

fractions/jose/src/main/java/org/wildfly/swarm/jose/provider/JoseFactory.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Obtain the JoseFactory using the ServiceLoader pattern.
+     *
+     * @return the factory instance
+     */
+    public static JoseFactory instance() {
+        if (instance == null) {
+            synchronized (JoseFactory.class) {
+                if (instance != null) {
```
Ladicek, line 40: Looks like double-checked locking, which only works if `instance` is `volatile`.

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
Ladicek: Also, not quite sure as a user what "support for JSON is optional" means. What guarantees do I have?

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
Ladicek: Also, not quite sure as a user what "support for JSON is optional" means. What guarantees do I have?

boms/src/main/resources/bom-template.xml
sberyozkin: Oh, I thought I fixed it, I was testing your fix and it made it into the jose branch by accident and somehow it ended up deleted in the end, will put it back

fractions/jose/pom.xml
sberyozkin: Sure, you are right, the release is expected shortly, I'll update once it is there. I only wonder how to get it to the new repo without delaying the migration. May be remove the links to the fraction/tests from the root pom while SNAPSHOT is there...

fractions/jose/src/main/java/org/wildfly/swarm/jose/DecryptionOutput.java
sberyozkin: No, on the output it is not modifiable, these are the headers which came with the signed/encryped payload, they can contain the crypto related infor but also custom values, I will add the docs. And rename 'metadata' to 'headers' because this is what it is :-). On the input, it is the way to  customize the process further and also pass along the custom values - since the headers are also protected.

fractions/jose/src/main/java/org/wildfly/swarm/jose/Jose.java
sberyozkin: Compiler does not like it.

Ladicek: Ah of course! Sorry for the noise :-)

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
sberyozkin: Any type recognized by Java KeyStore (.jks is default) and additionally, "jwk" which is a collection of JWK keys

sberyozkin: I'll update the various docs a bit

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
sberyozkin: Right now it is only classpath; I haven't looked yet at supporting the external locations. I'm not sure how to handle it. I'd like to introduce a resolver interface at some point such that it works not only for say HTTPS URLs but also also custom mediums (hardware, etc).

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
sberyozkin: I'll update the docs for this and other attributes.
I've added the 'optional' note to let Jose4J be used which does not support it. I guess it is like with some of other Java APIs which are optional for the implementers. Or like (async) CompletionCallback. This is why I'm sticking with CXF for now at least as the default impl given that it supports it

sberyozkin: It should work for the users who may not need JSON format. I'm not sure yet how to express the fact the default impl guarantees it, given no guarantee exists CXF will remain as a default one, I guess we'll do it right at some point :-).

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
sberyozkin: Same as with the signature, Jose4j does not support this format for the encryption, Jose4j is primed for the compact sequences which are use in OAuth2/OIDC

fractions/jose/src/main/java/org/wildfly/swarm/jose/provider/JoseFactory.java
sberyozkin: Copying and pasting does not always work :-), let me fix it. I'll also do a minor MP-JWT PR

fractions/jose/src/main/java/org/wildfly/swarm/jose/Jose.java
sberyozkin: Np :-)

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
Ladicek: This is not a feature request, just a request for clarifying the doc text :-)

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
Ladicek: Thanks!

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
sberyozkin: I see, let me do it quickly

sberyozkin:
Hi @starksm64, can you please review as well ?

sberyozkin:
retest this please

starksm64:
@sberyozkin, I'll try to take a look at this Tuesday night UK time as I'm in Newcastle for a meeting. I have some MP-JWT changes that I have been testing out for the MP-JWT 1.1 release due out by June 1 that I'll have to merge into these efforts. Unfortunately it looks like I forgot to push the changes to my fork so I don't have anything for you to look at, but they changes are relatively minor.

sberyozkin:
Hi @starksm64, thanks, have a look please when you get some spare time. This new fraction is about the generic JOSE support at the Thorntail level. Ken agreed that when it eventually makes to the master it will only be available to the upstream users initially, so the product won't be affected until some interest for it is expressed.

sberyozkin:
retest this please

kenfinnigan:
retest this please

sberyozkin:
The docker subproject is failing all the time for this specific PR

sberyozkin:
retest this please

sberyozkin:
I probably need to close this pull request and create a new one

sberyozkin:
Reopening it as it has all the comments, sorry for the noise, Ken spotted it was related to the current SNAPSHOT dependency

kenfinnigan:
Is there a release for the dependency that can be used yet?

sberyozkin:
I've updated the versions, can have a look in a week on the status.

sberyozkin:
retest this please

sberyozkin:
both jose and jose4j tests pass on Linux, strange warning on Win around accessing a keystore resource, but all jose tests are shown as passed.
@kenfinnigan, @starksm64, what do you think ?

kenfinnigan:
The exception didn't appear to fail the tests, which is unusual.

Build appears to fail because the enforcer found temp files present

sberyozkin:
@kenfinnigan interesting, I've tried to move it above the Keycloak server tests, Keycloak server explodes all the themes to the temp folder by default (have a pending PR to optionally disable it)

sberyozkin:
It looks like it is due to the fact there are 13 tests there, 5/6 separate suites, and Arquillian does not clean them up in time or something like that. The tests themselves or the jose fraction does not create the temp files (as far as I'm aware). Not sure what else can be done to get the Win build progressing

sberyozkin:
H Martin, can you please have a look, when you are back, at the way testsuite-jose is setup ? I wonder if there is something there that can be improved to avoid it consistently failing on Win with the enforcer-plugin error... thanks

mkouba:
@sberyozkin the `BulkheadAsynchTest.testBulkheadClassAsynchronousDefault` failure can be safely ignored on windows.

sberyozkin:
@mkouba, @kenfinnigan if you OK with it then please consider merging; I look forward to doing some more work around it going forward, thanks

sberyozkin:
Windows build is green :-). Has been awhile since I saw it passing...

fractions/jose/src/main/java/org/wildfly/swarm/jose/DecryptionOutput.java
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.jose;
+
+import java.util.Collections;
+import java.util.Map;
+
+public class DecryptionOutput {
```
michalszynkiewicz, line 21: Is it worth to have two separate classes for decryption output and encryption input? 
They are the same and seem to represent the same thing - unencrypted data.

fractions/jose/src/main/resources/META-INF/beans.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<beans xmlns="http://xmlns.jcp.org/xml/ns/javaee">
+
+  <scan>
```
michalszynkiewicz, line 5: Does an empty `scan` element change anything?
From the docs I'd guess it doesn't have any meaning if it doesn't define any `exclude`s...

michalszynkiewicz (approved):

PR #930, opened 2018-05-21
THORN-2031: Project rename

mkouba:

-----

This PR needs two more updates:
1. Upgrade to `io.thorntail:parent:10` (once released)
2. Upgrade to `io.thorntail:fraction-plugin:87` (once released)


Ladicek:
The need to create artificial fraction classes looks suspicous. In my naive understanding, that shouldn't be necessary.

mkouba:
@Ladicek I agree that this hack is not very nice. The problem is that we decided NOT to change package and module names (which is imho reasonable and could ease migration, etc.). However, when `wildfly-swarm-fraction-plugin` generates fraction manifest: if there is some java code and any class named `*Fraction` then its package is used to derive a module name (e.g. `org.wildfly.swarm.undertow`), otherwise `groupId` + `artifactId` is used for module name (e.g. `org.wildfly.swarm.container` before but `io.thorntail.container` after the change -> bang, the build crashes).

We can probably find a better solution but for now my goal was to make the build work with minimal changes. In other wors, we can always improve this later.

Ladicek:
Aaaa gotcha! Thanks for explanation.

kenfinnigan:
ok to test

mkouba:
test this please

mkouba:
test this please

mkouba:
test this please

kenfinnigan:
retest this please

mkouba:
@kenfinnigan This PR is ready for final review/merge.

kenfinnigan (approved):

PR #931, opened 2018-05-21
[THORN-2038] Update the resolver to support the subresources and throw ISE to hide KC NPEs

sberyozkin:


fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/deployment/KeycloakAdapterConfigResolver.java
```diff
@@ -34,6 +34,7 @@ public static void setPathDeployments(Map<String, KeycloakDeployment> map) {
     @Override
     public KeycloakDeployment resolve(OIDCHttpFacade.Request request) {
         String path = request.getRelativePath();
-        return pathDeployments.get(path);
+        return pathDeployments.entrySet().stream().filter(e -> path.startsWith(e.getKey())).findFirst()
+            .map(e -> e.getValue()).orElseThrow(() -> new IllegalStateException("KeycloakDeployment is null"));
```
Ladicek: Better error message? Something like `"No Keycloak configuration for URL " + path`?

testsuite/testsuite-keycloak/src/test/java/org/wildfly/swarm/keycloak/KeycloakMultitenancyTest.java
```diff
@@ -83,9 +84,14 @@ public void testAccessResourceWithToken() throws Exception {
         
         Client client = ClientBuilder.newClient();
         try {
-            String serviceResponse = 
-                client .target("http://localhost:8080/secured")
-                    .request()
+            WebTarget target = client .target("http://localhost:8080/secured");
```
Ladicek: Space between `client .target`.

Ladicek:
Otherwise LGTM.

Ladicek:
Just thinking -- do we have docs for this somewhere? If yes, should explain ordering (first matching wins, so order should be from most specific to least specific).

Ladicek (dismissed):

sberyozkin:
Hi Ladislav, 
Thanks for the approval so far. Re the matching, I'm not sure yet the case will exist for having say 2 sub paths with the same root having diff realm config. For ex, "/a" having one requirement and and "/a/1" - some other one.
However, what I can do right now, is do a pure key lookup first, and only if that returns null - do a startsWith check - it will be faster for cases where a direct match exists, and will also make sure that even if we have something like "/a, /a1, /a2" then no matter how these 3 paths are configured the right config will be loaded if one adds individual path configs

Ladicek:
I don't think your change actually _avoids_ the ordering "issue". If you have multiple paths with the same prefix, order still matters. And I don't think that's wrong -- I just think we should mention this in some docs.

EDIT: I actually might be wrong, it's too late and my brain doesn't really work :-)

sberyozkin:
The initial exact match will give the users a chance not to worry about the ordering by adding as many precise keys as required. I'd hesitate to talk in the docs that one has to add "/a, then a/1 but not a/1 and then /a", that would be not cool at all :-), If we have a case of "a/1 and /a" and we do not want an "/a" request picking up the "/a/1" config then let the user add both keys. In the end of the day the filter is not really meant to cover all the possible non-standard variations - if the limitation is reached then let the users simply do a custom filter of their own :-)

sberyozkin:
Thinking more about, when we have a large number of paths/subpaths and the configuration is not the same for a given line, then indeed the ordering might help to make the configuration more optimal. Say given a/1, a/2, /a, with the last two have the same config, one could either have
/a: a.json
/a1: a1.json
/a2: a.json
(in any order)

or

/a1: a1.json
/a: a.json

and the more subpaths and diff config requirements we have the more savings can be made by the right ordering. 
However I'm not sure how generic the ordering recommendation can become :-) and how practical such cases are.
We can add something like "If you have multiple subpaths different KC config requitrements then either set the exact paths or configure more specific paths first". Not sure yet where to put a doc like that. May be in the doc in the resolver code at some point.

sberyozkin:
No, relying on the ordering just won't work, this would depend on the injectable Map impl be a LInkedHashMap, so if ever an ordering becomes an issue then it will be a sign that a custom filter is required :-)

Ladicek:
Ha, true with the linked map! So currently we get a random behavior in case one configured path is a prefix of another configured path. That's probably fine, if it's documented somewhere.

sberyozkin:
Only if the the users are not providing the exact keys, it is in their control :-). That said I'm fine with adding some comment, let me start with updating the attribute setter java docs

kenfinnigan:
retest this please

Ladicek (approved):
I don't think javadocs are used for generating the fraction reference, but this is probably good for now :-)

sberyozkin:
I was somewhat concerned about the verbosity, given AttributeDocumentation is probably expected to be a short string.

PR #946, opened 2018-05-23
THORN-2041: Adding support for Keycloak based security for Teiid ODat…

rareddy:
…a interface

- [ x] Have you followed the guidelines in our [Contributing](http://wildfly-swarm.io/community/contributing/) document?
- [x ] Have you created a [JIRA](https://issues.jboss.org/browse/SWARM) and used it in the commit message?
- [ x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/wildfly-swarm/wildfly-swarm/pulls) for the same issue?

-----

### DIFF
--- a/fractions/teiid/odata/src/main/java/org/wildfly/swarm/odata/ODataFraction.java
+++ b/fractions/teiid/odata/src/main/java/org/wildfly/swarm/odata/ODataFraction.java
@@ -15,6 +15,7 @@
  */
 package org.wildfly.swarm.odata;
 
+import static org.wildfly.swarm.spi.api.Defaultable.bool;
 import static org.wildfly.swarm.spi.api.Defaultable.string;
 
 import org.wildfly.swarm.spi.api.Defaultable;
@@ -22,9 +23,14 @@ import org.wildfly.swarm.spi.api.Fraction;
 import org.wildfly.swarm.spi.api.annotations.Configurable;
 
 public class ODataFraction implements Fraction<ODataFraction> {
-
-    @Configurable("swarm.odata.context")
+    @Configurable("swarm.teiid.odata.context")
     private Defaultable<String> context = string("odata4");
+
+    @Configurable("swarm.teiid.odata.secured")
+    private Defaultable<Boolean> secure = bool(false);
+
+    @Configurable("swarm.teiid.odata.role")
+    private Defaultable<String> role = string("odata");
 
     public String getContext() {
         return context.get();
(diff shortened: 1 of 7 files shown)


fractions/teiid/odata/src/main/java/org/wildfly/swarm/odata/ODataFraction.java
```diff
@@ -15,17 +15,23 @@
...
 import static org.wildfly.swarm.spi.api.Defaultable.string;
 
 import org.wildfly.swarm.spi.api.Defaultable;
 import org.wildfly.swarm.spi.api.Fraction;
 import org.wildfly.swarm.spi.api.annotations.Configurable;
 
 public class ODataFraction implements Fraction<ODataFraction> {
-
-    @Configurable("swarm.odata.context")
+    @Configurable("swarm.teiid.odata.context")
```
Ladicek, line 26: Might want to keep the original one for compatibility? (But add it as a _2nd_ annotation, as the 1st one is what is documented.)

rareddy: So, I would able to do two @Configurable annotations? Also I willing to break the compatibility for sake of correcting it.

Ladicek: I don't know how many users of this fraction are out there, but if you're willing to break compat, then fine :-)

rareddy: I don't think many yet, we are primary consumers of this in another OpenShift based UI project.

sberyozkin:
Should the KC fraction dependency be optional/provided ?

rareddy:
@sberyozkin It is optional as the user can choose to secure the service or not, but I was not sure how I can use the reference to the "Secured.class" in the code but make the pom.xml define the dependency as optional

sberyozkin:
I guess having a KC integration out of the box for this fraction is fine

rareddy:
Sure. if there is a way we can move Secured.class as a common annotation that would remove this need. If no further questions please consider a merge.

sberyozkin:
I've been looking at adding something similar to the management console, based on your PR. There I added 'keycloak.secured' as opposed to 'secured' only given that ManagementConsole can be secured using the other mechanisms. I've added 'keycloak.role' too. I guess Teiid OData can be secured differently as well ? Do you think it can make sense to update a little bit your PR ?
FYI, I've no rights to approve

sberyozkin:
I ended up going the different route as the Thorntail YAML config mostly works, so my comment that I added a "keycloak.secured" property is no longer true, for now at least, IMHO it might make sense to qualify it in your case too but it is up to you whether to update it, may be securing Teiid OData with KC is the only reasonable option and as such no need to do the extra qualification...thanks

rareddy:
@sberyozkin On standalone Teiid, we supported HTTPBasic and OAuth with Keycloak, but on OpenShift we want to tighten the offering to Keycloak.

kenfinnigan (approved):

PR #947, opened 2018-05-24
add initial HibernateSearch support, closes #936

### LINKED ISSUE #936: Hibernate Search
Evaluate what can be done regarding integration with Hibernate Search.

Is it just a testsuite? Is there "sugar" we can add on top to simplify?

kg6zvp:
- [x] Have you created a GitHub Issue and used it in the commit message?

-----
All `@PersistenceContext` annotated `EntityManager`s are `FullTextEntityManager`s by default if Hibernate Search is on the classpath.

Does not inject `QueryBuilder`. This feels like it should be integrated using a CDI service with a custom annotation so that it can be scoped to a specific PU. I didn't see anything like this in Seam docs, so this is new territory, but I'd love for some expert input.

### DIFF
--- a/core/hibernate-search/src/main/java/io/thorntail/hibernate_search/impl/QueryFactory.java
+++ b/core/hibernate-search/src/main/java/io/thorntail/hibernate_search/impl/QueryFactory.java
@@ -1,0 +1,27 @@
+package io.thorntail.hibernate_search.impl;
+
+import javax.enterprise.inject.Produces;
+import javax.enterprise.inject.spi.InjectionPoint;
+import org.hibernate.search.jpa.FullTextEntityManager;
+import org.hibernate.search.query.dsl.QueryBuilder;
+import io.thorntail.hibernate_search.QueryBuilderContext;
+import static io.thorntail.hibernate_search.impl.InjectionPointUtils.getAnnotation;
+import javax.enterprise.inject.Disposes;
+
+/**
+ * @author kg6zvp
+ */
+public class QueryFactory {
+    @Produces
+    @QueryBuilderContext(Object.class)
+    public QueryBuilder getQueryBuilder(InjectionPoint ip) {
+        QueryBuilderContext qbc = getAnnotation(ip, QueryBuilderContext.class);
+        FullTextEntityManager em = EntityManagerUtils.getFullTextEntityManager(qbc.persistenceUnit());
+        QueryBuilder qb = em.getSearchFactory().buildQueryBuilder().forEntity(qbc.value()).get();
+        return new QueryBuilderDelegate(qb, em);
+    }
+
+    public void disposeQueryBuilder(@Disposes @QueryBuilderContext(Object.class) QueryBuilder qb) {
+        ((QueryBuilderDelegate)qb).close();
+    }
+}
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -106,6 +106,7 @@
     <module>bean-validation</module>
     <module>datasources</module>
     <module>health</module>
+    <module>hibernate-search</module>
     <module>jaxrs</module>
     <module>jca</module>
     <module>jdbc</module>
--- a/testsuite/simple/hibernate-search/src/test/java/io/thorntail/testsuite/hibernate_search/HibernateSearchAppTest.java
+++ b/testsuite/simple/hibernate-search/src/test/java/io/thorntail/testsuite/hibernate_search/HibernateSearchAppTest.java
@@ -1,0 +1,54 @@
+package io.thorntail.testsuite.hibernate_search;
+
+import io.thorntail.hibernate_search.EntityManagerContext;
+import io.thorntail.hibernate_search.QueryBuilderContext;
+import io.thorntail.hibernate_search.impl.QueryBuilderDelegate;
+import io.thorntail.test.ThorntailTestRunner;
+import javax.inject.Inject;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import javax.persistence.EntityManager;
+import javax.persistence.PersistenceContext;
+import static org.assertj.core.api.Assertions.assertThat;
+import org.hibernate.search.jpa.FullTextEntityManager;
+import org.hibernate.search.query.dsl.QueryBuilder;
+
+/**
+ * @author kg6zvp
+ */
+@RunWith(ThorntailTestRunner.class)
+public class HibernateSearchAppTest {
+    @PersistenceContext
+    EntityManager em;
+
+    @Inject
+    FullTextEntityManager searchEm;
+
+    @Inject
+    @EntityManagerContext("SearchPU")
+    FullTextEntityManager searchEmScoped;
+
+    @Inject
+    @QueryBuilderContext(AppUser.class)
+    QueryBuilder userQueryBuilder;
+
+    @Inject
+    @QueryBuilderContext(value = AppUser.class, persistenceUnit = "SearchPU")
+    QueryBuilder userQueryBuilderScoped;
+
+    @Test
+    public void test() {
+        assertThat(em).isNotNull();
+        assertThat(em).isInstanceOf(FullTextEntityManager.class);
+
+        assertThat(searchEm).isNotNull();
+        assertThat(searchEmScoped).isNotNull();
+
+        assertThat(userQueryBuilder).isNotNull();
+        assertThat(userQueryBuilder).isInstanceOf(QueryBuilderDelegate.class);
+
+        assertThat(userQueryBuilderScoped).isNotNull();
+        assertThat(userQueryBuilderScoped).isInstanceOf(QueryBuilderDelegate.class);
+    }
+}
--- a/testsuite/simple/pom.xml
+++ b/testsuite/simple/pom.xml
@@ -17,6 +17,8 @@
   <modules>
     <module>bean-validation</module>
     <module>config</module>
+    <module>hibernate-search</module>
+    <module>hibernate-search-should-fail</module>
     <module>jaxrs</module>
     <module>jaxrs-jsonp</module>
     <module>jpa</module>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -572,6 +572,11 @@
       </dependency>
       <dependency>
         <groupId>io.thorntail</groupId>
+        <artifactId>thorntail-hibernate-search</artifactId>
+        <version>4.0.0-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>io.thorntail</groupId>
         <artifactId>thorntail-ogm</artifactId>
         <version>4.0.0-SNAPSHOT</version>
       </dependency>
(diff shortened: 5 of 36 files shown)


testsuite/simple/pom.xml
```diff
@@ -23,6 +23,7 @@
     <module>jpa-contracts</module>
     <module>jpa-jta</module>
     <module>jpa-opentracing</module>
+    <module>hibernate-search</module>
```
Ladicek: The list looks alphabetically sorted, which your addition here breaks :-)

testsuite/simple/hibernate-search/src/test/java/io/thorntail/testsuite/hibernate_search/HibernateSearchAppTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+import javax.persistence.EntityManager;
+import javax.persistence.PersistenceContext;
+import static org.assertj.core.api.Assertions.assertThat;
+import org.hibernate.search.jpa.FullTextEntityManager;
+
+@RunWith(ThorntailTestRunner.class)
+public class HibernateSearchAppTest {
+    @PersistenceContext
+    EntityManager em;
```
Ladicek: Can also test that it's possible to do `@PersistenceContext FullTextEntityManager em;`?

core/pom.xml
```diff
@@ -114,6 +114,7 @@
     <module>jms-artemis</module>
     <module>jpa-support</module>
     <module>jpa</module>
+    <module>hibernate-search</module>
```
Ladicek: Also breaks alphabetical order.

core/hibernate-search/src/main/java/io/thorntail/hibernate_search/HibernateSearchFraction.java
```diff
@@ -0,0 +1,6 @@
+package io.thorntail.hibernate_search;
+
+/**
+ * @author kg6zvp
+ */
+public class HibernateSearchFraction { }
```
Ladicek: Are fraction classes needed in v4?

kenfinnigan: Yes, this isn't needed

core/pom.xml
kg6zvp: Will correct. I was not aware there was a convention

testsuite/simple/hibernate-search/src/test/java/io/thorntail/testsuite/hibernate_search/HibernateSearchAppTest.java
kg6zvp: I looked into that quite extensively and from what I can tell it would involve a re-write of some of Weld's internal classes.

I can't remember which file it's in, but I think the line is something like `Class.forName("javax.persistence.PersistenceContext");` in the `weld/api` repo. I am not certain it can't be done, but it appears to be very difficult. I'd love to get a Weld expert to weigh in on this. 👍

core/hibernate-search/src/main/java/io/thorntail/hibernate_search/HibernateSearchFraction.java
kg6zvp: This is to force jar packaging

testsuite/simple/hibernate-search/src/test/java/io/thorntail/testsuite/hibernate_search/HibernateSearchAppTest.java
Ladicek: Ah, so can we at least `@Inject` it? By no means do I insist on `@PersistenceContext` :-)

kg6zvp: No, frankly I was hoping another dev had a silver bullet for this one to allow `@PersistenceContext`. I'll use a Qualifier for the PU on this and `Query` injection to match old Seam behavior.

kg6zvp: I considered writing an `InjectionTarget<T>` wrapper or trying to register `FullTextEntityManager` as a delegate, but couldn't find anything on how to register a delegate programmatically.

In the meantime, I'll push updated code with `@Inject` when I'm free later today.

scottmarlow: Does the Hibernate ORM EntityManager#unwrap method know how to obtain the FullTextEntityManager?

kg6zvp: Not from what I saw in the Hibernate search docs (I had the same thought initially), but that would be really nice. I will try that later today when I get to a computer.

core/pom.xml
kenfinnigan: More a convenience than convention

core/hibernate-search/src/main/java/io/thorntail/hibernate_search/HibernateSearchFraction.java
kenfinnigan: package-info.java isn't enough?

kg6zvp: Will verify when I get access to my computer. I believe I had tried that initially, though.

core/hibernate-search/src/main/java/io/thorntail/hibernate_search/impl/QueryFactory.java
```diff
@@ -0,0 +1,48 @@
...
+public class QueryFactory {
+    @PersistenceUnit
+    EntityManagerFactory emf;
+
+    @Produces
+    @QueryBuilderContext(Object.class)
+    public QueryBuilder getQueryBuilder(InjectionPoint ip) {
+        QueryBuilderContext qb = getQueryAnnotation(ip);
+        if (qb == null) {
+            throw new RuntimeException("@QueryBuilderQualifier not present on QueryBuilder InjectionPoint"); //Will be replaced with custom exception and better message if QueryBuilder feature is required
```
kg6zvp: This is where using `@Inject` on `QueryBuilder` gets messy, as each instance of a `QueryBuilder` pertains to a single Entity class. It was mentioned this would be a handy inclusion, but this seems messy.

Feedback and comments appreciated as always.

testsuite/simple/hibernate-search/src/test/java/io/thorntail/testsuite/hibernate_search/HibernateSearchAppTest.java
kg6zvp: @scottmarlow It looks like the answer is no, unwrap does not understand that. It makes sense since ORM doesn't depend on Search.

core/hibernate-search/src/main/java/io/thorntail/hibernate_search/impl/QueryFactory.java
kenfinnigan: I'm not certain, but I'm not sure `Object.class` is needed on `@QueryBuilderContext`, or there's some other way to make it generic. @mkouba should know better than me.

Also, If we're producing an instance for `@QueryBuilderContext`, should qb ever be null? If it was wouldn't that imply we're producing an instance for an injection point that doesn't have the annotation?

kg6zvp: Using Object.class was the only way to let `@QueryBuilderContext` require a `Class` property.

We're producing an instance of `QueryBuilder`. `@QueryBuilder`Context` is an annotation I added because instances of `QueryBuilder` themselves must be built for a specific entity class.

As to `@QueryBuilderContext` being null, I will add an observer for the event that CDI fires as it is scanning fields annotated with `@Inject` if it turns out we actually want `QueryBuilder` injected. The observer would act as a validator to make sure every `QueryBuilder` field annotated with `@Inject` has a `@QueryBuilderContext` annotation on it. It could also validate that the given `Class` is an entity in that persistence unit, but I don't think there's a need since Hibernate Search would throw an error if a `QueryBuilder` was attempted to be built without a valid `@Entity`-annotated class belonging to the default persistence unit.

kenfinnigan: I don't think you need to observe the event or anything like that. My understanding is that by virtue of the `@Producer` being marked with `@QueryBuilderContext` that an injection point that just specifies a return type of `QueryBuilder` without the annotation would not trigger the producer.

Maybe it's worth using a CDI extension to output a warning that an injection for `QueryBuilder` was made without a corresponding `@QueryBuilderContext` so an instance won't be injected? Now realizing that this is maybe what you meant above

kg6zvp: @kenfinnigan Yes, that is dead code from an earlier implementation.

Yes, CDI `Extension` is exactly what I meant as they are allowed to observe such lifecycle events. Do we have a convention for custom errors that I should follow?

kenfinnigan: Nothing other than them being created in a message class like https://github.com/thorntail/thorntail/blob/4.x/core/jpa-support/src/main/java/io/thorntail/jpa/impl/JpaMessages.java

Sanne:
Starting to look good!

By default Hibernate Search indexes objects into an Apache Lucene index, stored on disk.

For integration tests I would recommend to use `hibernate.search.default.directory_provider=local-heap`.

This property can be set as a system property, environment variable or just listed in the `persistence.xml` among the other Hibernate configuration properties.

kg6zvp:
@sanne Awesome! Thanks for the tip. I was on the fence about that in the tests, but I think it also serves as a good example.

I'll begin writing the documentation up. Requests welcome.

Sanne:
I don't know much about injection, so not sure if it's solvable but I see a limitation:

`@PersistenceContext` can take some attributes as one might have multiple persistent units in the same application, e.g. connecting to two different databases is not extremely uncommon.

Would you need to have some (optional?) qualifier for the various injection points, to clarify which persistent unit to relate to?

scottmarlow:
Is this something that could be done by adding Search support for ORM
EntityManager.unwrap(Class ).  Then, see if the WildFly extension for
injecting PersistenceContext into an EM unwrap class, works?

Apps can do this today to inject a Hibernate Session, they just need to be
mindful of the container managed entity manager scope.

Scott

Sanne:
@scottmarlow that's an interesting idea we could do in ORM, but this version has shipped and WildFly won't upgrade for months.

Also, I'm not sure how that would allow people to qualify their injection points.

kg6zvp:
@Sanne Agreed on `@PersistenceContext`. I have added a qualifier for it now.

@Sanne Documentation has been added. Depending on whether your `JPA22` branch or this gets merged first, one of us will need to deal with the Hibernate Search version referred to in the docs. Feedback on the docs especially appreciated.

kenfinnigan (approved):

PR #950, opened 2018-05-25
THORN-2051: MP metrics: MetricRegistry.getTimers and similar methods don't filter by type

Ladicek:
Motivation
----------
MP Metrics allow to get all registered metrics of given type
using `MetricRegistry.getTimers()` (and similar: `getMeters`,
`getCounters` etc.). However, these methods currently always
return all registered metrics of all types. That's clearly
wrong and actually ultimately results in `ClassCastException`s.

Modifications
-------------
The internal method `MetricsRegistryImpl.getMetrics`,
which is ultimately called in all these cases, already takes
a `MetricType`. It just wasn't used. Modified the loop to,
in addition to checking the filter, also test the metric type.

Result
------
`MetricRegistry.getTimers` and similar methods now work correctly.

-----

### DIFF
--- a/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime/MetricsRegistryImpl.java
+++ b/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime/MetricsRegistryImpl.java
@@ -185,8 +185,8 @@ public class MetricsRegistryImpl extends MetricRegistry {
             LOGGER.infof("Register metric [name: %s, type: %s]", name, type);
             register(metadata, m);
         } else if (!metadataMap.get(name).getTypeRaw().equals(metadata.getTypeRaw())) {
-            throw new IllegalArgumentException("Type of existing previously registered metric " + name + " does not " +
-                                                "match passed type");
+            throw new IllegalArgumentException("Previously registered metric " + name + " is of type "
+                    + metadataMap.get(name).getType() + ", expected " + metadata.getType());
         }
 
         return (T) metricMap.get(name);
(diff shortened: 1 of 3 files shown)


kenfinnigan (dismissed):

fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime/MetricsRegistryImpl.java
```diff
@@ -286,13 +286,15 @@ public void removeMatching(MetricFilter metricFilter) {
...
+        SortedMap<String, T> out = new TreeMap<>();
 
-        Iterator<Map.Entry<String, Metric>> iterator = metricMap.entrySet().iterator();
-        while (iterator.hasNext()) {
-            Map.Entry<String, Metric> entry = iterator.next();
+        for (Map.Entry<String, Metric> entry : metricMap.entrySet()) {
             if (filter.matches(entry.getKey(), entry.getValue())) {
-                out.put(entry.getKey(), (T) entry.getValue());
+                Metadata metadata = metadataMap.get(entry.getKey());
+                assert metadata != null : "missing metadata for registered metric";
```
mkouba: So if metadata are missing and assertions are disabled the next line would just throw NPE?

mkouba:
Is this use case tested by tck? Shouldn't we add our own test?

mkouba:
By the way, there is no thorntail-specific test suite for metrics yet. But I have a pom prepared and I'm going to include this in PR for [THORN-2050](https://issues.jboss.org/browse/THORN-2050).

Ladicek:
I added a test in https://github.com/Ladicek/swarm-test-suite/commit/f7e4e3256f3c803468658de3b1bfc100390e66a5

fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime/MetricsRegistryImpl.java
Ladicek: See https://github.com/Ladicek/wildfly-swarm/blob/5ac9db2cee25adc0a44d51e753e40458692dd2a8/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime/MetricsRegistryImpl.java#L82-L84 and https://github.com/Ladicek/wildfly-swarm/blob/5ac9db2cee25adc0a44d51e753e40458692dd2a8/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime/MetricsRegistryImpl.java#L120-L121

Ladicek:
Rebased and changed order of tests: first test if metric is of expected type, then test if it matches the filter. This is undefined by MP Metrics (filed https://github.com/eclipse/microprofile-metrics/issues/248), but I believe that it makes more sense like this -- if someone calls `MetricRegistry.getTimers(myFilter)`, they can easily assume that `myFilter` will only be passed timers.

Also added a test for the fix.

Ladicek:
WAT?

retest this please

Ladicek:
Windows build failed with the 2 known MP FT TCK failures.

mkouba (approved):

PR #952, opened 2018-05-28
[THORN-1774] Security Constraints configured in YAML behave different than in WEB.XML

McIntozh:
## Motivation
- Issue THORN-1774
- When configuring web security using web.xml it is possible to partially exclude url-patterns to have another (or no) security-role, than another wildcard pattern also matching this url.
- When configuring security via YAML, it is not possible do define a non-security url-pattern

## Modifications
- HttpSecurityPreparer used an empty list as fallback, if no roles where defined in the YAML. I changed it to the web.xml behaviour where no declared roles lead to a "allow all".
-----
- [ x] Have you created a [JIRA](https://issues.jboss.org/browse/THORN) and used it in the commit message?
- [ x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/thorntail/thorntail/pulls) for the same issue?

-----


kenfinnigan:
ok to test

kenfinnigan (approved):

PR #953, opened 2018-05-28
THORN-2053: MP Metrics API - declare dependency on javax.enterprise.api.

mkouba:
Motivation
----------
MP Metrics API module does not declare a dependency on
javax.enterprise.api and so CDI/Interceptors annotations declared on MP
Metrics annotations are ignored.

Modifications
-------------
Added module depednency to MP Metrics API.

Result
------
CDI/Interceptors annotations should not be ignored anymore.


mkouba:
@pilhuhn @Ladicek pls review.

Ladicek (approved):

PR #954, opened 2018-05-28
THORN-2050: Inject base and vendor MetricRegistry.

mkouba:
Motivation
----------
CDI beans for base and vendor MetricRegistry are not available.

Modifications
-------------
Add relevant MetricRegistry producer methods.

Result
------
MP Metrics impl follows the spec requirements again.

### DIFF
--- a/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/deployment/MetricProducer.java
+++ b/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/deployment/MetricProducer.java
@@ -64,6 +64,18 @@ public class MetricProducer {
         return get(MetricRegistry.Type.APPLICATION);
     }
 
+    @RegistryType(type = MetricRegistry.Type.BASE)
+    @Produces
+    MetricRegistry getBaseRegistry() {
+        return get(MetricRegistry.Type.BASE);
+    }
+
+    @RegistryType(type = MetricRegistry.Type.VENDOR)
+    @Produces
+    MetricRegistry getVendorRegistry() {
+        return get(MetricRegistry.Type.VENDOR);
+    }
+
     @Produces
     <T> Gauge<T> getGauge(InjectionPoint ip) {
         // A forwarding Gauge must be returned as the Gauge creation happens when the declaring bean gets instantiated and the corresponding Gauge can be injected before which leads to producing a null value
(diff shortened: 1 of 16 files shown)


mkouba:
@pilhuhn @Ladicek pls review.

fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime/MetricRegistryFactory.java
```diff
@@ -16,52 +16,34 @@
...
  * @author hrupp
  */
 public class MetricRegistryFactory {
 
-    private static final Map<MetricRegistry.Type, MetricRegistry> registries = new HashMap<>();
+    private static final Map<MetricRegistry.Type, MetricRegistry> REGISTRIES = new ConcurrentHashMap<>();
 
-    private MetricRegistryFactory() { /* Singleton */ }
+    private MetricRegistryFactory() {
+        /* Singleton */ }
```
Ladicek: :-)

mkouba: Ah, forgot to remove this comment during refactoring... of course, there is not a single instance at all (and never was).

mkouba: And it's not a factory... so let's rename this class to `MetricRegistries`?

Ladicek (dismissed):

fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime/MetricRegistryFactory.java
Ladicek: Ah, I actually referred to the formatting, but I actually agree with you :-) `MetricRegistries` sounds good.

Ladicek: It could also be `MetricRegistryRegistry` (as in, registry of `MetricRegistry`) :-D

mkouba: Yeah, and we could also register an instance of `MetricRegistryRegistry` to a global registry of registry registries ;-)

pilhuhn:
I guess it is ok, but you know so much more about CDI than I do, that my opinion can be neglected.

mkouba:
@pilhuhn That's ok. Thanks. I've requested a general sanity check, maybe also test review ;-)

Ladicek (approved):

PR #962, opened 2018-05-31
THORN-2041: Fix Teiid OData dependency.

mkouba:
- org.wildfly.swarm:keycloak does not exist anymore

This is a follow-up of https://github.com/thorntail/thorntail/pull/946 that was merged but not rebased (groupIds changed).


sberyozkin (approved):
+1

PR #981, opened 2018-06-05
THORN-2068: MP Metrics - basic support of MP Rest Client proxies.

mkouba:
Motivation
----------
Recently, we added CDI interceptors support to MP Rest Client proxies
(non-portable feature, see also THORN-2030). The primary use case was to
support MP FT annotations. It would be nice to support MP Metrics
annotations as well.

Modifications
-------------
We collect all interfaces annotated with metrics annotations and
register relevant metrics during AfterDeploymentValidation phase. All
metrics interceptors were modified to work with non-contextual instances
such as Java EE components and Rest Client proxies.

Result
------
MP Rest Client proxies can use MP Metrics annotations to register and
update metrics.


mkouba:
@pilhuhn Pls review.

Ladicek (approved):
Nice!

kenfinnigan:
As it's already done it's fine, but for now we need to stop additional MP development until the code is available in SmallRye

mkouba:
@kenfinnigan Sure, I just wanted to catch the 2.0.0.Final train as the migration to SmallRye might take some time ;-)

PR #983, opened 2018-06-06
Enable the v4 core/health EndpointTest

sberyozkin:


sberyozkin:
Looks like Eclipse has reordered the imports a bit, let me know if the original order needs to be preserved

kenfinnigan:
If you can preserve the original order that would be great

core/health/src/test/java/io/thorntail/health/EndpointTest.java
```diff
@@ -48,11 +50,18 @@ public void test() {
...
         });
     }
 
     @Inject
     @Management
     URL url;
+    
+    @Health
+    @Produces
+    public HealthCheck check() {
```
Ladicek, line 64: I'd say it's more idiomatic to annotate the `TestHealthCheck` class with `@Health` instead of creating a producer. But both should work.

kenfinnigan (approved):

PR #989, opened 2018-06-11
THORN-2082: Thorntail Maven plugin uses wrong classifier name for packaging and multistart

antoinesd:
fixing maven plugin to use correct name for classifier and artifacts names during packaging, start and multistart

### DIFF
--- a/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/MultiStartMojo.java
+++ b/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/MultiStartMojo.java
@@ -98,7 +98,7 @@ public class MultiStartMojo extends AbstractSwarmMojo {
         String classifier = process.getChild("classifier").getValue();
         Artifact artifact = findArtifact(groupId, artifactId, classifier);
         if (artifact == null) {
-            artifact = findArtifact(groupId, artifactId, "swarm");
+            artifact = findArtifact(groupId, artifactId, "thorntail");
         }
 
         if (artifact != null) {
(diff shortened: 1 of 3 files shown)


plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
```diff
@@ -51,9 +51,9 @@
 )
 public class PackageMojo extends AbstractSwarmMojo {
 
-    static final String UBERJAR_SUFFIX = "-thorntail";
+    static final String UBERJAR_SUFFIX = "thorntail";
 
-    static final String HOLLOWJAR_SUFFIX = "-hollow" + UBERJAR_SUFFIX;
+    static final String HOLLOWJAR_SUFFIX = "hollow" + UBERJAR_SUFFIX;
```
Ladicek: So previously the hollow JAR suffix was supposed to be `hollow-thorntail`, but that's now `hollowthorntail`. Not sure if that's an issue or not.

mkouba: It's definitely an issue. `myapp-hollow-swarm.jar` should be now `myapp-hollow-thorntail.jar`.

plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/MultiStartMojo.java
```diff
@@ -98,7 +98,7 @@ protected void start(XmlPlexusConfiguration process) throws PluginConfigurationE
         String classifier = process.getChild("classifier").getValue();
         Artifact artifact = findArtifact(groupId, artifactId, classifier);
         if (artifact == null) {
-            artifact = findArtifact(groupId, artifactId, "swarm");
+            artifact = findArtifact(groupId, artifactId, "thorntail");
```
mkouba: We should use the `PackageMojo.UBERJAR_SUFFIX` constant here.

Ladicek (approved):

PR #990, opened 2018-06-11
WIP: KIALI-754 | Allow MP-opentracing get tracer from Jaeger fraction

pavolloffay:
https://issues.jboss.org/browse/KIALI-754

Signed-off-by: Pavol Loffay <ploffay@redhat.com>

### DIFF
--- a/fractions/jaeger/module.conf
+++ b/fractions/jaeger/module.conf
@@ -3,4 +3,3 @@ org.wildfly.swarm.undertow
 org.jboss.logging
 javax.api
 io.jaegertracing
-io.opentracing export=true
--- a/fractions/jaeger/pom.xml
+++ b/fractions/jaeger/pom.xml
@@ -64,19 +64,10 @@
       <groupId>io.thorntail</groupId>
       <artifactId>undertow</artifactId>
     </dependency>
-    <dependency>
-      <groupId>io.thorntail</groupId>
-      <artifactId>opentracing</artifactId>
-    </dependency>
 
     <dependency>
       <groupId>io.jaegertracing</groupId>
       <artifactId>jaeger-core</artifactId>
-      <version>${version.jaeger}</version>
-    </dependency>
-    <dependency>
-      <groupId>io.jaegertracing</groupId>
-      <artifactId>jaeger-tracerresolver</artifactId>
       <version>${version.jaeger}</version>
     </dependency>
 
--- a/fractions/jaeger/src/main/resources/modules/io/jaegertracing/main/module.xml
+++ b/fractions/jaeger/src/main/resources/modules/io/jaegertracing/main/module.xml
@@ -1,13 +1,11 @@
 <module xmlns="urn:jboss:module:1.3" name="io.jaegertracing">
   <resources>
     <artifact name="io.jaegertracing:jaeger-core:${version.jaeger}"/>
-    <artifact name="io.jaegertracing:jaeger-tracerresolver:${version.jaeger}"/>
     <artifact name="io.jaegertracing:jaeger-thrift:${version.jaeger}"/>
   </resources>
 
   <dependencies>
     <module name="org.slf4j"/>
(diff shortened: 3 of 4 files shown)


Ladicek:
So this is a breaking change in how users are supposed to use the `jaeger` fraction, right?

pavolloffay:
It's WIP, we haven't decided yet what is the best. 

For sure we would like to reuse Jaeger fraction for MP-OT fraction to provide thorntail native way for Jaeger configuration. The problem is that jaeger fraction depends on OT fraction which adds servlet filter to trace requests - this would result in double server spans (if we also add MP-OT as a dep).

I can think of two solutions:
* change dependency from OT to mp-ot in jaeger fraction - it's a breaking change if users configured something explicitly (eg. skippattern)
* remove all tracing deps from jaeger fraction - users have to add dependency on tracing fractions explicitly - a breaking change but the Jaeger fraction can be reused for different instrumentations

Ladicek:
Breaking change is fine (well, the `jaeger` and `opentracing` fractions are now in product, so avoiding breaking changes would be nice...), but needs to be well understood and documented.

pavolloffay:
@Ladicek can be `opentracing` fraction deprecated in favor of `microprofile-opentracing`?

kenfinnigan:
Couple of things:

* the build failed because of a checkstyle error
* this needs to be thought about carefully. Will discuss when I’m back from a conference

Ladicek:
I don't know anything about MP OpenTracing, but I'd naively guess that, similarly to `monitor` vs. MP Health, the `opentracing` fraction would eventually be deprecated in favor of the MP counterpart. But please don't take it as a promise really :-)

pavolloffay:
> the build failed because of a checkstyle error

It's just WIP to bootstrap discussion and direction we want to go.

I will try to explain differences between these two fractions: 

`opentracing` - provides tracing at servlet level (only server side)
`microprofile-opentracing` - provides tracing at jax-rs level (server and client), cdi beans tracing (needs to be explicitly enabled).

From the functionality standpoint they are very similar and overlapping. The only feature which is not supported in MP is `skipPatter` where users can specify which URLs are not traced. It will be added to MP-OT in the next version https://github.com/eclipse/microprofile-opentracing/issues/16.

kenfinnigan:
@pavolloffay does `microprofile-opentracing` support tracing at the servlet level like `opentracing` does?

If not, then it's not a like for like change, so we'd need to keep both for now.

In terms of `jaeger`. I think it's probably best to remove the dependency to `opentracing` as opposed to switching to `microprofile-opentracing`. That way `jaeger` can be used with either tracing API.

pavolloffay:
> @pavolloffay does microprofile-opentracing support tracing at the servlet level like opentracing does?

As the previous comment says No - it's on jax-rs level.

>In terms of jaeger. I think it's probably best to remove the dependency to opentracing as opposed to switching to microprofile-opentracing. That way jaeger can be used with either tracing API.

+1 if we don't want to deprecate `opentracing` fraction.

Ladicek:
I'd guess it should be possible to add a vendor-specific extension to our MP OpenTracing fraction to support tracing on the servlet level, no? Or even make it an optional part of the MP spec (for implementations that base their JAX-RS impl on Servlet).

Obviously you want to avoid double traces, but that should be solvable I guess?

kenfinnigan (approved):

pavolloffay:
@kenfinnigan I had some tests prepared locally. I don't think that this current PR works as expected.

I will submit another PR soon. Please do not merge (my) PRs marked as WIP.

kenfinnigan:
bit late for that, will revert

Ladicek:
FTR, I'd like to have a THORN JIRA that would record that a breaking change happened.

Ladicek:
For the record, the follow-up PR is (I believe) https://github.com/thorntail/thorntail/pull/1016

PR #991, opened 2018-06-13
Update v4 to the latest RestEasy version

sberyozkin:
It will bring JAX-RS 2.1 support and is backward compatible with the JAX-RS 2.0, so it will be safe. Users can easily go to the earlier version, but I've never seen a JAX-RS 2.0 application failing on top of the JAX-RS 2.1 implementation :-)

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -50,7 +50,7 @@
     <version.hibernate-search>5.10.1.Final</version.hibernate-search>
     <version.ironjacamar>1.4.7.Final</version.ironjacamar>
     <version.keycloak>3.4.3.Final</version.keycloak>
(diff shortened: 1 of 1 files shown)


sberyozkin:
Hmm... tck/rest-client failed with the 

setupClient(org.eclipse.microprofile.rest.client.tck.DefaultExceptionMapperTest)  Time elapsed: 0.341 sec  <<< FAILURE!
java.lang.IllegalStateException: Multiple RestClientBuilderResolver implementations found: org.jboss.resteasy.client.microprofile.MicroprofileClientBuilderResolver and io.thorntail.restclient.api.RestClientBuilderResolverImpl

PR #992, opened 2018-06-14
[THORN-2059] Workaround to avoid the KC NPE

sberyozkin:


sberyozkin:
I don't know how to test it given this NPE shows only after logging in into the admin console

sberyozkin:
This should be safe because there was only a single query on the forum about using the 'modules' property to load the themes and it did not work. FYI, https://issues.jboss.org/browse/KEYCLOAK-7608, it appears that the team has agreed that some fix may have to be done at the KC level.  
THORN-2075 also uses this workaround and it would be easy to reproduce this issue if THORN-2075 is fixed (may be after 2.0.0.Final ?), I can provide an initial PR without the 'dir' workaround and you can then run a standalone_servers/keycloak and see yourself NPE in the logs :-)

kenfinnigan:
retest this please

kenfinnigan (approved):

PR #997, opened 2018-06-18
[THORN-2084] Update ElytronFraction not to create audit.log by default

sberyozkin:
**Motivation**

Empty audit.log files are created when the Thorntail v2 servers are starting

**Changes**

Remove the code creating the audit.log by default from ElytronFraction

**Result**

Better user experience; audit.log files will be created only if the users need them by configuring the elytron subsystem.

### DIFF
--- a/fractions/wildfly/elytron/src/main/java/org/wildfly/swarm/elytron/ElytronFraction.java
+++ b/fractions/wildfly/elytron/src/main/java/org/wildfly/swarm/elytron/ElytronFraction.java
@@ -19,7 +19,6 @@ import java.util.ArrayList;
 import java.util.HashMap;
 
 import org.wildfly.swarm.config.Elytron;
-import org.wildfly.swarm.config.elytron.Format;
 import org.wildfly.swarm.config.elytron.SimplePermissionMapper;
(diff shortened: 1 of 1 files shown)


fractions/wildfly/elytron/src/main/java/org/wildfly/swarm/elytron/ElytronFraction.java
```diff
@@ -80,11 +79,6 @@ public ElytronFraction applyDefaults() {
             providers.provider(ELYTRON);
             providers.provider(OPENSSL);
         });
-        fileAuditLog(LOCAL_AUDIT, (log) -> {
-            log.path("audit.log");
-            log.relativeTo("jboss.server.log.dir");
-            log.format(Format.JSON);
-        });
```
Ladicek, line 87: In place of these 4 removed lines, could you add a comment with a link to the JIRA? That way, when someone in the future compares the default Thorntail and WildFly configuration, they will know this is intentional.

Ladicek (approved):

Ladicek:
retest this please

kenfinnigan (approved):

PR #1013, opened 2018-06-22
Introduce CDI-based Vertx integration

mkouba:
TL;DR: Renames the original vertx integration to `thorntail-vertx-jca`. Introduces `thorntail-vertx` built on CDI and `thorntail-vertx-web` which could be an alternative to the undertow/servlet module.

Most of the ideas are taken from the https://github.com/weld/weld-vertx project.

---

Features:
* `thorntail-vertx`
  * Starts `Vertx` instance, unlike `thorntail-vertx-jca` leverages the built-in config mechanisms which allow you to fine-tune the `VertxOptions` used (see also `VertxInitializer`)
  * Registers `Vertx` CDI bean
  * CDI observers `@VertxConsume("address") VertxEvent` are notified when a message is sent via vertx `EventBus`
  * Makes it possible to propagate CDI events to vertx `EventBus` - see also `@VertxPublish` and `@VertxSend`
  * Discovers and automatically deploys all `Verticle`s that are registered as beans
* `thorntail-vertx-web`
  * Starts `HttpServer`, leverages the built-in config mechanisms which allow you to fine-tune the `HttpServerOptions` used (see also `HttpServerInitializer`)
  * Makes it possible to programatically add route handlers - see also `io.thorntail.testsuite.vertx.web.NextHandlerRegistrar` in test
  * Makes it possible to add vertx `Route` handler in a declarative way through `@WebRoute` annotation; like http://docs.jboss.org/weld/weld-vertx/latest/#_declaring_routes_in_a_declarative_way

The only missing parts are probably **opentracing** - I'm not an expert in this area so I need to explore `io.thorntail.tracing` a little bit, and **docs**.

There are few more ideas/functionalities we could port from weld-vertx but the fundamental features are implemented.

I did not remove the `thorntail-vertx-jca` module entirely but I believe it does not provide any feature which is not implemented in the new modules (except for tracing of course ;-).


bom/pom.xml
```diff
@@ -226,6 +226,21 @@
...
+        <version>${version.vertx}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.vertx</groupId>
+        <artifactId>vertx-web</artifactId>
+        <version>${version.vertx}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.vertx</groupId>
+        <artifactId>vertx-hazelcast</artifactId>
```
Ladicek, line 241: Not trying to push anything, but I'd expect we'd utilize the Infinispan-based cluster manager :-)

mkouba: This one is only used for tests...

core/vertx-web/src/main/java/io/thorntail/vertx/web/WebRoute.java
```diff
@@ -0,0 +1,140 @@
...
+@Retention(RUNTIME)
+@Target({ TYPE, METHOD })
+@Stereotype
+public @interface WebRoute {
+
+    /**
+     *
+     * @see Router#route(String)
+     */
+    String value() default "";
```
Ladicek: In case other parameters are often present (not sure, but looks like that's the case), would it be better to have a more descriptive name instead of the generic `value`?

mkouba: Makes sense. `@WebRoute("/path")` would probably never be used in real apps.

bobmcwhirter:
Ohh this is nice. Lemme close my PR.

bobmcwhirter (approved):
This looks f'awesome!

PR #1017, opened 2018-06-22
Produce a Vert.x DnsClient.

bobmcwhirter:

-----


core/vertx/pom.xml
```diff
@@ -29,6 +29,11 @@
       <artifactId>junit</artifactId>
     </dependency>
 
+    <dependency>
```
kenfinnigan: test scope?

bobmcwhirter: Already is via dep-man:


      <dependency>
        <artifactId>thorntail-testing</artifactId>
        <version>4.0.0-SNAPSHOT</version>
        <scope>test</scope>
      </dependency>

kenfinnigan:
Not sure how, but OpenTracing test is failing

bobmcwhirter:
Yah, saw that.

How do we cause CI to retrigger for v4?

kenfinnigan:
I clicked rebuild, don't think there's a way through comments

core/vertx/pom.xml
kenfinnigan: ignore me

kenfinnigan:
not sure why we had the CI hiccup, hopefully not a race condition

core/vertx/src/main/java/io/thorntail/vertx/DnsClientProducer.java
```diff
@@ -0,0 +1,24 @@
...
+import io.vertx.core.dns.DnsClient;
+
+/**
+ * Created by bob on 6/22/18.
+ */
+@ApplicationScoped
+public class DnsClientProducer {
+
+    @Produces
+    @ApplicationScoped
```
mkouba: If you produce an `@ApplicationScoped` bean it usually does not make sense to make the producer `@ApplicationScoped` unless it holds some state. In this case, I would use `@Dependent DnsClientProducer` - it will be created to produce a `DnsClient` and destroyed afterwards - and you can even inject `Vertx` directly in the producer method.

core/vertx/src/test/java/io/thorntail/vertx/VertxTest.java
```diff
@@ -0,0 +1,25 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Created by bob on 6/22/18.
+ */
+@RunWith(ThorntailTestRunner.class)
+public class VertxTest {
```
mkouba: We already test `Vertx` injection here:
https://github.com/thorntail/thorntail/blob/4.x/testsuite/simple/vertx/src/test/java/io/thorntail/testsuite/vertx/Emitter.java#L16 
and here:
https://github.com/thorntail/thorntail/blob/4.x/testsuite/simple/vertx/src/test/java/io/thorntail/testsuite/vertx/VertxInitializerTest.java#L49

core/vertx/pom.xml
mkouba: Hm, I wonder where to place the tests using `ThorntailTestRunner`? There is a dedicated test suite here:
https://github.com/thorntail/thorntail/tree/4.x/testsuite/simple/vertx

I think that `src/test/java` should only contain unit tests.

mkouba (dismissed):

bobmcwhirter:
You’re correct in all cases. I’m seriously out of practice.

kenfinnigan:
@bobmcwhirter when you have a chance, can you rebase this and update based on @mkouba comments?

kenfinnigan (approved):

PR #1020, opened 2018-06-25
TracingInterceptor - minor perf. improvements.

### LINKED ISSUE #1018: TracingInterceptor performance improvements
Right now, every invocation of a traced component includes:
* iteration over `Instance<SpanHandler>` and thus creation of new bean instances for `@Dependent SpanHandler` implementations
* construction and sorting of a new list (see `TracingInterceptor.sortedHandlers()`)

Proposal:
* cache a sorted list of `SpanHandler`s if only `@ApplicationScoped` or  `@Dependent` handlers are found; initialize the list during interceptor instance creation

mkouba:
- resolves #1018

### DIFF
--- a/core/kernel/src/main/java/io/thorntail/tracing/impl/TracingInterceptor.java
+++ b/core/kernel/src/main/java/io/thorntail/tracing/impl/TracingInterceptor.java
@@ -2,34 +2,30 @@ package io.thorntail.tracing.impl;
 
 import java.lang.reflect.Method;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import javax.annotation.Priority;
 import javax.enterprise.inject.Any;
-import javax.enterprise.inject.Instance;
 import javax.inject.Inject;
 import javax.interceptor.AroundInvoke;
 import javax.interceptor.Interceptor;
 import javax.interceptor.InvocationContext;
 
-import io.opentracing.ActiveSpan;
-import io.opentracing.Tracer;
 import org.eclipse.microprofile.opentracing.Traced;
+import org.jboss.weld.inject.WeldInstance;
+
+import io.opentracing.ActiveSpan;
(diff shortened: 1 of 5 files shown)


bobmcwhirter (dismissed):
LGTM. Why WeldInstance instead of Instance?

mkouba:
In fact, I think we can always cache the contextual references. It should work for all built-in scopes and for all custom normal scopes too.

@bobmcwhirter I used `WeldInstance` to:
* test scopes of injected `SpanHandler` instances (I'll remove this - see above)
* sort the beans correctly; `HighToLowPriorityComparator` would not work correctly for custom beans and beans created from modified `AnnotatedType`... corner cases I know, but it's probably more convenient than to maintaining stuff like `io.thorntail.util.Annotations`

core/kernel/src/main/java/io/thorntail/tracing/impl/TracingInterceptor.java
```diff
@@ -82,10 +75,23 @@ protected boolean isTraced(Object target, Method method) {
         return classTraced.value();
     }
 
+    private List<SpanHandler> spanHandlers() {
+        List<SpanHandler> handlers = spanHandlers;
```
Ladicek, line 79: Why? Is this some smart way to make the number of volatile loads/stores lower?

mkouba: Yes, the local copy ensures we only do 1 volatile read if the cached value is already initialized. It might look like a premature optimization here... but I think it's ok ;-)

Ladicek: Ah I see, the `return` statement would have to do another volatile read. I see, thanks.

kenfinnigan (approved):

PR #1023, opened 2018-06-26
use SmallRye Metrics fixes #968

### LINKED ISSUE #968: SmallRye Metrics integration

michalszynkiewicz:
- [X] [v4] Have you created a [GitHub Issue](https://github.com/thorntail/thorntail/issues) and used it in the commit message?

-----


core/metrics/src/main/resources/io/smallrye/metrics/mapping.yml
```diff
@@ -0,0 +1,110 @@
+base:
```
kenfinnigan: What's the mapping file for, given it wasn't present before?

michalszynkiewicz: It's defining the server specific metrics. V2 used mapping.yml file for it and v4 used properties files. I pulled out the class that parses the yaml config into [SmallRye](https://github.com/smallrye/smallrye-metrics/pull/1/files#diff-c04dd7f3444f13aa6ceeeed1140a63aaR45) and made v4 use it.

bobmcwhirter:
If we could do properties we could avoid the extra dependency on snakeyaml. That’s why v4 moved away from yaml.

kenfinnigan:
Was just about to mention that, beat me to it.

I think SmallRye should also not automatically do YAML, but if YAML is present then do it. If that makes sense

michalszynkiewicz:
okay, I'll try to switch SmallRye to properties.

michalszynkiewicz:
@bobmcwhirter @kenfinnigan could we use a single property file, like https://gist.github.com/michalszynkiewicz/844b9da1f67fb57696ca391ac306e835 for the mapping? It's easier to get hold of it than file per metric

bobmcwhirter:
Sure I just wanted to avoid yet another dependency. This is a-okay.

michalszynkiewicz:
restest this please

kenfinnigan (approved):

PR #1024, opened 2018-06-26
Vertx CDI integration - implement opentracing support

### LINKED ISSUE #1019: Vertx CDI module - implement opentracing support
- implement `SpanHandler` that handles `VertxMessage` observers and `WebRoute` handlers/observers
- if needed, register `EventBus` interceptor; see `io.vertx.core.eventbus.EventBus.addInterceptor(Handler<SendContext>)`

mkouba:
- resolves #1019


mkouba:
@pavolloffay @bobmcwhirter I'm no tracing expert so I would appreciate any review/feedback.

kenfinnigan (approved):
Looks good to me. @bobmcwhirter any thoughts?

mkouba:
Thanks for the review Ken!

core/vertx-web/src/main/java/io/thorntail/vertx/web/tracing/WebRouteSpanHandler.java
```diff
@@ -0,0 +1,68 @@
...
+            return null;
+        }
+        VertxWebLogger.LOG.startSpanForWebRoute(ctx.getMethod());
+        HttpMethod httpMethod = routingContext.request().method();
+        SpanContext parent = tracer.extract(Format.Builtin.HTTP_HEADERS, new RoutingContextTextMap(routingContext));
+        SpanBuilder builder = tracer.buildSpan(String.format("%s:%s.%s", httpMethod, ctx.getMethod().getDeclaringClass().getName(), ctx.getMethod().getName()));
+        builder.withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER);
+        builder.withTag(Tags.HTTP_METHOD.getKey(), httpMethod.toString());
+        builder.withTag(Tags.HTTP_URL.getKey(), routingContext.request().absoluteURI());
+        if (parent != null) {
```
pavolloffay, line 62: nit: null check is not necessary

core/vertx/src/main/java/io/thorntail/vertx/tracing/VertxMessageSpanHandler.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+        VertxLogger.LOG.startSpanForVertxMessageObserver(ctx.getMethod());
+        SpanContext parent = tracer.extract(Format.Builtin.HTTP_HEADERS, new VertxMessageTextMap(vertxMessage));
+        SpanBuilder builder = tracer.buildSpan(String.format("%s:%s.%s", OP_NAME_PREFIX, ctx.getMethod().getDeclaringClass().getName(), ctx.getMethod().getName()));
+        builder.withTag(Tags.MESSAGE_BUS_DESTINATION.getKey(), vertxMessage.address());
+        builder.withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_CONSUMER);
+        if (parent != null) {
+            builder.asChildOf(parent);
+        }
+        return builder.startActive();
```
pavolloffay, line 61: I am not sure whether this is correct. (also related to https://github.com/thorntail/thorntail/pull/1024/files#diff-d31eddb96f67822dfa6f43f5f08c4a01R37) 

I am not sure what the vert.x integration in swarm is, but in vertx you cannot rely on active span because it's stored in thread locals.

core/vertx/src/main/java/io/thorntail/vertx/tracing/VertxEventBusInterceptor.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Inject
+    private Tracer tracer;
+
+    @Inject
+    private VertxInitializer vertxInitializer;
+
+    @Override
+    public void handle(SendContext sendContext) {
+        ActiveSpan parent = tracer.activeSpan();
```
pavolloffay, line 37: Again if this is running on event loop then getting span from thread locals would not work

pavolloffay (commented):
Setting parent span in event bus would not work. You should avoid using the active span altogether in vertx.

mkouba:
@pavolloffay Thanks for review! I was just trying to copy the logic from the original JCA integration. But you're right that thread locals wouldn't work on the event loop thread so in our case only blocking handlers would work (dispatched to working threads can leverage active spans). Is there a different technique we can use for non-blocking handlers?

pavolloffay:
Future version of vert.x  should have instrumentation API to allow these kinds of things https://github.com/vert-x3/wiki/wiki/RFC:-continuation-instrumentation

core/vertx-web/src/main/java/io/thorntail/vertx/web/tracing/WebRouteSpanHandler.java
mkouba: Hm, does `io.opentracing.Tracer.extract()` throw an `IllegalArgumentException` if no `SpanContext` is propagated? I think it shouldn't...

pavolloffay: `io.opentracing.Tracer`  is an interface :) 

The impl should not thow any exception, definitely if the context is missing or even it's corrupted

pavolloffay: please report a bug on the concrete tracer impl if something like that happens

mkouba: Well, I was talking about your null check note... but now I can see the `SpanBuilder.asChildOf()` javadoc: `If parent==null, this is a noop`...

And of course, the `Tracer` is an interface but we're programming to the interfaces so we don't care about the impls ;-)

core/vertx/src/main/java/io/thorntail/vertx/tracing/VertxMessageSpanHandler.java
mkouba: I think this should work even if using the event loop thread  because `VertxMessageSpanHandler` is only used to start an active span for a CDI observer notification (span is deactivated once the invocation completes).

PR #1009, opened 2018-06-21
Simple keycloak test v4

sberyozkin:
KC adapter hangs waiting if no KC server is available and throws NPE if it finds no deployment so the test checks it is 500. (I'm following on it on the KC users list)

Resolves #1008

PR #1029, opened 2018-06-28
Replace microprofile-config impl with smallrye-config.

bobmcwhirter:
- Adjust for lack of Optional<Collection<T>> in mp-config.
- Provide ConfigSourceProviders for extra functionality.

Fixes https://github.com/thorntail/thorntail/issues/967

- [x] [v4] Have you created a [GitHub Issue](https://github.com/thorntail/thorntail/issues) and used it in the commit message?

-----


bom/pom.xml
```diff
@@ -90,7 +93,7 @@
     <version.commons-logging>1.2</version.commons-logging>
     <version.netty>4.1.24.Final</version.netty>
 
-    <version.mp-config>1.2.1</version.mp-config>
+    <version.mp-config>1.3</version.mp-config>
```
kenfinnigan, line 96: Just wondering, do we need the MP Config API in the build as well as SmallRye Config?

bobmcwhirter: Nope, in fact we don't. I removed the <dependency> but missed the property, I think.

sberyozkin:
the fault-tolerance test is failing

bobmcwhirter:
Hrm, it passes locally. I'll double-check.

sberyozkin:
Hi Bob, not sure if it is related but while I was looking at the keycloak test I ended up lowering the the event priority in one of the customizers to ensure the ordering is predictable, see #1009

sberyozkin:
This branch has few conflicts now

bobmcwhirter:
I’ll rebase

bobmcwhirter:
Let's see how that rebase goes. Passing locally now.

bom/pom.xml
bobmcwhirter: We do need the `version.mp-config` to reference the TCK, and in that case, I think keeping the TCK and API in sync within the context of Thorntail might be smart?

kenfinnigan: is there a way to ensure one doesn't get updated without the other?

trying to minimize moving pieces

bobmcwhirter: I think at this point, not easily. We can keep our API + TCK in sync, but then SmallRye might vary.  We can keep SmallRye + API in sync, but then TCK might vary.

sberyozkin (approved):
Ready to go, though the 4.x line is currently broken

kenfinnigan (approved):

PR #1032, opened 2018-07-03
[THORN-2062] Migrate MP Health implementation to SmallRye

sberyozkin:


sberyozkin:
I'll release smallrye-health once the PR is reviewed

sberyozkin:
For the moment I've kept the code which reflectively discovers health annotations on JAX-RS resources in Thorntail v2 given that I have found no specific refs to the algorithm in the spec text so thought it was a bit early to put it in the smallrye code

kenfinnigan:
You mean scanning through JARs within the WAR?

I deliberately removed it for v4 as we presumed "all in one"

sberyozkin:
Sorry, I meant this code:
https://github.com/thorntail/thorntail/blob/master/fractions/microprofile/microprofile-health/src/main/java/org/wildfly/swarm/microprofile/health/runtime/HealthAnnotationProcessor.java#L72

My understanding is that adding an MP Health annotation to the custom JAX-RS class (Application, or resource class at the class or method level) should instruct the Health endpoint delegate to such custom resources but I'm not sure yet this code should be migrated to SmallRye just yet

kenfinnigan:
It probably should as I believe it's common for `@Health` to be added to JAX-RS Resources

fractions/microprofile/microprofile-health/src/main/resources/modules/org/eclipse/microprofile/health/main/module.xml
```diff
@@ -2,7 +2,8 @@
 <module xmlns="urn:jboss:module:1.3" name="org.eclipse.microprofile.health">
     <resources>
         <artifact name="org.eclipse.microprofile.health:microprofile-health-api:${version.microprofile-health}"/>
-        <artifact name="io.thorntail:health-api:${project.version}"/>
+        <artifact name="io.smallrye:smallrye-health:${version.smallrye.health}"/>
+        <artifact name="org.glassfish:javax.json:${version.org.glassfish.javax.json}"/>
```
michalszynkiewicz, line 6: do you know what's the difference between using it as an artifact and as a module?
MP JWT depends on it this way: https://github.com/michalszynkiewicz/wildfly-swarm/blob/a61c6db0d897fd38a0f1f70e5c9e1be8f4844b32/fractions/microprofile/microprofile-jwt/module.conf#L15

sberyozkin:
@kenfinnigan, sure, I'll try to get this code also moved into small rye, ideally into the  SmallRyeHealthReporter which will return a list of custom URIs if such are found

sberyozkin:
@kenfinnigan Hi Ken, I've been looking at moving some of the code from

https://github.com/thorntail/thorntail/blob/master/fractions/microprofile/microprofile-health/src/main/java/org/wildfly/swarm/microprofile/health/runtime/HealthAnnotationProcessor.java#L72

to smallrye health and I'm starting hesitating again as it seems it may be a bit early for it.

For example, Thorntail v4 servlet is listening on "/health". If a user provides a JAX-RS application/service with some method returning some Health payload then it will go via a different path and Thorntail v4 servlet won't be even aware of it and whatever the introspection SmallRye will do won't be of use. 

So we can register the v4 servlet on the wildcard URL pattern  and it will pass the JAX-RS requests through itself via RequestDispatcherProvider but it can become brittle. May be it should become a servlet filter instead. But again I'm not sure what it will add if JAX-RS resources can simply handle it directly.

I'm also looking at Thorntail v2 code and I'm really not sure that the code which checks these annotations has any effect right now(except that the SecurityHttpContexts code ensures the security is applied to such custom routes).  The Undertow HttpHandler named HttpContexts in v2 only handles "/health" queries and only if HealthChecks are available. 

Right now it seems to me that the only value that having @Health on the custom JAX-RS resource brings to Undertow HttpHandler is that it can actually decide that it should not block "/health" requests if a handler is available further down the line (I'll open a minor enhancement request to Thorntail v2).

Should we give this particular refactoring some pause and revisit once we sort out the issue of supporting multiple HTTP providers in some pluggable or modular way (servlets, JAX-RS, etc) ?

kenfinnigan:
If it's currently specific to v2, then let's not move it to SmallRye right now

bobmcwhirter:
Mayhap migrate from v4 to SmallRye. I think the v4 variant is de-coupled a bit.

sberyozkin:
Sure, though I'll open an issue for the v2 handler to actually use the info about the custom handlers :-). In general it looks like it is useful exactly in such cases, where we have a pass through filter with say listens on a "/health" path, and a web context root is "/", so it needs some hint that some other handler listens further in the chain

sberyozkin:
Hi Bob, 
The v4 servlet should probably be fine as it listens on a specific path. So if both this servlet and a custom JAX-RS resource with @health  are on the same primary port then it will not be possible for both to have the same "/health" root so I can imagine in such cases the users won't even add the health servlet or will use different paths for JAX-RS handlers. Only if a default health handler is a filter or a wildcard servlet then it makes sense for it to to be aware of the other health handlers to avoid blocking such handlers.
may be later on the v4 servlet can become a servlet filter...

bobmcwhirter:
I guess I'm confused. I thought @Health-related things were not specific to Servlet or JAX-RS resources, but just happen to be served up on /heatlh as a magic end-point, akin to metrics-like things.

I think all bets are off if someone binds a servlet or jax-rs resource specifically to `/health`, else it's a magical/implicit URL.  No?

sberyozkin:
May be I'm more confused :-). My understanding is, that @Health supports the injection of the custom HealthChecks. 
But Thorntail v2 also checks if JAX-RS resources have custom handlers qualified with Health available - though it does not do much at the moment (see https://issues.jboss.org/browse/THORN-2102).

kenfinnigan:
Does v2 do that because health checks could be JAX-RS resources, and that's the only way they're discovered?

If so, not sure that's something we want to push further upstream

bobmcwhirter:
Swarm's first impl of Health was non-stanard, prior to the existence of mp-health, so that might be part of the confusion.  My reading of mp-health doesn't seem terribly bound to HTTP other than saying there is a `/heath` endpoint with a particular body.

https://microprofile.io/project/eclipse/microprofile-health

sberyozkin:
No, it is an extra bit, and yes it is probably specific to Swarm :-)
Here it checks these methods:
https://github.com/thorntail/thorntail/blob/master/fractions/microprofile/microprofile-health/src/main/java/org/wildfly/swarm/microprofile/health/runtime/HealthAnnotationProcessor.java#L72

and here it checks HealthChecks
https://github.com/thorntail/thorntail/blob/master/fractions/microprofile/microprofile-health/src/main/java/org/wildfly/swarm/microprofile/health/deployment/HealthExtension.java

So I'm fine with closing THORN-2102 :-).
If otherwise this and v4 PRs are OK then I'll release smallrye-health next week.
thanks

Ladicek:
IIUC, @bobmcwhirter is exactly right here -- the MP Health implementation seems to contain bits and pieces that are totally irrelevant to MP Health and just serves the need of the Swarm custom `monitor` fraction, which has been deprecated a few months ago.

sberyozkin:
retest this please

sberyozkin:
The usual Win MP FT test failures

kenfinnigan (approved):

PR #1033, opened 2018-07-03
Migrate to SmallRye Health Implementation for 4.x

sberyozkin:


sberyozkin:
I'll release smallrye-health once the PR is reviewed

core/health/src/main/java/io/thorntail/health/impl/HealthServlet.java
```diff
@@ -1,110 +1,9 @@
...
-            builder.add("data", data.build());
-        });
-
-        return builder.build();
-    }
-
-    @Inject
-    @Health
-    private Instance<HealthCheck> checks;
+public class HealthServlet extends io.smallrye.health.SmallRyeHealthServlet {
```
michalszynkiewicz: Why do we need this class?

sberyozkin: You are right, right now there's no obvious need for it. I thought it would have an option to have some Thorntail specific in Thorntail own servlet if needed. The other thing is that it is not clear yet if SmallRye Health should have its own servlet, it has at the moment but I know there are concerns it won't fit into Undertow HttpHandler...

michalszynkiewicz: yeah, I had trouble making it work with servlet in metrics, I ended up doing a implementation-agnostic request handler in SmallRye and then creating a servlet in v4 and undertow handler in v2.
Either way, I think it would be good to remove this class now if it's not needed. It will be easy to add it back in future if needed.

sberyozkin: Sure, have just removed the servlet, thanks

sberyozkin:
Michal, the servlet is back since the smallrye health is no longer providing it

sberyozkin:
retest this please

sberyozkin:
@kenfinnigan Ken, this should be ready to go, have a look please

kenfinnigan (approved):

PR #1035, opened 2018-07-04
Improvements to the Keycloak Config Resolution process

### LINKED ISSUE #949: Keycloak module should also have a multitenancy resolver
It will use a request path to pick up the request path specific deployment from the map

sberyozkin:
**Motivation**
- Make the resolution of KeycloakDeployments instant or nearly instant at the request time by avoiding searching for and parsing them at every request.
- Complement the original CDI-driven injection of custom resolvers by supporting "keycloak.json.path" which can point to the classpath or file-based KC adapter config file not necessarily named 'keycloak.json' and "keycloak.multitenancy.paths" offering a simple KC multitenancy support. Important point is that these properties (esp "keycloak.json.path") do not exclude the use of the custom resolvers but offer a simpler way to customize the classpath/filesystem resolution. Both of these properties will also offer the continuity as far as the configuration of KC adapter in v2 and v4 is concerned
  
**Modifications**

Added these two properties, removed the default MetaInf resolver, added Static and Multitenancy resolvers, updated Classpath and FileSystem resolvers to reuse the code which is also used during the setup of Static and Multitenancy ones. This is already partially tested by the simple/keycloak test but will I add more tests as part of the docker KC test (wip)

**Result**
Faster resolution of KC deployments at runtime, possibility to set up the resolvers by using these two properties as well.

Resolves #949


core/security-keycloak/src/main/java/io/thorntail/security/keycloak/impl/KeycloakUtils.java
```diff
@@ -0,0 +1,61 @@
...
+        if (path == null) {
+            return null;
+        }
+        if (path.startsWith("classpath:")) {
+            path = path.substring("classpath:".length());
+        }
+        URL url = KeycloakUtils.class.getClassLoader().getResource(path);
+        if (url == null) {
+            url = KeycloakUtils.class.getClassLoader().getResource("META-INF/" + path);
+        }
```
kenfinnigan, line 29: I don't believe we need WEB-INF parsing as there is no "WAR" in v4. It's all just JARs

sberyozkin: Sure, will fix now.

bobmcwhirter (approved):

PR #1042, opened 2018-07-09
THORN-1955: switched MP JWT web.xml sec. constraints to JAXRS filters

michalszynkiewicz:

-----
Motivation
----------
Currently Thorntail v2 maps security annotations to web xml security constraints.
This makes it impossible to have different security constraints for e.g. methods that share the same paths and http methods but produce different content types.

See https://issues.jboss.org/projects/THORN/issues/THORN-1955 and https://issues.jboss.org/projects/THORN/issues/THORN-1972

Modifications
-------------
Instead of creating web.xml security constraints, MP JWT implementation now uses JAX-RS filters.

Also, tests based on the aforementioned bug reports have been added.

Result
------
The user is able to use more fine-grained security constraints.


fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MpJwtFilterRegistrator.java
```diff
@@ -0,0 +1,99 @@
...
+                .collect(Collectors.toList());
+        switch (annotations.size()) {
+            case 0:
+                return null;
+            case 1:
+                return annotations.iterator().next();
+            default:
+                // mstodo: what kind of exception should we throw here?
+                // mstodo check if annotation names are printed properly or need to be mapped to class names
+                throw new RuntimeException("Duplicate MicroProfile JWT annotations found on "
```
michalszynkiewicz: Is there a more suitable exception that we could throw here?

Ladicek:
Nice work! Just wondering if the annotations can be inherited. JAX-RS annotations can, but obviously these are not JAX-RS annotations. The Common Annotations spec doesn't mention it, the MP JWT spec doesn't either. We probably didn't support that in the past, so we don't regress -- I'm just being curious here.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
```diff
@@ -17,37 +17,28 @@
...
-import org.wildfly.swarm.undertow.descriptors.SecurityConstraint;
 import org.wildfly.swarm.undertow.descriptors.WebXmlAsset;
 
+import javax.inject.Inject;
+import java.io.File;
+import java.util.Collection;
+
 /**
  * A DeploymentProcessor implementation for the MP-JWT custom authentication mechanism that adds support
  * for that mechanism to any war the declares a login-config/auth-method = MP-JWT.
```
Ladicek, line 44: The lines below mention that this class maps security annotations to `web.xml` security constraints, which is no longer true.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
```diff
@@ -174,192 +141,9 @@ public void process() throws Exception {
...
-        httpMethods.add(DELETE);
-        httpMethods.add(HEAD);
-        httpMethods.add(OPTIONS);
-        for (AnnotationInstance customHttpMethod : index.getAnnotations(HTTP_METHOD)) {
-            httpMethods.add(customHttpMethod.name());
-        }
-        return httpMethods;
+        JAXRSArchive jaxrsArchive = archive.as(JAXRSArchive.class);
+        jaxrsArchive.findWebXmlAsset()
+                .setContextParam("resteasy.providers", MpJwtFilterRegistrator.class.getName());
```
Ladicek: Is this required, if the class is annotated `@Provider`?

Ladicek (dismissed):
Just a couple minor questions. I'm no expert on the JAX-RS SPI, but the code looks "obviously correct" to me! :-)

michalszynkiewicz:
About the inheritance, IIUC the section 2.1 of the JSR 250 (that was mentioned in the MP issue regarding these annotations), says that the annotations should not be inherited. I didn't find anything that would say otherwise.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
michalszynkiewicz: It is not working for me without it, although the javadoc says it should: https://docs.oracle.com/javaee/7/api/javax/ws/rs/container/DynamicFeature.html

Pavol did the same for opentracing:
https://github.com/michalszynkiewicz/wildfly-swarm/blob/10ff026f7cec05bb938ac9c066b21191e9770a1e/fractions/microprofile/microprofile-opentracing/src/main/java/org/wildfly/swarm/mpopentracing/runtime/TracingInstaller.java#L50-L51

https://github.com/michalszynkiewicz/wildfly-swarm/blob/10ff026f7cec05bb938ac9c066b21191e9770a1e/fractions/microprofile/microprofile-opentracing/src/main/java/org/wildfly/swarm/mpopentracing/deployment/MicroprofileServerTracingFeature.java#L33


I think `@Provider` can be removed in this case, I'll test it and update the PR if it can.

Ladicek:
Good point re section 2.1, I was wrong and you're right -- JSR 250 actually _does_ define how annotations interact with inheritance. Thanks!

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
Ladicek: Ah I see. I think this is fine, was just curious!

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
michalszynkiewicz: fixed

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
michalszynkiewicz: it worked without `@Provider` so I removed it.

Ladicek: Ah maybe one other thing here, what happens if the `web.xml` from deployment already contains the `resteasy.providers` context param, with a user-provided value?

michalszynkiewicz: I didn't think about it, I'll check it after PTO but I guess I will have to merge it with what user provided.

dvddhln:
Hi, I have been testing locally with this PR and looks much better with our parametrized methods for our resources. Nice work :+1:

sberyozkin:
Might make sense to open a RestEasy issue if @Provider annotated DynamicFeature is not picked up - though it might've been fixed in the later versions.
By the way RestEasy itself may also be shipping RBAC filters (I thought I saw some blog/docs awhile back) - it makes sense to do it at the JAX-RS level as well - as it will work for all JAX-RS apps irrespectively of how the SecurityContext has been populated.
One thing which also might make sense to confirm - does the security constraints allow for the 'all' match ? I.e, if the security constraints are for both 'A' and 'B' roles - is it possible to enforce that the user is in both of these roles ? if yes then the roles filter should also support this option.
Supporting this at the JAX-RS level is interesting - should we then consider updating other fractions like openapi, health, etc to the JAX-RS (This is prob a separate duscussion :-) )

Ladicek:
> does the security constraints allow for the 'all' match ? I.e, if the security constraints are for both 'A' and 'B' roles - is it possible to enforce that the user is in both of these roles

Is it even possible to express that using the common security annotations (`@PermitAll`, `@DenyAll`, `@RolesAllowed`)?

> should we then consider updating other fractions like openapi, health, etc to the JAX-RS (This is prob a separate duscussion :-) )

Definitely a separate discussion IMHO :-), because MP JWT specifies that the security annotations can be used on JAX-RS resources, while the other specs don't say anything about JAX-RS.

sberyozkin:
Right, I checked a bit and looks like it is not possible and the general advise is to create a composite role if the all inclusive is needed.

Re RestEasy, this seems relevant:
https://github.com/resteasy/Resteasy/blob/master/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/interceptors/RoleBasedSecurityFeature.java

Might make sense to see if it can be reused though given the simplicity of the roles filter it might be simpler to have an MP-JWT own impl if it is easier to plugin it into the MP JWT code :-)

Ladicek:
Looks like the RESTEasy implementation is exactly what we need here, except I like how this PR distinguishes 401 and 403. The RESTEasy filter always returns 403.

Maybe we should also add `@Priority(Priorities.AUTHORIZATION)`, like RESTEasy does in https://github.com/resteasy/Resteasy/blob/master/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/interceptors/RoleBasedSecurityFilter.java#L18?

sberyozkin:
Yes, 401 if no user principal is available is better :-).  Good point, adding the priority can minimize a chance of some custom JAX-RS filters getting prioritized in front and doing some side-effects before our filter gets a chance to enforce the roles...

michalszynkiewicz:
the distinction between 401 and 403 is required by the spec so if there's no easy way to make RESTEasy work this way we'd have to stay on our filters. I will take a look.

Thanks for your reviews!

bobmcwhirter:
I filed a (disputed) bug a few months ago and TTv4 has a fix for this
issue.

bobmcwhirter:
https://issues.jboss.org/plugins/servlet/mobile#issue/RESTEASY-1850

michalszynkiewicz:
@bobmcwhirter thanks, I will check it out.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
michalszynkiewicz: done

michalszynkiewicz:
This change is now rebased on top of SmallRye changes and it is ready for review

Ladicek (dismissed):

Ladicek:
I don't recall seeing the `ManagementConsoleTest` failing (even on Windows). Anyone?

michalszynkiewicz:
@Ladicek I didn't see it failing before either. I found an unnecessary entry in the module.conf so had to add another commit to the PR, let's see if the test fails again.

Ladicek (approved):

PR #1043, opened 2018-07-10
[THORN-2104] Fixing the product build

sberyozkin:


sberyozkin:
Michal, FYI, the openapi fraction was failing with what I suspect the duplicate dependency management info, so I just removed the eclipse dep from the fraction pom as it is also coming via the smallrye pom

fractions/microprofile/microprofile-openapi/pom.xml
```diff
@@ -64,12 +64,6 @@
       <artifactId>microprofile-config</artifactId>
     </dependency>
 
-    <!-- OpenAPI Spec -->
-    <dependency>
```
michalszynkiewicz: did this thing cause problems for the product build?

michalszynkiewicz (approved):

sberyozkin:
The required build passed, I'd like to continue experimenting with the product builds tomorrow hence merging

PR #1047, opened 2018-07-18
THORN-2063: Migrate MP Fault Tolerance implementation to SmallRye

antoinesd:


fractions/microprofile/microprofile-fault-tolerance/pom.xml
```diff
@@ -35,6 +35,7 @@
     <swarm.fraction.tags>Eclipse MicroProfile,MicroServices,Fault Tolerance</swarm.fraction.tags>
     <version.arquillian-weld-embedded>2.0.0.Beta5</version.arquillian-weld-embedded>
     <version.awaitility>3.1.0</version.awaitility>
+    <version.smallrye-fault-tolerance>1.0.0</version.smallrye-fault-tolerance>
```
mkouba: This should go into the `build-parent` pom:
https://github.com/thorntail/thorntail/blob/master/build-parent/pom.xml#L112-L116

fractions/microprofile/microprofile-fault-tolerance/module.conf
```diff
@@ -8,8 +8,7 @@ org.jboss.logging
 org.wildfly.swarm.configuration.microprofile.config
 org.wildfly.extension.microprofile.config
 org.eclipse.microprofile.config.api
-
-org.eclipse.microprofile.faulttolerance export=true
+org.jboss.weld.core export=true
```
mkouba: Why is this module dependency needed?

fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/MicroProfileFaultToleranceFraction.java
```diff
@@ -16,27 +16,18 @@
...
     public MicroProfileFaultToleranceFraction() {
     }
 
-    public boolean isSynchronousCircuitBreakerEnabled() {
-        return synchronousCircuitBreaker.get();
-    }
-
-    @AttributeDocumentation("Enable/disable synchronous circuit breaker functionality. If disabled, `CircuitBreaker#successThreshold()` of value greater than 1 is not supported. Moreover, circuit breaker does not necessarily transition from `CLOSED` to `OPEN` immediately when a fault tolerance operation completes. However, applications are encouraged to disable this feature on high-volume circuits.")
-    @Configurable("swarm.microprofile.fault-tolerance.synchronous-circuit-breaker")
-    private Defaultable<Boolean> synchronousCircuitBreaker = Defaultable.bool(true);
```
mkouba: If we remove this it won't be possible to configure this functionality using yaml config files. Are you sure it's a good idea?

mkouba (changes requested):

fractions/microprofile/microprofile-fault-tolerance/pom.xml
mkouba: Also add `io.smallrye:smallrye-fault-tolerance` dependency to the `<dependencyManagement>` section in build-parent; there is a `<!-- SmallRye -->` comment where other MPs reside.

fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/MicroProfileFaultToleranceFraction.java
```diff
@@ -39,4 +40,5 @@ public boolean isSynchronousCircuitBreakerEnabled() {
     @AttributeDocumentation("Enable/disable synchronous circuit breaker functionality. If disabled, `CircuitBreaker#successThreshold()` of value greater than 1 is not supported. Moreover, circuit breaker does not necessarily transition from `CLOSED` to `OPEN` immediately when a fault tolerance operation completes. However, applications are encouraged to disable this feature on high-volume circuits.")
     @Configurable("swarm.microprofile.fault-tolerance.synchronous-circuit-breaker")
     private Defaultable<Boolean> synchronousCircuitBreaker = Defaultable.bool(true);
```
mkouba: Is the config option used somewhere?

antoinesd: Excellent question. After checking the answer is "no" :-(. Do you have a suggestion to make it usable? I think about creating a configuration bean in smallrye ft that we would specialize here... Do you have a better idea?

mkouba: Yes, you could either provide a bean which provides additional configuration or even try to implement an MP `ConfigSource`?

antoinesd: I added a new ConfigSource to read config from Thorntail but it is ignored. Any clue @mkouba ?

mkouba:
@antoinesd `@DeploymentModule(name = "org.wildfly.swarm.microprofile.faulttolerance", metaInf = MetaInfDisposition.IMPORT, export = true, services = Module.ServiceHandling.IMPORT)` adds the "main" slot by default (see also `org.wildfly.swarm.spi.api.annotations.DeploymentModule.slot()`). You should probably add "deployment" slot and put `ConfigurableConfigSource` under `org.wildfly.swarm.microprofile.faulttolerance.deployment` package.

antoinesd:
@mkouba I changed config to
`@DeploymentModule(name = "org.wildfly.swarm.microprofile.faulttolerance", metaInf = MetaInfDisposition.IMPORT, export = true, services = Module.ServiceHandling.IMPORT, slot = "deployment")`
and moved config source classes from `runtime` to `deployment` package.
Now I have a NCDFE on `org/wildfly/swarm/spi/api/config/ConfigView`. In the previous version, injection point of type `ConfigView` was unsatisfied...
[EDIT] Just pushed this new version

mkouba:
I think we need to do the same what `io.thorntail:cdi-config` does:

1. modify `module.conf` so that it also contains: 

```
org.wildfly.swarm.spi
org.wildfly.swarm.spi:runtime
org.wildfly.swarm.container:runtime export=true
```
("deployment" slot module does not declare a dependency on SPI module where `ConfigView` is defined)

2. Copy the functionality from `InjectConfigViewExtensionFactory` - I'm afraid that `CDI.current().select(ConfigView.class)` wouldn't work because the `ConfigView` bean lives in the internal CDI container instance

antoinesd:
Thanks Martin

mkouba:
In fact, I think we can move the config source (`ConfigView` to `ConfigSource` adapter) to the MP config fraction, in the MP config fraction declare a dependency on `cdi-config` fraction (so that we're able to inject `ConfigView`) and then it would be possible to access Thorntail internal config (yaml etc.) using MP config structures. @kenfinnigan @antoinesd @michalszynkiewicz WDYT?

antoinesd:
+1
That was my initial idea before I start loosing so much time in this class loader / container / dependencies hell ;-)

kenfinnigan:
I think that makes sense @mkouba

mkouba:
Ok, I've created https://issues.jboss.org/browse/THORN-2118 and will send a separate PR later today.

kenfinnigan:
What's still required to get this working?

antoinesd:
It's working now. Yaml configuration is taken into account and has priority on MP config

fractions/microprofile/microprofile-fault-tolerance/src/main/java/org/wildfly/swarm/microprofile/faulttolerance/MicroProfileFaultToleranceFraction.java
```diff
@@ -16,27 +16,24 @@
...
 /**
  * @author Antoine Sabot-Durand
  */
-@DeploymentModule(name = "org.wildfly.swarm.microprofile.faulttolerance", export = true, slot = "deployment", metaInf = MetaInfDisposition.IMPORT)
+@DeploymentModule(name = "io.smallrye.faulttolerance", metaInf = MetaInfDisposition.IMPORT, export = true, services = Module.ServiceHandling.IMPORT)
 public class MicroProfileFaultToleranceFraction implements Fraction<MicroProfileFaultToleranceFraction> {
 
     public MicroProfileFaultToleranceFraction() {
-    }
+        String synchronousCircuitBreaker = System.getProperty("swarm.microprofile.fault-tolerance.synchronous-circuit-breaker");
```
mkouba: I'm not so sure this would work. `@Configurable` is repeatable, so you can try to add another `@Configurable()` alias to the config property instead. 

By the way, I've created https://github.com/smallrye/smallrye-fault-tolerance/issues/10 ...

Anyway, a simple test would be good.

mkouba (approved):

PR #1053, opened 2018-07-19
THORN-2065 | Use smallrye-opentracing

pavolloffay:
Depends on: https://github.com/smallrye/smallrye-opentracing/issues/10

https://issues.jboss.org/browse/THORN-2065

Signed-off-by: Pavol Loffay <ploffay@redhat.com>


testsuite/docker/opentracing-jaeger/src/it/java/io/thorntail/testsuite/opentracing/jaeger/OpenTracingJaegerAppIT.java
```diff
@@ -64,42 +65,43 @@ public void test() throws Exception {
         System.err.println(tree);
         System.err.println("---");
 
-        assertThat(tree).hasRootSpans(2);
+        // TODO what is the first root span?
```
pavolloffay: The other root span is, which is a bit weir. I think it's comming from jax-rs instrumentation and it should be linked to jax-rs span (maybe bug in `SpanNode` as it is not parsing `FOLLWOS_FROM` reference type).

> [ spanId: 938a897918d5ad5c, operation:deserialize, tags: {entity.type=java.lang.String, media.type=*/*}]

pavolloffay:
This is ready to review/merge.

core/kernel/pom.xml
```diff
@@ -46,19 +46,16 @@
     <dependency>
       <groupId>io.opentracing</groupId>
       <artifactId>opentracing-mock</artifactId>
-      <scope>provided</scope>
```
kenfinnigan, line 49: Why was "provided" removed from these dependencies?

@bobmcwhirter, I believe it's still needed right?

Or is "optional" enough?

bobmcwhirter: Optional may be sufficient.  If it's `optional`, I think that implies `provided`, but of course the inverse is not true.

bobmcwhirter (dismissed):

pavolloffay:
It says dismissed review bc I squashed commits

bobmcwhirter (approved):

PR #1056, opened 2018-07-20
[THORN-2070] Avoid creating the temp themes dirs and allow combining custom and default themes

sberyozkin:
**Motivation**
Avoid the unnecessary temp themes folder creation and support the composition of thin custom themes modules with the default one 
**Modifications**
Add a new property to support the composition of theme modules, optimize the code to use the existing module instead of exploding it to the temp foilder 

**Result**
Faster and cleaner keycloak-server start up, and the option to easily customize the themes without the custom module having to copy all the KC themes. The former change is tested by the existing tests (and manual tests), the latter will be integration tested and demoed as part of THORN-2039


fractions/keycloak-server/src/main/java/org/wildfly/swarm/keycloak/server/runtime/KeycloakThemesCustomizer.java
```diff
@@ -55,25 +53,27 @@ public void customize() throws ModuleLoadException, IOException {
...
+                    .dir(".");
+            });
+        }
 
     }
+
+    /**
+     * Whether or not to combine the default themes when the custom themes are also available.
+     */
+    @Configurable("swarm.keycloak-server.combineDefaultAndCustomThemes")
```
Ladicek: The config keys are usually `all-lower-case-with-dashes` (because that's how they're generated for all the Config APIs). Not always, of course, we are wildly inconsistent, but I think we should strive for improving that :-)

sberyozkin: +1, will do a bit later

sberyozkin:
Updated the property name and the test too, it is a simple update but is still good, KC just fails to start if the default themes are not visible OOB that is why the existing test was OK to confirm it, but now it is a bit more specific. The actual combination of the custom and default themes will be tested as part of the demo later on

fractions/keycloak-server/src/main/resources/modules/org/keycloak/keycloak-themes/main/module.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<module xmlns="urn:jboss:module:1.5" name="org.keycloak.keycloak-themes" slot="main">
+<module xmlns="urn:jboss:module:1.5" name="org.keycloak.keycloak-themes">
   <resources>
-    <resource-root path="."/>
+    <artifact name="org.keycloak:keycloak-themes:3.4.3.Final"/>
```
kenfinnigan: Should that refer to a Maven property for the version instead of being hard coded?

kenfinnigan (approved):

PR #1068, opened 2018-08-08
[THORN-1876] Use the trace level to log the message that no security realm is configured

sberyozkin:


kenfinnigan (approved):

sberyozkin:
This branch became out of date but since the only change was about changing the log level I proceeded with the merge

PR #1070, opened 2018-08-09
THORN-2047 OpenApiDeploymentProcessor should not process implicit dep…

manovotn:
…loyments. Add automated test for this.

JIRA - https://issues.jboss.org/browse/THORN-2047

The test is a combination of mgmt-console test and that of OpenAPI.

### DIFF
--- a/testsuite/testsuite-management-console-openapi/src/test/java/org/wildfly/swarm/management/console/openapi/ManagementConsoleAndOpenAPITest.java
+++ b/testsuite/testsuite-management-console-openapi/src/test/java/org/wildfly/swarm/management/console/openapi/ManagementConsoleAndOpenAPITest.java
@@ -1,0 +1,68 @@
+package org.wildfly.swarm.management.console.openapi;
+
+/**
+ * Copyright 2018 Red Hat, Inc, and individual contributors.
+ *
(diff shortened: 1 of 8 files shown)


mkouba:
add to whitelist

testsuite/testsuite-management-console-openapi/src/test/java/org/wildfly/swarm/management/console/openapi/ManagementConsoleAndOpenAPITest.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+@RunWith(Arquillian.class)
+public class ManagementConsoleAndOpenAPITest {
+
+    @Drone
+    WebDriver browser;
+
+    @Deployment
+    public static Archive createDeployment() throws Exception {
+        return ShrinkWrap.create(WARArchive.class);
```
mkouba: Empty deployment?

manovotn: Ah yea, well honestly it doesn't really matter what's in the WAR (can be empty) as long as there are two deployments up and one is implicit (e.g. test is still valid).
I'll fix this.

sberyozkin (approved):

PR #1071, opened 2018-08-09
THORN-2110: MP JWT - clean up principal after handling a request

michalszynkiewicz:

-----

### DIFF
--- a/testsuite/testsuite-microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/SubjectExposingResource.java
+++ b/testsuite/testsuite-microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/SubjectExposingResource.java
@@ -1,0 +1,51 @@
+/**
+ * Copyright 2018 Red Hat, Inc, and individual contributors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.microprofile.jwtauth;
+
+import org.eclipse.microprofile.jwt.Claim;
+import org.eclipse.microprofile.jwt.Claims;
+
+import javax.annotation.security.RolesAllowed;
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+
+/**
+ * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ * <br>
+ * Date: 8/8/18
+ */
+@Path("subject")
+public class SubjectExposingResource {
+
+    @Inject
+    @Claim(standard = Claims.sub)
+    private String subject;
+
+    @GET
+    @RolesAllowed("MappedRole")
+    @Path("secured")
+    public String getSubjectSecured() {
+        return subject;
+    }
+
+    @GET
+    @Path("unsecured")
+    public String getSubjectUnsecured() {
+        return subject;
+    }
+
+}
--- a/testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/PrincipalLeakTest.java
+++ b/testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/PrincipalLeakTest.java
@@ -1,0 +1,79 @@
+/**
+ * Copyright 2018 Red Hat, Inc, and individual contributors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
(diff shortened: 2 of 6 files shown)


testsuite/testsuite-microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/SubjectExposingResource.java
```diff
@@ -0,0 +1,51 @@
+/**
```
sberyozkin, line 1: Looks like this resource should be part of the test JAX-RS Application ?

michalszynkiewicz: I'm not sure I understand, it's used here: https://github.com/thorntail/thorntail/pull/1071/files#diff-528eab1afb2045a3318c303d74d58aabR49

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/PrincipalLeakTest.java
```diff
@@ -0,0 +1,79 @@
...
+        deployment.addAsManifestResource(new ClassLoaderAsset("keys/public-key.pem"), "/MP-JWT-SIGNER");
+        return deployment;
+    }
+
+    @RunAsClient
+    @Test
+    public void subjectShouldNotLeakToNonSecuredRequest() throws Exception {
+        long start = System.currentTimeMillis();
+        try {
+            // By default, there are 64 threads/tasks
```
michalszynkiewicz, line 60: is there a way to limit the number of tasks/threads for a test?
If we limit it to 1 we have a faster and more reliable test...

testsuite/testsuite-microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/SubjectExposingResource.java
sberyozkin: Sorry, my fault, I thought for a moment it was part of the fraction source...

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/PrincipalLeakTest.java
sberyozkin: Single call should be enough, though it is up to you...

sberyozkin:
By the way, does MP-JWT require that the Claims can only be visible to the methods with RolesAllowed ?
(the subject claim inside the signed only JWT can easily be introspected in any case, for ex in the custom JAX-RS filter by simply Base64URL decoding the Authorization Bearer middle value)

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/PrincipalLeakTest.java
michalszynkiewicz: The problem the PR is fixing is that the principal is kept on thread local and it was not cleared.
So if some thread handled an authenticated request, the principal has been set for it. And then, if the same thread handled an unauthenticated request, the principal was still accessible.

Single thread is not enough to reproduce the problem because thorntail/undertow has a thread pool for handling the requests and by default the pool size is 64

sberyozkin:
It is a good idea, ex, JAX-RS SecurityContext is expected to return a null Principal if the user is not authenticated, so at least at the MP-JWT level and the custom code this 'sub' claim can become invisible but it probably won't block its visibility, example, MP-JWT can return a raw content and it will be there...And I'm not sure MP-JWT restricts the claims visibility to only RBAC-protected methods... 
Looks nice in general but might require a bit more investigation

sberyozkin:
Looking more carefully at the JIRA issue, IMHO the fix should be somewhere in the MPJwtProducer only, having a thread local state is problematic... Or, JAXRS ContainerResponseFilter will cleanup the TL state on every request, etc... The issue seems to be about not making 'sub' invisible to the application code in some cases but about preventing the thread-safety issues related to the injection of stale 'sub' values...

sberyozkin:
Yes, MPJWTProducer does not remove TL value proactively, so there is also a window there that the stale value can be picked up... Not only 'sub' but any other claim from the given JsonWebToken

michalszynkiewicz:
@sberyozkin It's refactored now to use undertow handler to clean up the principal.

sberyozkin:
@michalszynkiewicz Hi Michal, looks good to me. Might be worse updating the test a bit (future PR is ok), each thread have its own subject value, this way one would test better that a stale subject value from the previous call is not picked up by the current thread...

sberyozkin (approved):

michalszynkiewicz:
@sberyozkin I will add a todo to create such test

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/PrincipalLeakTest.java
Ladicek: I'm not even sure if it's hardcoded to 64, the number might be derived from number of CPU cores or something. You should be able to change the number of threads by configuring the default worker pool in the `io` subsystem.

michalszynkiewicz: @Ladicek thanks! I was looking at undertow configuration and didn't find anything, IO has `swarm.io.workers.KEY.task-max-threads` :)

PR #1073, opened 2018-08-10
Fixing few minor MP deps issues

sberyozkin:


mkouba (approved):
Looks good

fractions/microprofile/microprofile/pom.xml
```diff
@@ -125,4 +121,21 @@
     </dependency>
   </dependencies>
 
+  <profiles>
```
kenfinnigan, line 124: Why was this added?

The entirety of MP should be part of product

mkouba: Hm, sorry for wrong approval. Now I'm a little bit confused. What MP do we acutally implement in community/product? When looking at MP APIs versions it's like a mix of MicroProfile 1.4 (config, jwt) and 1.3 (openapi, rest-client, opentracing, FT, health, metrics).

kenfinnigan: Technically it's MP 1.3, but some of the sub specs are actually further along

PR #1074, opened 2018-08-13
[THORN-2132] Update the fraction plugin version to 88

sberyozkin:


sberyozkin:
The stage PNC build passes with this update

mkouba (approved):

PR #1075, opened 2018-08-13
[THORN-2133] Moving the resource-adapters fraction into the product

sberyozkin:


kenfinnigan:
I think the testsuite needs to be moved as well

kenfinnigan (approved):

PR #1084, opened 2018-08-16
THORN-1564 - Enable java.util.Properties for configurations

michbeck100:

-----

This PR enables Fractions to use `java.util.Properties` as configuration properties. This was explicitly blacklisted [here](https://github.com/thorntail/thorntail/blob/79b59c0cf3729fcd33bf1beefb9e1c931a216602/core/container/src/main/java/org/wildfly/swarm/container/runtime/ConfigurableManager.java#L70) before. But most of the code was there already. 
Also the support for `java.util.Map` and `java.util.List` seems to be complete, but stays blacklisted.

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/container/config/ConfigNode.java
+++ b/core/container/src/main/java/org/wildfly/swarm/container/config/ConfigNode.java
@@ -19,6 +19,7 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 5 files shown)


kenfinnigan (dismissed):
Looks ok to me. Any comments @bobmcwhirter?

michbeck100:
There seems to be general problem with the semaphore builds. It fails at `mvn clean`.

michbeck100:
Had to add the snapshot repository for thorntail-maven-plugin to fix the semaphore build (674af97).

kenfinnigan:
Semaphore build only works with 4.x, so adding the snapshot repository isn't needed

michbeck100:
Reverted with 071a6e6

kenfinnigan (approved):

PR #1087, opened 2018-08-17
[THORN-2141] Deprecating the archaius and hystrix fractions 

sberyozkin:


kenfinnigan (approved):
Are the hystrix and archaius fractions part of the "supported" profile already?

If not, the netflix module might need to be split into entries for each sub module

sberyozkin:
Hi Ken,
https://github.com/thorntail/thorntail/blob/master/fractions/netflix/pom.xml#L183
and fractions/netflix is referenced from both supported and unsupported profiles. I believe this is how archaius and hystrix end up being in the supported profile

PR #1088, opened 2018-08-20
MP JWT fixes

Ladicek:
- [THORN-2142](https://issues.jboss.org/browse/THORN-2142)
- [THORN-2143](https://issues.jboss.org/browse/THORN-2143)
- [THORN-2144](https://issues.jboss.org/browse/THORN-2144)

Tests for THORN-2142 and THORN-2143 are already present (and currently failing) in https://github.com/rhoar-qe/thorntail-test-suite. Test for THORN-2144 added there as well.

### DIFF
--- a/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MpJwtFilterRegistrar.java
+++ b/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MpJwtFilterRegistrar.java
@@ -18,10 +18,13 @@ package org.wildfly.swarm.microprofile.jwtauth;
 import javax.annotation.security.DenyAll;
 import javax.annotation.security.PermitAll;
 import javax.annotation.security.RolesAllowed;
+import javax.ws.rs.HttpMethod;
 import javax.ws.rs.container.DynamicFeature;
 import javax.ws.rs.container.ResourceInfo;
 import javax.ws.rs.core.FeatureContext;
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.net.URL;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -47,10 +50,16 @@ public class MpJwtFilterRegistrar implements DynamicFeature {
     public void configure(ResourceInfo resourceInfo, FeatureContext context) {
         Annotation mpJwtAnnotation = getMpJwtAnnotation(resourceInfo);
         if (mpJwtAnnotation != null) {
-            if (DenyAll.class.isInstance(mpJwtAnnotation)) {
+            if (mpJwtAnnotation instanceof DenyAll) {
                 configureDenyAll(context);
-            } else if (RolesAllowed.class.isInstance(mpJwtAnnotation)) {
+            } else if (mpJwtAnnotation instanceof RolesAllowed) {
                 configureRolesAllowed((RolesAllowed) mpJwtAnnotation, context);
+            }
+        } else {
+            // the resource method is not annotated and the class is not annotated either
+            if (hasSecurityAnnotations(resourceInfo) && shouldNonannotatedMethodsBeDenied()) {
+                // some other method has a security annotation and this one doesn't, it should be @DenyAll by default
+                configureDenyAll(context);
             }
         }
     }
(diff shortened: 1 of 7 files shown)


fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MpJwtFilterRegistrar.java
```diff
@@ -92,4 +100,21 @@ private Annotation getAnnotation(Annotation[] declaredAnnotations,
                         ". Expected at most 1 annotation, found: " + annotations);
         }
     }
+
+    private boolean hasSecurityAnnotations(ResourceInfo resource) {
+        // TODO check all methods of the class, or just resource methods? public/private/etc.? inherited?
```
Ladicek: @mkouba I recall you were the one to add `swarm.microprofile.jwt.default-missing-method-permissions-deny-access`. Could you advise here? I think our docs say that only resource methods are considered here (http://docs.wildfly-swarm.io/2.1.0.Final/#_microprofile_jwt_rbac_auth), but not sure about e.g. inheritance. If only JAX-RS's `ResourceInfo` provided a list of all resource methods :-)

Ladicek: Ah Martin is on PTO this week. Hmm. Anyone? :-)

sberyozkin: Does the spec require that only security annotated methods can be invoked ? Role based access is an extra restriction...I.e, one service can be happy enough with the token be validated (client - authenticated only)

Ladicek: Implicit `@DenyAll` for all methods in a class that don't have a security annotation yet some other method does mirrors the behavior we have for EJBs. The configuration key and its default value does the same. Not sure if that answers your question :-)

sberyozkin: Hi Ladislav, I was not sure, if the token can only be used as the authorization token, which what would happen with your change. I've looked at the article about MP JWT, my understanding it should be also usable as only the authentication token, ie, if it is valid hence the client is authenticated and the service is free to access the claims without having to put RolesAllowed, etc...
thanks

Ladicek: I originally assumed the same -- see https://issues.jboss.org/browse/THORN-1954. In this PR, I'm just returning to our own previous behavior.

sberyozkin: Right, if adding something like @RolesAllowed("*") will still work then it is fine, even if it might be a bit of an extra work.

sberyozkin (dismissed):

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MpJwtFilterRegistrar.java
Ladicek: `@RolesAllowed("*")` is another thing I'm restoring in this PR :-)

Ladicek:
Hmm I'll take a look at the failing test.

EDIT: yea, that's very simple, the test expects that a method without security annotation is always accessible, which is a bug. Test is very easy to fix.

Ladicek:
I've finished the implementation of implicitly denying access to methods without security annotations. It no longer checks all the methods in the class, just the JAX-RS resource methods, as documented.

Ladicek:
Ah and in addition to fixing the `PrincipalLeakTest`, I also improved how the threads are handled. @michalszynkiewicz you might be interested: https://github.com/thorntail/thorntail/pull/1088/commits/26c76c75f6029e27cc52d165dd8cd0acf10ed10a :-)

Ladicek:
Windows failures are 1. the usual MP FT TCK (`BulkheadSynchRetryTest.testBulkheadClassSynchronousPassiveRetry55`), 2. not sure why but also MP Config TCK. Both seem unrelated.

kenfinnigan:
I'm not sufficiently versed in JWT to know whether the changes are good or not.

@sberyozkin, @michalszynkiewicz could one of you review and approve?

michalszynkiewicz (approved):

PR #1091, opened 2018-08-21
[THORN-2147] Update jaeger.gson version to 2.8.2

sberyozkin:


Ladicek (approved):

sberyozkin:
This Win failure does not seem to be related, jaeger.gson is not in the FT domain...

PR #1096, opened 2018-08-28
[THORN-2151] Updating smalrye-config, smallrye-health, smallrye-opentracing to the latest versions

sberyozkin:


Ladicek (approved):

sberyozkin:
JIRA issue is already created for tracking the MP config test failures (not only on Win but Linux as well), and MP FT win test failure is unrelated

PR #1098, opened 2018-08-28
THORN-2154: NPE in MetricsHttpHandler.handleRequest

Ladicek:
Motivation
----------
The `MetricsHttpHandler.handleRequest` method throws a NPE if there's
no `Accept` header.

Modifications
-------------
Do a `null` check.

Result
------
No NPE.

-----

### DIFF
--- a/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime/MetricsHttpHandler.java
+++ b/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime/MetricsHttpHandler.java
@@ -25,6 +25,7 @@ import io.undertow.util.HttpString;
 import org.jboss.logging.Logger;
 
 import java.util.concurrent.CountDownLatch;
+import java.util.stream.Stream;
 
 /**
  * @author hrupp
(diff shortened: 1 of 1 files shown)


fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime/MetricsHttpHandler.java
```diff
@@ -60,7 +61,8 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
 
         String method = exchange.getRequestMethod().toString();
         HeaderValues acceptHeaders = exchange.getRequestHeaders().get(Headers.ACCEPT);
-        metricsHandler.handleRequest(requestPath, method, acceptHeaders.stream(), (status, message, headers) -> {
+        Stream<String> acceptHeadersStream = acceptHeaders != null ? acceptHeaders.stream() : null;
+        metricsHandler.handleRequest(requestPath, method, acceptHeadersStream, (status, message, headers) -> {
```
Ladicek, line 65: TBH, I really don't think `Stream` is an acceptable parameter type or return type in a public API (for the same reason as `Iterator`), but SmallRye decided otherwise.

michalszynkiewicz:
@Ladicek THORN-2154 is a duplicate of THORN-2137, I was working on the latter today but you were faster :)

The only thing I did differently was I added a test that reproduces the problem: https://github.com/thorntail/thorntail/pull/1099/files#diff-6c12828f4fda3f5f080c9bf400773698
Could you add it to your commit?

Also, stream there is my fault, feel free to replace it with whatever you find better.

Ladicek:
You didn't really have to close your PR, I think it's better than mine because it has the test :-) (I didn't add one because we already have a test in the standalone test suite.)

michalszynkiewicz:
I closed mine when I realized you have a PR for exactly same thing :)

PR #1101, opened 2018-09-03
THORN-1890 Thorntail runner

michalszynkiewicz:
# Thorntail Runner
The change adds a new module - thorntail runner.
The main entry point for it is called `Runner` and it should be used to trigger a Thorntail v2 app in the IDE.

## What's going on
The `Runner` first runs `FatJarBuilder` to generate an uber jar, then sets up a classpath that only contains the generated jar and then invokes `Main#main` in it.

### Dependency filtering
In the standard (maven plugin) scenario, libraries are removed from user's WAR if they are not in the dependency tree for user-defined (non-Thorntail) dependencies.
Here, all the dependencies in the dependency tree of the Thorntail fractions are removed from the WAR. It means that some user dependency might get removed. There's a system property to prevent such dependencies from getting deleted (the dependencies have to be listed explicitly).

## Problems
Potential problems:
- the solution is based on reverse-engineering the classpath. Thorntail fractions (and other Thorntail elements) are identified by the metadata put by Maven into jars' `META-INF`. I'm afraid it might be a bit fragile
- it's all based on `BuildTool` and I used it as a rather black box
- dependency filtering is done based on artifact check sums. To speed it up, check sums for fraction dependencies are calculated when building fractions (calculated because the check sums in Maven Central are sometimes wrong - e.g. for Shrinkwrap). If someone changes an artifact in Central (or some other repository), the artifact won't be filtered out

So far it was tested only on IntelliJ IDEA on Linux and Windows and Eclipse on Linux

### DIFF
--- a/thorntail-runner/src/main/java/org/wildfly/swarm/runner/Runner.java
+++ b/thorntail-runner/src/main/java/org/wildfly/swarm/runner/Runner.java
@@ -1,0 +1,183 @@
+/**
+ * Copyright 2018 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.swarm.runner;
+
+import org.wildfly.swarm.runner.cache.RunnerCacheConstants;
+
+import java.io.File;
+import java.io.IOException;
+import java.lang.reflect.Method;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+
+/**
+ * The class to execute when running Thorntail in the IDE.
+ *
+ * During execution, it resolves project's dependencies to ${user.home}
+ *
+ * It first generates an uber-jar and then executes it.
+ * The jar is generated from the IDE classpath by taking the Thorntail dependencies, generating an uber-jar in the similar way
+ * the thorntail-maven-plugin does and removing all the transitive dependencies of the Thorntail elements from WEB-INF/lib
+ * in the internal WAR.
+ *
+ * The following options are available to control the execution:
+ * <ul>
+ *     <li>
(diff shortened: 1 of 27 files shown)


Ladicek:
I only took a cursory look and I've got 2 comments:

1. What if the application is not a WAR but a JAR?
2. Removing dependencies that are transitively brought in by Thorntail elements is fragile as hell. At first, I thought I could suggest inspecting the Maven project, but then I realized this is about running from IDE. Oh well. To be 100% honest, I believe we should adjust `mvn wildfly-swarm:run` to default to constructing an uberjar, and point our users towards that even for an IDE usecase. Performance is good, but correctness is better.

michalszynkiewicz:
@Ladicek thanks for taking a look!

1. Not supported (at the moment?)
2. That's why there's a property to include such dependencies anyway. It's not perfect though, it might be required to include a lot of them... The removed dependencies will only be exactly the same groupId:artifactId:version but it's still entirely possible. That's why I didn't touch the thorntail:run. My plan was to, as you wrote, make it always use the uber jar.

I'm working on introducing some caches in the jar building phase, it should let a user work a bit faster than with `thorntail:run`, also debugging is easier with `Runner` or `Swarm` than with the maven plugin.

Ladicek:
Yea, I understand the IDE experience is much better with this.

Supporting JARs is likely needed. Not sure if you need to do it right now, but should keep track of it.

mkouba:
I've tried the `Runner` with Eclipse IDE and it works fine, incl. debugging. But it seems to be quite slow - it takes 40 seconds to start so some caching would be definitely useful.

Note that I've only tried the WAR use case.

mkouba:
So caching improved the startup time to ~ 20 seconds for a simple app (jaxrs + cdi + microprofile-openapi).

mkouba:
And with `-Dthorntail.runner.caching` it's ~ 10 seconds!

thorntail-runner/src/main/java/org/wildfly/swarm/runner/Runner.java
```diff
@@ -0,0 +1,189 @@
...
+ *     <li>
+ *         <b>thorntail.runner.user-dependencies</b> - a comma-separated list of groupId:artifactId:version of artifacts
+ *         to keep in WEB-INF/lib even though they are dependencies of Swarm elements
+ *     </li>
+ *     <li>
+ *         <b>thorntail.runner.main-class</b> - user's main class replacing the default org.wildfly.swarm.Swarm class.
+ *         Please note that using custom main class is discouraged.
+ *     </li>
+ *     <li>
+ *         <b>thorntail.runner.webapp-location</b> - by default, Runner expects webapp in `src/main/webapp`. This property
```
kenfinnigan, line 61: Could this value be retrieved from Maven Project settings?

Ladicek: `Runner` lives outside of a Maven invocation.

michalszynkiewicz: As @Ladicek wrote, right now Thorntail Runner is build system agnostic and doesn't know anything about maven configuration. I'll add a TODO to try to take it from pom if present.

michalszynkiewicz:
I still need to remove the dependency on javax.xml.bind and bring back camel fractions

kenfinnigan (approved):

PR #1105, opened 2018-09-05
THORN-2157: make sure we document breaking changes in release announcements

Ladicek:
Motivation
----------
The script that fetches the list of fixed issues in given release
could also print the list of all breaking changes in given release
(that is, issues with label `breaking_change`). That of course
won't help with _tracking_ breaking changes, but if we track them,
this should at least help us to not forget about documenting them.

Modifications
-------------
Modified the `fetch-notes.js` script to also include a listing of
all JIRAs that are labeled `breaking_change`.

Result
------
Extra section in the release announcement with a list of breaking
changes. Hopefully, we won't forget to detail them in writing.

-----

### DIFF
--- a/release/HOWTO_RELEASE.md
+++ b/release/HOWTO_RELEASE.md
@@ -114,6 +114,7 @@ Go to the local `wildfly-swarm.io`.
 
 Create wildfly-swarm.io/src/posts/announce-2.0.0.Final.adoc, replace 2.0.0.Final with the new version.
 Copy the fetched contributors and change log into this document.
+Make sure at least the breaking changes, as listed in the output of `fetch-notes.js`, are documented in detail.
 
 # Update Che image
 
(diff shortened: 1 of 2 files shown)


kenfinnigan:
Does this mean the same JIRA is repeated twice? Once in changelog and once in breaking changes?

Should breaking changes come before changlog?

Ladicek:
> Does this mean the same JIRA is repeated twice? Once in changelog and once in breaking changes?

Yes. Not sure if that's good or bad, but I tend to think it definitely isn't terribly wrong :-)

> Should breaking changes come before changlog?

Not sure. My intent is to highlight the breaking changes to us, so that we don't forget to detail them in the written text. In such case, the breaking changes listing can easily be after the changelog.

kenfinnigan:
Usually the output generated from this is blindly copied into the release blog.

If there's more to be done than just that, the release doc probably needs updating to state what should be done

Ladicek:
I added a sentence about that.

kenfinnigan (approved):

PR #1107, opened 2018-09-07
[THORN-2149] Minor updates

sberyozkin:


kenfinnigan:
Isn't this something that should only be in product?

Wouldn't this change remove everything if we ran it on upstream to generate licenses?

sberyozkin:
I was asking in our email exchange, if it should make it to the upstream, and there were no objections then. I thought of syncing it with the upstream because we were caught in one of the product builds and had no idea what to do till Michal spotted there was a prod only change in 7.1.x. This should be safe given the upstream builds do not use this profile now. I've no strong preference, may be it should be documented in the productization docs, or may be kept in the upstream and removed when it becomes problematic. Or may be have another profile, -PlicensesCommunity ?

kenfinnigan:
I just remember that when it was written originally, we envisioned this setting only being in product stream.

If everyone's ok with that changing, then that's fine

sberyozkin:
Hi Ken
I'll restart the discussion to make sure all are happy with this change or if other options will be preferred.
Thanks, Sergey

Ladicek:
I'm honestly not sure what the options do (CC @michalszynkiewicz), but I thought the `licenses` profile is only useful for product anyway (e.g. it defines `<additionalBom>io.thorntail:bom-certified:pom:${project.version}</additionalBom>`)?

michalszynkiewicz (approved):

PR #1108, opened 2018-09-07
[THORN-2162] Dependencies and version updates to improve PME results

sberyozkin:


pom.xml
```diff
@@ -341,22 +341,22 @@
       <dependency>
         <groupId>io.thorntail</groupId>
         <artifactId>arquillian</artifactId>
-        <version>2.3.0.Final-SNAPSHOT</version>
```
kenfinnigan: Will this even work when releasing upstream?

I believe we had that issue in the past which is why the versions are explicitly in the pom

build-parent/pom.xml
```diff
@@ -122,6 +122,8 @@
     <version.smallrye.jwt>1.1.0</version.smallrye.jwt>
     <version.smallrye-fault-tolerance>1.0.1</version.smallrye-fault-tolerance>
 
+    <!-- For PME alignments -->
```
kenfinnigan: Why do we need to align slf4j? What's bringing in that dependency?

My concern is it will have knock-on effects with our dependencies when we're not really meant to be dictating a particular logging framework

pom.xml
sberyozkin: OK, I'll just revert this part

build-parent/pom.xml
sberyozkin: I don't recall right now, there were various minor variations, an this is what was set in the product branch in the previous product. But it will make sense to deal with the specific version updates at the build parent level in a separate PR

sberyozkin:
Ok, reverted some of them (slf4j alignment also helped with the community list, but yes, I suppose the changes like that should go with more details). MP rest client version update to 1.0.1 is needed as 1.0 is not used by smallrye-rest-client and was causing the duplication

sberyozkin:
commons-configuration 1.10 is productized, 1.8 not, minor version update, I saw no issues with the MP FT tests

kenfinnigan (approved):

juangon:
Hi @kenfinnigan , don't have merge permission. Can you merge it yourself please if no problems found? Thanks!

Ladicek:
Failing test on Windows (`org.eclipse.microprofile.config.tck.ConfigProviderTest.testEnvironmentConfigSource`) is unrelated and happens for a while now, merging.

PR #1112, opened 2018-09-12
THORN-2165 Fix topology-openshift

juangon:
Motivation
----------
topology-openshift fraction has some API discrepancy between third party libraries (okhttp and okhttp-ws) due to an upgrade in okhttp, causing this fraction to not work.

Modifications
-------------
As okhttp-ws doesn't have a newer version, downgrade okhttp so they are both compatible.

Result
------
topology-openshift fraction works again.

-----

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -67,7 +67,8 @@
 
     <version.openshift.client>5.6.0.Final</version.openshift.client>
     <version.com.squareup.okhttp>3.9.0</version.com.squareup.okhttp>
-    <version.com.squareup.okhttp-ws>3.4.2</version.com.squareup.okhttp-ws>
+    <!-- THORN-2165 Align okhttp and okhttp-ws for topology-openshift as there is API dependency and okhttp-ws has older version -->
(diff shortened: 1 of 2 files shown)


build-parent/pom.xml
```diff
@@ -67,6 +67,7 @@
 
     <version.openshift.client>5.6.0.Final</version.openshift.client>
     <version.com.squareup.okhttp>3.9.0</version.com.squareup.okhttp>
+    <version.com.squareup.okhttp-openshift>3.4.2</version.com.squareup.okhttp-openshift>
```
Ladicek: I think we should have just one version property for OkHttp for OpenShift, because `okhttp` and `okhttp-ws` need to be aligned. So I'd just remove the line below and used `version.com.squareup.okhttp-openshift` for both `okhttp` and `okhttp-ws` in the `module.xml`.

Ladicek: And probably add a comment in the `module.xml` why we need to stay on 3.4.2.

juangon:
@Ladicek forgot to say I pushed suggestions this morning

Ladicek (approved):

Ladicek:
Windows failures are unrelated (MP Config and MP FT TCKs).

PR #1113, opened 2018-09-12
[THORN-2162] Update findbugs version to 2.0.3

sberyozkin:
2.0.0 version is coming from archaius-core, no match was found for 2.0.0 by PME, the closest productized version is 2.0.3 and this change proved effective


kenfinnigan:
Silly question, does findbugs even need to be there at all?

Seems more like a dev type thing vs runtime

Ladicek:
To me, this looks like it's about the JSR 305 annotations. The JSR never completed, but the FindBugs project ships them and some projects in the Java world use them.

Ladicek (approved):

PR #1116, opened 2018-09-17
THORN-2121: add a HowTo for distributed tracing with MP OpenTracing and Jaeger

Ladicek:

-----


Ladicek:
CC @tradej @jclingan

docs/howto/distributed-tracing/proc_tracing-single-service.adoc
```diff
@@ -0,0 +1,85 @@
...
+----
+--
+
+. Configure the Jaeger tracer in `project-defaults.yml`.
++
+--
+[source,yaml]
+----
+swarm:
+include::src/main/resources/project-defaults.yml[tag=jaeger,indent=2]
```
Ladicek: IntelliJ IDEA renders all includes just fine except of this one. @tradej Do you maybe see an issue here?

docs/howto/distributed-tracing/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+    <dependency>
+      <groupId>io.thorntail</groupId>
+      <artifactId>jaeger</artifactId>
+    </dependency>
+    <!-- end::jaeger[] -->
+    <dependency>
+      <groupId>io.thorntail</groupId>
+      <artifactId>jaxrs</artifactId>
+    </dependency>
+    <!-- tag::mp-opentracing[] -->
```
kenfinnigan: I'm not 100% on this, but should it just be the whole of MP and jaeger fractions?

kenfinnigan:
Should we also mention that all JAX-RS resources are traced by default even without `@Traced`?

Also, should we document how to set the external reporter URL on `project-defaults.yml` which is needed in environments like OpenShift/Istio?

Ladicek:
> all JAX-RS resources are traced by default even without @Traced?

Are they? I had no idea.

> should we document how to set the external reporter URL on project-defaults.yml which is needed in environments like OpenShift/Istio?

I briefly mentioned that this is required, but didn't go into details. I probably should, but I don't understand the options that the `jaeger` fractions provides to achieve this.

kenfinnigan:
> > all JAX-RS resources are traced by default even without @Traced?
> 
> Are they? I had no idea.
> 
Check with @pavolloffay, but I believe so.

> > should we document how to set the external reporter URL on project-defaults.yml which is needed in environments like OpenShift/Istio?
> 
> I briefly mentioned that this is required, but didn't go into details. I probably should, but I don't understand the options that the `jaeger` fractions provides to achieve this.

Last two lines of https://github.com/kenfinnigan/mesh-microprofile-2018/blob/master/greeting-service/src/main/resources/project-defaults.yml are required for Istio usage, only last one to report to external location

Ladicek:
Yea, `remote-reporter-http-endpoint` seems to be one of the options to configure an external tracer. Looking at http://docs.wildfly-swarm.io/2.2.0.Final/#_jaeger, it seems there are others (`swarm.jaeger.agent-host` and `swarm.jaeger.agent-port`), which are probably required in different situations (UDP transport). I'll take a deeper look.

docs/howto/distributed-tracing/pom.xml
Ladicek: I could put them together. Actually I could even show `jaxrs` in the document, because MP OpenTracing only integrates on that level.

pavolloffay:
> all JAX-RS resources are traced by default even without @Traced?

yes all server endpoints are traced even without traced annotation, no need to add them. The annotation can be used to disable tracing or override operation name.

docs/howto/distributed-tracing/proc_tracing-single-service.adoc
Ladicek: I changed this to include the entire file, so no longer have this issue.

docs/howto/distributed-tracing/pom.xml
Ladicek: Done.

Ladicek:
I pushed a bunch of changes and I believe I resolved all the comments raised so far.

kenfinnigan (dismissed):

tradej:
I'll have a look at the docs later today, thank you.

docs/howto/distributed-tracing/src/main/resources/project-defaults.yml
```diff
@@ -0,0 +1,5 @@
+swarm:
+  jaeger:
```
pavolloffay, line 2: great that we are using project-defaults. We should also change example in the examples repo to use it - instead of using jaeger env vars.

docs/howto/distributed-tracing/src/main/resources/project-defaults.yml
```diff
@@ -0,0 +1,5 @@
...
+    service-name: my-traced-service
```
pavolloffay: It's better to name it with some business name or thorntail-example

Ladicek: Ayea, `greeter` would definitely be better.

Ladicek: Fixed.

tradej (changes requested):
The docs need tweaks because the procedures in the documentation are not semantically independent (they refer one another, and it seems to me that some semi-heavy reshuffling of the content is necessary). I'm sorry not to have managed this today, but I'll submit a PR with suggested changes in @Ladicek's repository tomorrow.

Ladicek:
@tradej I'd be happy to make any suggested changes, if you've got a general idea. Yes, I made the 2nd procedure dependent on the 1st, because merging the two procedures would make it harder to follow, but it would also be an option.

tradej (changes requested):
PR with changes submitted at https://github.com/Ladicek/thorntail/pull/1

Ladicek:
I merged @tradej's changes, so they are now present in this PR.

tradej (dismissed):

tradej:
I've submitted some more changes after an IRC discussion with @Ladicek in Ladicek#3.

Ladicek:
@tradej's changes have been merged and are now in this PR.

tradej (approved):

PR #1117, opened 2018-09-18
[THORN-2171] Remove health-api as dependency

ruromero:
Signed-off-by: Ruben Romero Montes <rromerom@redhat.com>

- [ ] Have you built the project locally prior to submission with `mvn clean install`? --> No, build fails for some other reason

Fix https://issues.jboss.org/browse/THORN-2171

-----

### DIFF
--- a/fractions/microprofile/microprofile-health/pom.xml
+++ b/fractions/microprofile/microprofile-health/pom.xml
@@ -65,15 +65,9 @@
 
     <dependency>
       <groupId>io.thorntail</groupId>
-      <artifactId>health-api</artifactId>
-      <version>${project.version}</version>
-    </dependency>
-
-    <dependency>
-      <groupId>io.thorntail</groupId>
       <artifactId>management</artifactId>
     </dependency>
-    
+
     <dependency>
       <groupId>io.thorntail</groupId>
       <artifactId>naming</artifactId>


Ladicek:
I wonder why does MP Health even have a dependency on `health-api`. It belongs to the old `monitor` fraction which is deprecated.

ruromero:
@Ladicek I should have used @org.eclipse.microprofile.health.Health

ruromero:
Anyhow, as you just said, microprofile-health  depends on health-api, I don't know if this is correct.

Ladicek:
Yea I think that's an actual bug.

ruromero:
@Ladicek updated commit accordingly

kenfinnigan:
ok to test

Ladicek (approved):

Ladicek:
Windows failures (MP Config TCK) are unrelated.

sberyozkin:
I've kept this dependency because the microprofile-health fraction still supports the legacy annotations, it checks the names of those annotations. That code has to be removed as well if it is not needed

Ladicek:
Interesting, I had no idea our implementation of MP Health still supports the old `monitor` stuff. It doesn't make much sense to me.

Ladicek:
Looking at the code and the only thing our impl of MP Health does is that it allows `@org.wildfly.swarm.health.Health` instead of `@org.eclipse.microprofile.health.inject.Health`, that's all. I wonder how many users use the old annotation with MP Health -- I suspect it's zero so we don't have to care.

And given how the code refers to the old annotation (it's just a Jandex lookup, nothing else), this could only be a compile-time breakage, which can be fixed by adding the dependency on `health-api` explicitly.

PR #1121, opened 2018-09-20
THORN-1830: arquillian ignores swarm.http.port setting in project-defaults.yml

roguexz:
Motivation
----------
Thorntail aims at simplifying the whole development experience. The ability to test code using the prescribed (Arquillian based) methods, right from within the IDE plays a huge role in the experience. I noticed that when running Arquillian based test cases from within IntelliJ, the resources (project-defaults.yml, etc.) were not being picked up as part of the built archive. More details about the IDE behavior can be found in the issue.

Modifications
-------------
Updated the scenario generator to not filter out the resource folders.

Result
------
I can now run Arquillian based tests from within the IDE itself.

-----

### DIFF
--- a/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/DefaultDeploymentScenarioGenerator.java
+++ b/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/DefaultDeploymentScenarioGenerator.java
@@ -9,13 +9,11 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
-import java.security.CodeSource;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import org.jboss.arquillian.container.spi.client.deployment.DeploymentDescription;
 import org.jboss.arquillian.container.test.impl.client.deployment.AnnotationDeploymentScenarioGenerator;
@@ -69,19 +67,10 @@ public class DefaultDeploymentScenarioGenerator extends AnnotationDeploymentScen
 
         ClassLoader cl = testClass.getJavaClass().getClassLoader();
 
-        Set<CodeSource> codeSources = new HashSet<>();
-
         URLPackageScanner.Callback callback = (className, asset) -> {
             ArchivePath classNamePath = AssetUtil.getFullPathForClassResource(className);
             ArchivePath location = new BasicPath(classPrefix, classNamePath);
             archive.add(asset, location);
-
-            try {
-                Class<?> cls = cl.loadClass(className);
-                codeSources.add(cls.getProtectionDomain().getCodeSource());
-            } catch (ClassNotFoundException | NoClassDefFoundError e) {
-                //e.printStackTrace();
-            }
         };
 
         URLPackageScanner scanner = URLPackageScanner.newInstance(
@@ -92,20 +81,10 @@ public class DefaultDeploymentScenarioGenerator extends AnnotationDeploymentScen
 
         scanner.scanPackage();
 
-        Set<String> prefixes = codeSources.stream().map(e -> e.getLocation().toExternalForm()).collect(Collectors.toSet());
-
         try {
             List<URL> resources = Collections.list(cl.getResources(""));
 
             resources.stream()
-                    .filter(e -> {
-                        for (String prefix : prefixes) {
-                            if (e.toExternalForm().startsWith(prefix)) {
-                                return true;
-                            }
-                        }
-                        return false;
-                    })
                     .filter(e -> e.getProtocol().equals("file"))
                     .map(e -> getPlatformPath(e.getPath()))
                     .map(e -> Paths.get(e))
--- a/testsuite/microprofile-tcks/config/pom.xml
+++ b/testsuite/microprofile-tcks/config/pom.xml
@@ -79,31 +79,76 @@
 
    </dependencies>
 
-   <build>
-      <plugins>
-         <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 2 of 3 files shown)


arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/DefaultDeploymentScenarioGenerator.java
```diff
@@ -92,20 +91,10 @@
 
         scanner.scanPackage();
 
-        Set<String> prefixes = codeSources.stream().map(e -> e.getLocation().toExternalForm()).collect(Collectors.toSet());
```
Ladicek, line 95: If we remove `prefixes`, we can also remove `codeSources` and everything related, as that's only used for this filtering.

kenfinnigan:
I'm confused as to why the comment mentions 4.x when this PR is against 2.x

roguexz:
@kenfinnigan I figured the master line makes it back to 4.x as well. If that's not the case, I can reword the comment.

arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/DefaultDeploymentScenarioGenerator.java
roguexz: @Ladicek I believe I will be able to remove the `try { codesources.add() }` block but keep the rest (as that is used for preparing the archive).

kenfinnigan:
@roguexz 4.x and master are very different, so separate PRs are required if you want something in both

roguexz:
@kenfinnigan Fixed the Git commit message (as well as the PR message).
@Ladicek Included the requested changes.

tools/src/main/java/org/wildfly/swarm/tools/exec/SwarmProcess.java
```diff
@@ -119,7 +119,9 @@ public int stop(long timeout, TimeUnit timeUnit) throws InterruptedException {
         }
         this.process.destroy();
         if (!this.process.waitFor(timeout, timeUnit)) {
-            process.destroyForcibly();
+            // Attempt to terminate the process forcibly and also wait for the same amount of time as the base attempt.
+            // This should potentially take care of the issues with the process not terminating on Windows CI build.
+            process.destroyForcibly().waitFor(timeout, timeUnit);
```
roguexz, line 124: Looking at the errors in the Windows build, looks like the process isn't terminating in 2 minutes, which in itself is weird and needs to be looked at. When invoking the "destroyForcibly()" method, we still need to check if the process is alive or not. Instead of adding a lot of logic, I am including an additional `waitFor` using the same arguments as the above invocation. Let's see if the Windows build goes through fine or not and how much longer does it take for completion.

roguexz:
Is there a problem with the Windows tests? I noticed the following,

The [MP config test](https://ci.wildfly-swarm.io/job/thorntail-pull-request-windows/io.thorntail.testsuite$thorntail-microprofile-tck-config/645/testReport/org.eclipse.microprofile.config.tck/ConfigProviderTest/testEnvironmentConfigSource/) failure seems to be due to the fact windows is case insensitive and we have two environment variables with [differing case](https://github.com/thorntail/thorntail/blob/master/testsuite/microprofile-tcks/config/pom.xml#L94). Not sure if this is intentional.

Please let me know your review comments regarding the PR.

roguexz:
Rebased and added a separate profile for executing the MP TCK tests for Windows. Hopefully, now the Windows CI build should proceed further.

roguexz:
@Ladicek I am not sure I can fix the rest of the windows failures 😄 .. For starters, I do not have access to a Windows VM .. Can you please review and let me know if I should make additional changes?

Ladicek:
I'm no expert on Windows nor MP Config TCK, but I definitely think that fixing MP Config TCK failures in this PR is _not_ expected. But thanks for doing that :-)

testsuite/microprofile-tcks/config/pom.xml
```diff
@@ -79,31 +79,76 @@
...
+                  <groupId>org.apache.maven.plugins</groupId>
+                  <artifactId>maven-surefire-plugin</artifactId>
+                  <configuration>
+                     <suiteXmlFiles>
+                        <suiteXmlFile>src/test/tck-suite.xml</suiteXmlFile>
+                     </suiteXmlFiles>
+                     <environmentVariables>
+                        <my_int_property>45</my_int_property>
+                        <MY_BOOLEAN_PROPERTY>true</MY_BOOLEAN_PROPERTY>
+                        <MY_STRING_PROPERTY>woohoo</MY_STRING_PROPERTY>
```
Ladicek, line 105: I was wondering if just removing `my_string_property` on Windows would work, but the answer is yes. The MP Config TCK has a special case for Windows here: https://github.com/eclipse/microprofile-config/blob/828ee2fee4ccbef6701943bb96d7be187408fd63/tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPropertyNameMatchingTest.java#L82-L88

testsuite/microprofile-tcks/config/pom.xml
```diff
@@ -79,31 +79,76 @@
...
+                     </environmentVariables>
+                     <!-- This workaround allows to run a single test using "test"
+                        system property -->
+                     <!-- https://issues.apache.org/jira/browse/SUREFIRE-569 -->
+                     <dependenciesToScan>
+                        <dependency>org.eclipse.microprofile.config:microprofile-config-tck</dependency>
+                     </dependenciesToScan>
```
Ladicek, line 112: Not to be changed in this PR, but I wonder if we need this. The [TCK documentation](https://github.com/eclipse/microprofile-config/blob/master/tck/running_the_tck.asciidoc) says that we either need `tck-suite.xml` and `<suiteXmlFile>`, _or_ we need `<dependenciesToScan>`. We have both. I wonder if that could possibly cause issues.

Ladicek (approved):

manovotn:
Win failures are unrelated to changes in this PR, going to merge.

PR #1126, opened 2018-09-24
THORN-2178: Ability to specify FileSystemLayout via System properties

roguexz:
Motivation
----------
I am trying to fix the Gradle Arquillian adapter. As part of the fixes, I wanted to test out the Gradle project in Thorntail examples. What I noticed was that since the project has a pom.xml as well as a build.gradle file, the FileSystemLayout will always select the Maven layout.

This is a corner case and wouldn't happen in most scenarios, because you are either using Gradle or Maven, very rarely will you use both (unless you are building a product like in the case of the examples repository).

This commit attempts to address this problem by providing the ability to customize the implementation on the basis of a System property.

Modifications
-------------
1. Define an expected signature for FileSystemLayout implementations.
2. Implemented the logic for loading custom file system layouts based on the System property.
3. Implemented a test case that validates different scenarios - valid & invalid.

Result
------
The test case proves that the developer can override the layout to use.

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
+++ b/core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2015-2017 Red Hat, Inc, and individual contributors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -15,11 +15,14 @@
  */
 package org.wildfly.swarm.internal;
 
+import java.lang.reflect.Constructor;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Optional;
 import java.util.UUID;
+
+import org.jboss.logging.Logger;
 
 /**
  * Build tool filesystem abstraction for use in IDE:run cases.
@@ -28,6 +31,11 @@ import java.util.UUID;
  * @since 02/08/16
  */
 public abstract class FileSystemLayout {
+
+    /**
+     * System property for overriding the layout class that should be initialized.
+     */
+    public static final String CUSTOM_LAYOUT_CLASS = "thorntail.filesystem.layout.class";
 
     public static final String MAVEN_CMD_LINE_ARGS = "MAVEN_CMD_LINE_ARGS";
 
@@ -76,6 +84,43 @@ public abstract class FileSystemLayout {
      */
     public static FileSystemLayout create(String root) {
 
+        // THORN-2178: Check if a System property has been set to a specific implementation class.
+        String implClassName = System.getProperty(CUSTOM_LAYOUT_CLASS);
+        if (implClassName != null) {
+            implClassName = implClassName.trim();
+            if (!implClassName.isEmpty()) {
+                FileSystemLayout layout = null;
+                // Attempt to load the specified class implementation.
+                ClassLoader loader = Thread.currentThread().getContextClassLoader();
+                try {
+                    Class<?> clazz = loader.loadClass(implClassName);
+                    // Check if clazz is an implementation of FileSystemLayout or not.
+                    if (FileSystemLayout.class.isAssignableFrom(clazz)) {
+                        Class<? extends FileSystemLayout> implClazz = clazz.asSubclass(FileSystemLayout.class);
+                        // Check if we have an appropriate constructor or not.
+                        Constructor<? extends FileSystemLayout> ctor = implClazz.getDeclaredConstructor(String.class);
+                        layout = ctor.newInstance(root);
+                    } else {
+                        String msg = String.format("%s does not subclass %s", implClassName, FileSystemLayout.class.getName());
+                        LOG.warn(SwarmMessages.MESSAGES.invalidFileSystemLayoutProvided(msg));
+                    }
+                } catch (ReflectiveOperationException e) {
+                    Throwable cause = e.getCause();
+                    String msg = String.format("Unable to instantiate layout class (%s) due to: %s", implClassName,
+                                               cause != null ? cause.getMessage() : e.getMessage());
+                    LOG.warn(SwarmMessages.MESSAGES.invalidFileSystemLayoutProvided(msg));
+                    LOG.debug(SwarmMessages.MESSAGES.invalidFileSystemLayoutProvided(msg), e);
+                    throw SwarmMessages.MESSAGES.cannotIdentifyFileSystemLayout(msg);
+                }
+
+                if (layout != null) {
+                    return layout;
+                }
+            } else {
+                LOG.warn(SwarmMessages.MESSAGES.invalidFileSystemLayoutProvided("Implementation class name is empty."));
+            }
+        }
+
         String mavenBuildFile = resolveMavenBuildFileName();
 
         if (Files.exists(Paths.get(root, mavenBuildFile))) {
@@ -101,6 +146,15 @@ public abstract class FileSystemLayout {
         return buildFileName;
     }
 
+    /**
+     * Constructs a new instance of {@code FileSystemLayout} which is rooted at the given path.
+     *
+     * @param path the root path for the layout.
+     */
+    protected FileSystemLayout(String path) {
+        this.rootPath = Paths.get(path);
+    }
+
     public abstract Path getRootPath();
 
     private static final String POM_XML = "pom.xml";
@@ -119,8 +173,11 @@ public abstract class FileSystemLayout {
 
     private static final String BUILD_RESOURCES_MAIN = "build/resources/main";
 
+    private static final Logger LOG = Logger.getLogger("org.wildfly.swarm");
+
     protected static final String TYPE_JAR = "jar";
 
     protected static final String TYPE_WAR = "war";
 
+    protected final Path rootPath;
 }
--- a/core/container/src/test/java/org/wildfly/swarm/internal/FileSystemLayoutTest.java
+++ b/core/container/src/test/java/org/wildfly/swarm/internal/FileSystemLayoutTest.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2018 Red Hat, Inc, and individual contributors.
+ *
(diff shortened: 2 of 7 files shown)


roguexz:
Updated the [examples](https://github.com/roguexz/thorntail-examples/commit/a4b5f522501ec64a5e44dc849f86d3594756d47b) as well. Will raise a PR for that once this is merged.

Ladicek (dismissed):

core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
```diff
@@ -1,4 +1,4 @@
-/**
```
manovotn, line 1: Evil auto-formatting? :)

core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
```diff
@@ -76,6 +79,40 @@ public static FileSystemLayout create() {
...
+                    String msg = String.format("Unable to instantiate layout class (%s) due to %s", implClassName, e.getMessage());
+                    LOG.warn(SwarmMessages.MESSAGES.invalidFileSystemLayoutProvided(msg));
+                    LOG.debug(SwarmMessages.MESSAGES.invalidFileSystemLayoutProvided(msg), e);
+                }
+
+                if (layout != null) {
+                    return layout;
+                }
+            } else {
+                LOG.warn(SwarmMessages.MESSAGES.invalidFileSystemLayoutProvided("Implementation class name is empty."));
```
manovotn, line 120: At this point we know user wants to leverage the property; wouldn't it make more sense to blow up with exception rather then having a warning (e.g. fail fast)?

core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
```diff
@@ -76,6 +79,40 @@ public static FileSystemLayout create() {
      */
     public static FileSystemLayout create(String root) {
 
+        // THORN-2178: Check if a System property has been set to a specific implementation class.
+        String implClassName = System.getProperty(FileSystemLayout.class.getName());
```
manovotn: Since it is no longer WFLY Swarm but rather Thorntail, I think the property name should be changed accordingly.

kenfinnigan:
Is this change really necessary?

What is the likelihood of a project having a pom.xml and build.gradle at the same time?

core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
roguexz: A lot of classes in the codebase use the Javadoc comment for copyright messages. So each file that I touch and I notice the issue, I fix it :)

core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
roguexz: Good point. I did not want to break the existing processes. Should we rather bail when the exception is raised? I log a warning and continue.

core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
roguexz: Thought about it ... then realized that the class itself was internal. This scenario would be very rare and we should be able to use the class name itself. So if the packages get renamed in the future, this would automatically change. WDYT?

roguexz:
@kenfinnigan I agree that the likelihood of this scenario is extremely rare and as such this change isn't required. Here is how I got to this change.

I like the idea of Swarm / Thorntail and strongly believe that it would help me spin up prototypes and micro-services with extreme ease. Hence all my recent contributions.

1. I prefer using Gradle and as such have been working toward improving the Gradle support for Swarm / Thorntail.
2. I am done with my fixes. This includes,
  * Updated Gradle configuration (almost on par with the Maven configuration)
  * Improved communication with the Arquillian adapter
  * Tested against the latest Gradle version (4.10.x)

I decided to run through some "real life examples" of expected workflows that my team or myself would go through.

1. Being able to execute the test cases from within the IDE (THORN-1830)
2. Looking / working with examples (this PR)

I was updating the example projects and that is when I ran in to this issue. It caught me by surprise as the issue occurred in code that I wasn't even looking at!

Apart from the above reasons, in my experience I have always found that if there are multiple adapters (with ranking) applicable to a given scenario, there will always be someone who wants to override the adapter choice. For example, when narrowing down the scope of an issue by using the mock / simplistic adapters. This fix is similar to that.

In any case, please let me know if you would want to see this (internal) feature as part of the product.

kenfinnigan:
I wouldn't see Maven or Gradle as an adapter choice, you're typically using one over the other and not both.

However, if this is going to be included there would need to be some reference to the new property and how to use it in /docs otherwise no one will know about it

roguexz:
@kenfinnigan 
Fair enough. I have currently included it in the [Gradle example project](https://github.com/roguexz/thorntail-examples/blob/master/gradle/build.gradle#L54). I will be pushing the same to the examples repo as well.

Is there a documentation section that talks about FAQs or corner cases? I can include details about it there.

roguexz:
@kenfinnigan Looks like the Gradle plugin related configuration and reference files have been removed. Can I include the documentation as part of my Gradle plugin PR?

kenfinnigan:
The current documentation is located in https://github.com/thorntail/thorntail/tree/master/docs

Not sure a "good place" for where this should go. I wouldn't bring back in all the removed Gradle plugin docs though, as though you've updated it, it's not something that's likely to be updated by us on a regular basis

roguexz:
So then .. should we just leave it in the examples for now? Till there is a change in direction? I could add a README to the examples' project if need be. What do you suggest?

kenfinnigan:
tbh I'm not sure. Maybe just a mention in the examples for now

core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
roguexz: @manovotn The PR needs a resolution on this issue alone now. What do you suggest the default behavior ought to be?

manovotn: > So if the packages get renamed in the future, this would automatically change

That's exactly why I wouldn't bind to to the class itself. You usually want to have these properties documented somewhere and don't want them to change as a side-effect of internal code changes.

So I would probably make a public constant with `thorntail` in its name.

core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
manovotn: Well, you have the use-case for this scenario. Does it make sense for you to continue operating if the property is malformed?
Either way, as long as the behaviour is documented, it should be fine.

manovotn:
+1 for mentioning it in examples

core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
roguexz: Done.

1. Making use of a constant called:`thorntail.filesystem.layout.class`
2. I raise an exception if the class initialization fails.

core/container/src/test/java/org/wildfly/swarm/internal/FileSystemLayoutTest.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    @Test
+    public void verifyCustomLayout() {
+        withProperty("org.wildfly.swarm.internal.CustomFileSystemLayout", layout -> {
+            Assert.assertEquals("Unexpected file system layout.", CustomFileSystemLayout.class, layout.getClass());
+        });
+    }
+
+    /**
+     * Verify that an instantiation exception in the custom layout will result in the default layout being returned.
```
manovotn: Minor thing - the description of the test doesn't fit with what it actually does

manovotn (commented):
Otherwise, looks good to me.
What's your take on this @kenfinnigan?

core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
Ladicek: I know @manovotn asked for this, so this is not really asking for change, just wondering -- should we start introducing the word `thorntail` into thins like configuration keys etc.? So far, everything has been `swarm`, which is legacy, but consistent.

core/container/src/test/java/org/wildfly/swarm/internal/FileSystemLayoutTest.java
roguexz: Good catch! Fixing it.

core/container/src/main/java/org/wildfly/swarm/internal/FileSystemLayout.java
roguexz: I wasn't sure about the naming convention which is why I had decided to go with the classname instead 😉. I can switch to `swarm....` if you prefer.

My preference would be to cut a release where *all* previous keys were *also* referenced as new keys. That way, we do not know have to worry about which keys are considered as old and which are considered as new. But then again, that would also mean that there was a convenient utility in place for getting the preference value from different data sources (YML, SysProps, SysEnv, etc.).

kenfinnigan:
Should be ok

roguexz:
Thank you @kenfinnigan 
Rebasing and addressing the comment from @manovotn. Will update the PR in a couple of hours (running a local build now).

roguexz:
@manovotn Changes incorporated and build is green (at least for the Linux environment). I have updated the [examples](https://github.com/roguexz/thorntail-examples/commit/e42c3e58dac1a9e3fdd828982258552ceaa23fb6) as well and verified that the changes are working as expected. Will raise that PR after this is merged.

Do let me know if you have further comments about this PR.

manovotn (approved):

PR #1141, opened 2018-10-08
THORN-2185: Change config keys to 'thorntail.'

kenfinnigan:

-----

Motivation
----------
It's confusing to have a root config key of 'swarm' when the project is named Thorntail.

Modifications
-------------
Modified config handling to convert 'swarm' to 'thorntail', for backwards compatibility, and updated all config keys to start with 'thorntail'.

Result
------
Config keys now start with 'thorntail', but 'swarm' still works as well.


Ladicek:
I've noticed that a top-level `swarm` key in YAML is automatically transformed to `thorntail`. Are we doing the same for system properties? I.e., will `swarm.*` system properties still work?

(One option how to achieve that is to scan all system properties and clone all that begin with `swarm.*` to a `thorntail.*` equivalent, somewhere soon enough in the boot process. But I guess there are other options, too.)

kenfinnigan:
Pretty sure env vars are already mapped, as they're sucked into ConfigKeys which does the rename from `swarm` to `thorntail`

Ladicek:
I totally forgot to say that I consider this pretty risky. For the system properties, we also access a lot of them through other means than the config system, often simply by `System.getProperty`.

kenfinnigan:
All those usages should also have been adjusted to use "thorntail"

Ladicek:
Yea, and if my understanding is correct, those places will no longer accept the `swarm.*` variant. As in, we won't provide compatibility. That's my concern.

PR #1142, opened 2018-10-08
4.x: flyway module

jonevn:

-----


kg6zvp:
Is this a WIP PR?

jonevn:
I added this as a POC, not knowing if it is something that should be included. Saw that it was included in 2.x version. A bunch of TODO:s are included here. Would be nice to get answers to them during this pull request.

core/flyway/src/main/java/io/thorntail/flyway/impl/FlywayMessages.java
```diff
@@ -0,0 +1,42 @@
...
+
+import io.thorntail.logging.impl.LoggingUtil;
+
+@MessageLogger(projectCode = LoggingUtil.CODE, length = 6)
+public interface FlywayMessages {
+
+	FlywayMessages MESSAGES = Logger.getMessageLogger(FlywayMessages.class, LoggingUtil.loggerCategory("flyway"));
+
+	// TODO: Move to kernel?
+	int OFFSET = 90000;
```
kg6zvp: Definitely move to kernel. :+1:

core/flyway/src/main/java/io/thorntail/flyway/impl/FlywayMigrator.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.flywaydb.core.Flyway;
+
+import io.thorntail.events.LifecycleEvent;
+
+@ApplicationScoped
+public class FlywayMigrator {
+
+	@Inject
+	private FlywayConfiguration flywayConfiguration;
```
kg6zvp, line 15: Any reason this is marked as private?

jonevn: No, not really language-wise. Just comes as a default when I write =)

kg6zvp: I would recommend marking it as package private instead, for testing if nothing else.

Ladicek:
Somewhat tangential to this PR, but I wonder if this really belongs to the core project. My understanding is that with v4, the core project is meant to be very small and focused, and extensions like this should live externally. That is also much easier to achieve -- with v2, creating out-of-repo fractions is pretty much impossible, but with the pure CDI nature of v4, it's crazy easy. I think we should have a repo / GitHub org `thorntail-contrib` or something like that, where extensions like this could live and be maintained separately.

core/flyway/src/main/java/io/thorntail/flyway/impl/FlywayConfiguration.java
```diff
@@ -39,22 +42,22 @@ public void init() {
 		try {
 			dataSourceId = config.getValue(DATASOURCE_PROPERTY, String.class);
 		} catch (Exception e) {
-			FlywayMessages.MESSAGES.dataSourceIdNotConfigured();
+			flywayMessages.dataSourceIdNotConfigured();
```
kg6zvp, line 45: This change does not follow conventions used in the project.

core/flyway/src/test/java/io/thorntail/flyway/impl/FlywayConfigurationTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.eclipse.microprofile.config.Config;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import io.thorntail.datasources.impl.DataSourceRegistry;
+
+@RunWith(MockitoJUnitRunner.class)
```
kg6zvp, line 21: Unit tests are good, but this project seems to favor integration tests with the built in Thorntail runner.

core/kernel/src/main/java/io/thorntail/logging/impl/MessageOffsets.java
```diff
@@ -4,23 +4,24 @@
  * @author Ken Finnigan
  */
 public interface MessageOffsets {
```
kg6zvp, line 6: Please use `git add -p` instead of `git add .` in future.

kg6zvp (commented):
I agree with @Ladicek that this seems like an external/community fraction. Docs are also missing.

core/kernel/src/main/java/io/thorntail/logging/impl/MessageOffsets.java
jonevn: Yeah, my bad

core/flyway/src/main/java/io/thorntail/flyway/impl/FlywayConfiguration.java
jonevn: This change was only made for unit-tests to function and in someway to challenge the aforementioned conventions =)

jonevn:
I can totally buy the point made by @Ladicek. So reviewing should be stopped and this PR could be closed. If the decision is to include things like this in core, then I could make a new PR (with code in better shape)

core/kernel/src/main/java/io/thorntail/logging/impl/MessageOffsets.java
kg6zvp: No worries. It happens, we've all been there. :)

Ladicek:
I don't necessarily think a review needs to stop. If there's a consensus on establishing a `thorntail-contrib` repo/org (anyone want to start a discussion on the Google Group?), we can move the code there, and we would probably want to do a review anyway.

kg6zvp:
@Ladicek I think what might be even better than a community repo would be to establish something like a marketplace for community add-ons/modules. I'll start a discussion on the google group.

michalszynkiewicz:
Adding a hold label because there seems to be no consensus as to what to do with the PR

Ladicek:
I'm closing this, as v4 is discontinued.

PR #1144, opened 2018-10-10
[THORN-2116] Mining the old 3.1.4.1 (System properties) and 4.2.2 (Datasources) sections

sberyozkin:


sberyozkin:
As far as https://wildfly-swarm.gitbooks.io/wildfly-swarm-users-guide/configuration_properties.html is concerned, I haven't added a dedicated 'Undertow' subsection, I guess most of the Thorntail endpoints are HTTP endpoints anyway, and I dropped few properties which I've never seen being used (the info about them is still in the linked java file)

Ladicek:
LGTM, but should wait for @tradej review :-)

tradej:
I will create a PR with suggested changes against @sberyozkin's fork.

tradej (changes requested):
PR with suggested changes filed at https://github.com/sberyozkin/thorntail/pull/1.

tradej (approved):

docs/modules/ref_commonly-used-system-properties.adoc
```diff
@@ -0,0 +1,98 @@
...
+|===
+|Default
+|
+|===
+--
+
+With the H2, MySQL, and Postgres database fractions,  use the following properties to configure the datasource:
+
+.Datasource-related system properties
+swarm.ds.name:: The name of the datasource
```
kenfinnigan, line 64: I'm not sure if we want to be promoting these shortcuts, as opposed to the actual DS config that is required. I say that because these settings alone aren't enough to configure a DataSource, it also needs to define the JDBC driver, etc.

docs/modules/ref_commonly-used-system-properties.adoc
```diff
@@ -0,0 +1,98 @@
+
+[id='commonly-used-system-properties_{context}']
+= Commonly used system properties
+
+This is a non-exhaustive list of system properties you are likely to use in your application:
+
+.General system properties
+swarm.bind.address:: The interface to bind servers
```
kenfinnigan, line 8: These need to use the new "thorntail" config key

PR #1153, opened 2018-10-17
[THORN-2160] Including JWE key encryption alias as a header by default

sberyozkin:


sberyozkin:
Jose4J examples page says it is a 'polite thing to do' to report a (public) JWE encryption key alias :-), hence, as we agreed with David, that should be done by default

sberyozkin:
Michal or Ladislav, please have a quick look, these are just minor updates across two tested implementations (the default one and Jose4j)

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
```diff
@@ -257,4 +267,11 @@ public String contentEncryptionAlgorithm() {
     @Configurable("thorntail.jose.encryption.key.alias")
     @AttributeDocumentation("Alias to the encryption key entry in the keystore")
     private Defaultable<String> encryptionKeyAlias = string(DEFAULT_KEY_ALIAS);
+
+    /**
+     * Include Encryption Key Alias as Header.
+     */
+    @Configurable("thorntail.jose.encryption.include.alias")
+    @AttributeDocumentation("Include the encryption key alias as a 'kid' header: true (default)")
```
Ladicek: What "header"? It's not an HTTP header, right?

Also, I'd replace `: true (default)` with ` (defaults to true)`.

Ladicek:
I don't quite understand what "header" are we talking about, but if you clarify that in the attribute documentation, I think that would be enough.

Otherwise LGTM.

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
dvddhln: It is the JWE(JSON Web Encryption) header it is referring to. Described in detail here
https://tools.ietf.org/html/rfc7516

Ladicek: That RFC uses a term "JOSE header", so if this read "Include the encryption key alias as the JOSE `kid` header", or something like that, that would be great IMHO.

sberyozkin: Sure, np, will update shortly

dvddhln: Sounds good to me!

Ladicek (approved):

sberyozkin:
retest this please

sberyozkin:
Lets see what happens to https://ci.wildfly-swarm.io/job/thorntail-pull-request-linux/717/ :-)

sberyozkin:
Jose tests are green

Issue #1155, opened 2018-10-20
Add postgres as a driver in jdbc module


johnaohara:
There are a number of drivers supported ootb in the jdbc module.  One jdbc driver that is not present is postgres

I am proposing including postgresql as a shortname as the postgresql driver in jdbc module

PR #1156, opened 2018-10-20
Add postgres as a driver in jdbc module #1155

johnaohara:
Add postgres as a driver in jdbc module #1155

provide postgres jdbc driver with short name postgresql


kg6zvp (approved):
Looks good to me.

PR #1158, opened 2018-10-22
THORN-2195: Use the correct TCCL when health checks are obtained.

mkouba:

### DIFF
--- a/fractions/microprofile/microprofile-health/src/main/java/org/wildfly/swarm/microprofile/health/api/Monitor.java
+++ b/fractions/microprofile/microprofile-health/src/main/java/org/wildfly/swarm/microprofile/health/api/Monitor.java
@@ -54,6 +54,12 @@ public interface Monitor {
 
     Object getHealthReporter();
 
+    void unregisterContextClassLoader();
+
+    void registerContextClassLoader(ClassLoader classLoader);
+
+    ClassLoader getContextClassLoader();
+
     List<HealthMetaData> getHealthURIs();
 
     Optional<SecurityRealm> getSecurityRealm();
--- a/fractions/microprofile/microprofile-health/module.conf
+++ b/fractions/microprofile/microprofile-health/module.conf
@@ -24,3 +24,4 @@ org.jboss.as.domain-http-interface
 
 org.jboss.logging
 io.undertow.servlet
+org.wildfly.security.elytron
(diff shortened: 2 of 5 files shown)


fractions/microprofile/microprofile-health/src/main/java/org/wildfly/swarm/microprofile/health/api/Monitor.java
```diff
@@ -54,6 +54,12 @@ static Monitor lookup() throws NamingException {
 
     Object getHealthReporter();
 
+    void unregisterContextClassLoader();
+
+    void registerContextClassLoader(ClassLoader classLoader);
+
+    ClassLoader getContextClassLoader();
+
```
Ladicek, line 62: Just wondering if we want to make this part of the public API?

Ladicek:
Otherwise LGTM.

fractions/microprofile/microprofile-health/src/main/java/org/wildfly/swarm/microprofile/health/api/Monitor.java
sberyozkin: That should be fine, all that class is there for is to facilitate the interaction between various microprofile health parts as far as I understand, not really meant for the public use. May be it should be noted in the docs somehow

Ladicek: I agree it's not _meant_ for public use, but IIUC, it _is_ exposed publicly.

mkouba: Well, all public classes/methods/etc. are exposed and can be considered a part of the API ;-)

sberyozkin: I don't mind either way :-). I recall looking at that class and I'm not sure any of its methods could be meaningfully used by the non Thorntail code...

Ladicek: We actually have a delineation between public API and private details, e.g. the `runtime` packages are not considered public API. The delineation is defined by the script that generates Javadoc :-)

Ladicek: But @sberyozkin is right, this API isn't really all that useful and is "here be dragons" already :-)

Ladicek (approved):

PR #1164, opened 2018-10-31
THORN-2205: the "swarm.*" system properties no longer work

Ladicek:
Motivation
----------
In https://github.com/thorntail/thorntail/pull/1141, all the
configuration properties were renamed from `swarm.*` to
`thorntail.*`. Compatibility is provided for the configuration
system (YAML files and system properties that tie into it), but
not for system properties accessed directly (`System.getProperty()`).

Modifications
-------------
Add a piece of code near the beginning of the main Swarm lifecycle
and also near the beginning of the the Arquillian container lifecycle
that converts all `swarm.*` system properties to `thorntail.*`. Rest
of the code can remain agnostic of the compatibility concern and just
use `thorntail.*`.

Result
------
The old `swarm.*` system properties continue to work.
There's one exception: the Maven plugin. The way Maven injects
properties into the plugins make is impossible to do a similarly
localized change; the change would have to touch all places
that previously used `swarm` and now use `thorntail`.

-----

### DIFF
--- a/core/container/src/main/java/org/wildfly/swarm/Swarm.java
+++ b/core/container/src/main/java/org/wildfly/swarm/Swarm.java
@@ -76,6 +76,7 @@ import org.wildfly.swarm.bootstrap.logging.BootstrapLogger;
 import org.wildfly.swarm.bootstrap.modules.BootModuleLoader;
 import org.wildfly.swarm.bootstrap.performance.Performance;
 import org.wildfly.swarm.bootstrap.util.BootstrapProperties;
+import org.wildfly.swarm.bootstrap.util.BootstrapUtil;
 import org.wildfly.swarm.cli.CommandLine;
 import org.wildfly.swarm.container.DeploymentException;
 import org.wildfly.swarm.container.config.ClassLoaderConfigLocator;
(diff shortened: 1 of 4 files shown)


Ladicek:
Unfortunately, it seems impossible to me to do a similarly localized change for the Maven plugin, so compatibility remains broken on that front. Despite that, I think this is a good improvement. We can keep saying that there's a breaking change, but this will serve as a safety net.

core/container/src/main/java/org/wildfly/swarm/Swarm.java
```diff
@@ -256,6 +258,17 @@ public Swarm(boolean debugBootstrap, Properties properties, Map<String, String>
         this.isConstructing = false;
     }
 
+    private static void convertSwarmSystemPropertiesToThorntail() {
```
michalszynkiewicz: both `core` and `arquillian-adapter` modules seem to depend on `bootstrap`. Can we have this method in one place only?

Ladicek: I tried in `container` and hit a classloader wall. Not sure about `arquillian`, will try.

Ladicek: Looking at `BootstrapModuleFinder`, I realized that `org/wildfly/swarm/bootstrap/util` is open, so I put the method into `BootstrapUtil` and voila, it's accessible from everywhere. Thanks for insisting on it :-)

michalszynkiewicz: cool :)

michalszynkiewicz (approved):

PR #1166, opened 2018-11-07
[THORN-2231] Create a jsonp fraction test

sberyozkin:


sberyozkin:
I will then check how it will do on the WF14 branch :-)

testsuite/testsuite-jsonp/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>io.thorntail</groupId>
+      <artifactId>jsonp</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>javax.json</groupId>
+      <artifactId>javax.json-api</artifactId>
+      <version>${version.javax.json.api}</version>
```
Ladicek: Isn't this brought in transitively by the `jsonp` fraction?

testsuite/testsuite-jsonp/src/test/java/org/wildfly/swarm/jsonp/JavaxJsonpTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.swarm.undertow.WARArchive;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+@RunWith(Arquillian.class)
+public class JavaxJsonpTest {
+
+    @Deployment
```
Ladicek: I guess `@DefaultDeployment` should work?

Ladicek:
Otherwise LGTM.

testsuite/testsuite-jsonp/pom.xml
sberyozkin: The servlet class was not compiling, the glassfish dep is in the test scope - I thought it would test better that the fraction's module.conf is effective ? May be I can change the scope of the glassfish dep to the normal one ?

testsuite/testsuite-jsonp/src/test/java/org/wildfly/swarm/jsonp/JavaxJsonpTest.java
sberyozkin: I thought so, I copied the default deployment fragment from some other test, but the servlet was just not available... I was keeping getting (HTTP) MethodNotAllowed - which I assume was caused by the default servlet (not the test one) rejecting POSTs

testsuite/testsuite-jsonp/pom.xml
sberyozkin: I've just recompiled back to the WF14 branch, let me recompile back and check what happens if I just exclude glassfish and javax.json (as far as the servlet is concerned)

Ladicek: I mean, the `jsonp` fraction has this in the POM:

```xml
    <dependency>
      <groupId>org.glassfish</groupId>
      <artifactId>javax.json</artifactId>
    </dependency>
```

That, in turn, has

```xml
        <dependency>
            <groupId>javax.json</groupId>
            <artifactId>javax.json-api</artifactId>
        </dependency>
```

So I'd assume that `javax.json:javax.json-api` would be available transitively (in the Maven sense) just by including the `jsonp` fraction. That's how other fractions work, at least. If not, that sounds like a bug.

sberyozkin:
@Ladicek, yes, just having a fraction dependency is OK, I think adding the test glassfish dependnecy was a mistake, thanks

testsuite/testsuite-jsonp/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+
+  <name>Test Suite: JSONP</name>
+  <description>Test Suite: JSONP</description>
+
+  <packaging>jar</packaging>
+  
+  <properties>
+    <version.javax.json.api>1.0</version.javax.json.api>
+    <version.org.glassfish.javax.json>1.0.4</version.org.glassfish.javax.json>
+  </properties>
```
Ladicek: Remove these then?

sberyozkin: Done :-)

Ladicek (dismissed):

sberyozkin:
@Ladicek, sorry, I've just confirmed that the default deployment works after all, I honestly no idea it did not work originally, please approve again

Ladicek (approved):

Ladicek:
Great, thanks!

PR #1171, opened 2018-11-13
THORN-2241: fix build so that all artifacts are properly built and SNAPSHOTs don't have to be downloaded when generating BOMs

Ladicek:
Motivation
----------
Some artifacts (Camel components) don't get built properly,
because of missing feature pack ZIP dependencies. Some artifacts
(Camel components and hollow JARs) that are built properly don't
generate source JARs, because they don't have actual sources.

Both these problems cause the BOM generation process to download
`SNAPSHOT`s of these artifacts from external locations. That is wrong
and also causes huge slowdowns. It's also pretty easy to fix.

Modifications
-------------
Added all 3 WildFly feature packs as dependencies to all Camel
components (via the parent POM).

Fixed erroneous module names in `module.conf` files in Camel
components.

Configured the Maven source plugin to force source JAR generation.

Result
------
Build is now correct and also much faster.

-----

### DIFF
--- a/fractions/camel/components/pom.xml
+++ b/fractions/camel/components/pom.xml
@@ -47,6 +47,43 @@
           </exclusion>
         </exclusions>
       </dependency>
+
+      <dependency>
+        <groupId>org.wildfly.core</groupId>
+        <artifactId>wildfly-core-feature-pack</artifactId>
+        <scope>provided</scope>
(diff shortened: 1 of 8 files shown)


kenfinnigan (approved):

Ladicek:
Windows build failed because of disk space.

PR #1175, opened 2018-11-23
[THORN-2156] Add support for accepting verification and decryption ke…

dvddhln:
…y aliases

Motivation
----------
Provide support for accepting verification and decryption key aliases in
configuration.

Modifications
-------------
Add logic for verification and decryption with tests.

Result
------
This will allow for accepting and using incoming headers when performing
verification or decryption.

-----


manovotn:
test this please

manovotn:
ok to test

manovotn (approved):

sberyozkin:
@dvddhln Hi David, it looks very neat, thanks :-) 
I have only 2 minor suggestions: 
* if the kid can be accepted but the incoming headers have no kid set, then continue with the default kid selection; 
* For the tests to really show that the accept alias works, the default alias value should be null. Right now it is set to the "server", but it should really be not set to some default value, I set it initially just to make things simpler fotr a start; you'd only need to fix few tests (which work with JKS stores), set the alias to 'server' in yaml, once you remove the default key alias value.

Update: I guess we can just keep the default alias, I've realized it is not really relevant to this PR because it works with the JWK store where no 'server' kid is available :-). So please ignore my 2nd suggestion which would only apply if there was a an accept alias test for the JKS store. 

Update: This morning I've realized that the 1st suggestion does not really make much sense either. Having a production situation where the kid can be either the locally specified kid or, now, and then, the incoming kid, is a bit unrealistic. We can always tweak it later if a policy like that would be of interest  

Thanks, Sergey

sberyozkin (approved):

PR #1178, opened 2018-12-03
Gradle Arquillian Adapter Improvements

roguexz:
* Support for Gradle 5.0
* Handle the scenario where the IdeaModule marks a library dependency as "provided", because it is being provided by a project-dependency.

-----

### DIFF
--- a/arquillian/gradle-adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/gradle/GradleDependencyAdapter.java
+++ b/arquillian/gradle-adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/gradle/GradleDependencyAdapter.java
@@ -15,19 +15,28 @@
  */
 package org.wildfly.swarm.arquillian.adapter.gradle;
 
+import java.io.File;
 import java.nio.file.Path;
-import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 
 import org.gradle.tooling.GradleConnector;
 import org.gradle.tooling.ProjectConnection;
 import org.gradle.tooling.model.ExternalDependency;
 import org.gradle.tooling.model.GradleModuleVersion;
 import org.gradle.tooling.model.idea.IdeaModule;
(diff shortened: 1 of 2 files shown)


arquillian/gradle-adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/gradle/GradleDependencyAdapter.java
```diff
@@ -15,20 +15,29 @@
  */
 package org.wildfly.swarm.arquillian.adapter.gradle;
 
+import java.io.File;
```
roguexz, line 18: Please review the [original file](https://github.com/roguexz/thorntail/blob/master/arquillian/gradle-adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/gradle/GradleDependencyAdapter.java) to view the details of the approach better.

build-parent/pom.xml
```diff
@@ -76,7 +76,7 @@
 
     <version.flyway>4.2.0</version.flyway>
 
-    <version.gradle>3.0</version.gradle>
+    <version.gradle>3.1</version.gradle>
```
roguexz, line 79: By bumping to 3.1, I am able to make use of an API (not available in 3.0) that is available all the way till 5+ version. The original API that I was making use of was marked as deprecated in 3.0 and finally removed in 5.0.

roguexz:
@Ladicek I can confirm that the changes work with Gradle 4.10.2 & 5.0 both from the command line as well as within the IDE. I have also verified that it works well with [composite builds](https://docs.gradle.org/current/userguide/composite_builds.html) too. Please review and provide your comments.

Ladicek (approved):
I don't know much about Gradle or our Gradle plugin anymore, but there's nothing obviously wrong here :-)

roguexz:
Thank you @Ladicek. Once I am done with all the performance changes, I shall try convincing you to switch to Gradle :) .. Till then, I will keep hacking and using it.

Ladicek:
I was playing with Gradle back in the 1.0-RCx days, but learned Maven a lot better since then. And Gradle changed massively as well. I need to try the Gradle Kotlin integration, that might be a very nice thing :-)

kenfinnigan (approved):

PR #1179, opened 2018-12-04
THORN-2277: tracing filter registration no longer overrides other ones

michalszynkiewicz:
Motivation
----------
OpenTracing, when mixed with JWT, made JWT filters not registered.
The cause was that `TracingInstaller` overwrote RESTEasy providers.
It resulted in an application being unsecured.

Modifications
-------------
Tracing provider is now appended to the already defined providers.
Also, a test that uses the whole microprofile dependency and checks if
MP JWT works was added in a newly created module - testsuite-microprofile

Result
------
MP OpenTracing and JWT can be safely used together now.

-----


kenfinnigan (approved):

Ladicek:
It seems there's quite a bit of code duplication in the test suite part :-)

michalszynkiewicz:
@Ladicek yep, do we have some module that is shared between all testsuites?

Ladicek:
I don't think so, but we probably should.

PR #1180, opened 2018-12-04
THORN-2264: Runner: pass classpath in env variable

michalszynkiewicz:
Motivation
----------
On Windows, Thorntail Runner was failing with a project having many dependencies.
The cause was the fact that classpath was passed to FatJarBuilder
with -cp in the command.

Modifications
-------------
The classpath is passed as an environment variable now

Result
------
The Runner no longer fails on a big project
Tested in IntelliJ IDEA and Eclipse on Windows 10

-----

### DIFF
--- a/thorntail-runner/src/main/java/org/wildfly/swarm/runner/Runner.java
+++ b/thorntail-runner/src/main/java/org/wildfly/swarm/runner/Runner.java
@@ -27,6 +27,7 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


Ladicek:
Environment variables on Windows are limited as well, see https://blogs.msdn.microsoft.com/oldnewthing/20100203-00/?p=15083. I think in the Runner case, we should probably set the classpath in the manifest of the (uber)jar.

michalszynkiewicz:
@Ladicek Could you elaborate how you see it working with setting classpath in a manifest?
The problem is in passing the classpath to FatJarBuilder

Ladicek:
Ah, I see. You'd have to create a synthetic JAR with a manifest that contains the classpath.

Ladicek:
Now thinking about it, my suggestion is probably a premature fix for problem that we don't have (yet).

michalszynkiewicz:
@Ladicek it's a good catch. A problem with doing much stuff in the Runner process is that if we execute anything using java util logging, jboss logging will fail and so will Thorntail.
I'm thinking of passing the full classpath in some file and just the elements required for FatJarBuilder in the env variable. But it would require figuring out what's required for the builder.
WDYT of using this PR for 2.3.0 and implementing something better later (if there is time or if people hit this problem ;) )

mkouba:
@michalszynkiewicz is this the PR I was helping to verify on a windows machine?

michalszynkiewicz:
@mkouba yes

mkouba (approved):

Ladicek:
Sure, we can always improve later!

PR #1181, opened 2018-12-06
[THORN-2126] upgrade to WildFly 14

Ladicek:


manovotn:
@Ladicek great job with WFLY-11373 workaround :)

Ladicek:
Thanks @manovotn, though I kinda hoped I'd find a `Literal` for `@Resource` somewhere ;-)

Ladicek:
Also could someone cleanup the disk on the Windows Jenkins agent? :-)

sberyozkin (commented):
All of these changes look good to me. @kenfinnigan, please provide your review

docs/howto/pom.xml
```diff
@@ -88,18 +91,6 @@
       </resource>
     </resources>
     <plugins>
-      <plugin>
```
kenfinnigan, line 91: Why was this removed?

fractions/javaee/jpa/pom.xml
```diff
@@ -51,6 +51,11 @@
       <groupId>io.thorntail</groupId>
       <artifactId>ee</artifactId>
     </dependency>
+    <!-- WF14 -->
```
kenfinnigan: There's a hard dependency on Infinispan fraction for JPA? That's not good

build-parent/pom.xml
```diff
@@ -248,6 +252,42 @@
         <version>${version.org.jboss.shrinkwrap.descriptors}</version>
       </dependency>
 
+      <dependency>
```
kenfinnigan, line 255: Why are all these aether dependencies needed here? I didn't see them added anywhere

fractions/javaee/jpa/pom.xml
sberyozkin: This is because of https://github.com/wildfly/wildfly/blob/14.0.1.Final/feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/jpa/main/module.xml#L49 I believe

docs/howto/pom.xml
Ladicek: Because each howto defines `thorntail-maven-plugin` on its own -- unfortunately, using a different `<id>` of the `<execution>`. So the Thorntail Maven plugin's `package` goal used to be invoked twice.

build-parent/pom.xml
Ladicek: Hmm, I don't recall adding this. I'll check.

fractions/javaee/jpa/pom.xml
Ladicek: The Hibernate+Infinispan integration has moved from Hibernate to Infinispan, that might be the reason. But I'll look deeper.

Ladicek: Unfortunately it seems it really is required. Otherwise I'm getting

```
    "WFLYCTL0412: Required services that are not installed:" => ["org.wildfly.clustering.infinispan.cache-configuration.hibernate.entity"],
    "WFLYCTL0180: Services with missing/unavailable dependencies" => [
        "jboss.persistenceunit.\"CdiJpaJtaTest.jar#MyPU\".__FIRST_PHASE__ is missing [org.wildfly.clustering.infinispan.cache-configuration.hibernate.entity]",
        "jboss.persistenceunit.\"CdiJpaJtaTest.jar#MyPU\" is missing [org.wildfly.clustering.infinispan.cache-configuration.hibernate.entity]"
    ]
```

That is, the persistence unit service (from the JPA subsystem) depends on the Hibernate cache services (from the Infinispan subsystem).

Not sure how it worked before.

build-parent/pom.xml
Ladicek: @sberyozkin It seems you added this, do you recall why?

Ladicek: Got it. Previously, Eclipse Aether was managed in `wildfly-core-parent` (which we import in `<dependencyManagement>`), but that's no longer the case. So we have to manage it ourselves.

Ladicek:
Pushed a commit that addresses the remaining WF14 pieces. Unfortunately, it seems JPA having a hard dependency on Infinispan is now needed.

build-parent/pom.xml
sberyozkin: Yes, that is exactly why I added them, sorry, forgot I did it

kenfinnigan: Ok, thanks for clarification

fractions/javaee/jpa/pom.xml
kenfinnigan: Well that's going to blow up uber jar sizes.

Oh well

docs/howto/pom.xml
kenfinnigan: Makes sense, thanks

kenfinnigan (dismissed):

fractions/javaee/jpa/pom.xml
Ladicek: It does :-( The `jpa-jaxrs-cdi-jta` example has 110 MB with 2.2.1 and 126 MB now.

Ladicek:
Linux build is green, yay! It would be great if we could get the disk cleaned up on the Windows agent, as getting a Windows build would be good for a change of this size.

kenfinnigan:
retest this please

kenfinnigan (approved):

PR #1183, opened 2018-12-06
[THORN-2202] Support for inlined JWK sets

sberyozkin:


sberyozkin:
@michalszynkiewicz Hi Michal, can you please have a look as well, this patch is a simple enhancement for the Jose fraction as suggested by David, it will allow users who work with JWK users inline the JWK sets inside the YAML config

dvddhln (commented):
Hi Sergey, this looks good to me! Cheers for adding this feature!

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
```diff
@@ -200,13 +209,13 @@ public String contentEncryptionAlgorithm() {
     }
 
 
-    public JoseConfiguration setAcceptDecryptionAlias(boolean acceptDecryptionAlias) {
-        this.acceptDecryptionAlias.set(acceptDecryptionAlias);
+    public JoseConfiguration setAcceptEncryptionAlias(boolean acceptEncryptionAlias) {
```
michalszynkiewicz, line 212: what's the reason of the decryption -> encryption change

testsuite/testsuite-jose-jose4j/src/test/resources/project-jwk-inline-store.yml
```diff
@@ -0,0 +1,31 @@
+swarm:
+  jose:
+    keystore:
+      path: inline
+      type: jwk
+      jwkset: |
+        {
+          "keys":
```
michalszynkiewicz, line 8: to be sure I understand it, there's one key for encryption and one for signing and on a specific operation the proper key is selected based on `alias == kid`?

sberyozkin: Hi Michal, yes, exactly. In some cases this 'kid' (key identifier) will come from the incoming secured payload but that is tested in a different test case

fractions/jose/src/main/java/org/wildfly/swarm/jose/JoseConfiguration.java
sberyozkin: Sure, we have [discussed it](https://issues.jboss.org/browse/THORN-2156?focusedCommentId=13669229&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13669229) with David who originally created the earlier patch for the accept alias/kid user case. That case is about accepting a kid (alias) which was used for either signing or encryption from the incoming payload, as opposed to configuring this kid in YAML.

This change is only a cosmetic one, make it more consistent with the signature case (where we do not say acceptVerification alias): when the client encrypted the payload and included the kid/alias of the key, it was an encryption key alias, and the above property says accept the encryption key alias for the purpose of decryption

michalszynkiewicz (approved):

PR #1191, opened 2018-12-25
THORN-2293: Arquillian tests require the bootstrap module added as a compile time dependency

roguexz:
Motivation
----------
The "findWildFlySwarmBootstrapJar" method in org.wildfly.swarm.tools.DependencyManager searches for the bootstrap jar in the "COMPILE" dependencies only.

When running Arquillian based test cases, it is not necessary that every project has an explicit "compile-time" dependency on the bootstrap module.

E.g.,
In a multi-module project, I might have modules that generate stand-alone libraries, but are still being tested using the Thorntail integration.

Modifications
-------------
Changed the DependencyManager to look for bootstrap jar in test dependencies as well.

Result
------
Defining the Arquillian dependency as a test scope is sufficient to get the test cases running.

-----

### DIFF
--- a/tools/src/main/java/org/wildfly/swarm/tools/DependencyManager.java
+++ b/tools/src/main/java/org/wildfly/swarm/tools/DependencyManager.java
@@ -70,7 +70,7 @@ public class DependencyManager implements ResolvedDependencies {
 
     @Override
     public ArtifactSpec findWildFlySwarmBootstrapJar() {
-        return findArtifact(FractionDescriptor.THORNTAIL_GROUP_ID, WILDFLY_SWARM_BOOTSTRAP_ARTIFACT_ID, null, JAR, null, false);
+        return findArtifact(FractionDescriptor.THORNTAIL_GROUP_ID, WILDFLY_SWARM_BOOTSTRAP_ARTIFACT_ID, null, JAR, null, true);
     }
 
     @Override


roguexz:
Alternatively, what do you guys think of the following change,

1. BuildTool has a new property called `testExecutionMode: boolean`
2. BuildTool passes this information across to the `DependencyManager` so that the decision to search for dependencies within the "test" scope is made optional and requested only on-demand.
3. The Arquillian `UberjarSimpleContainer` can construct the appropriate `BuildTool` instance.

If need be, `1` & `2` can be coded to be immutable.

Just an idea .. i.e., if the overall request makes sense.

kenfinnigan:
`bootstrap` isn't needed as a direct compile dependency, but it should be present transitively as a compile dependency of something else.

It also appears this change broke the MP TCKs

roguexz:
@kenfinnigan I don't expect this PR to be merged as is, because even I am not convinced of the changes. I wanted to start a discussion about it.

Based on your explanation, the only time this issue would show up is if I have a library module that is executing tests using Thorntail+Arquillian *and* this library does not have any direct dependencies on Thorntail modules.

Would that be a correct statement?

kenfinnigan:
I believe so yes.

roguexz:
Thank you Ken. We can close this PR for now.

As part of the Gradle enhancements, if I find a better way of dealing with this scenario, I will definitely share.

We can continue further discussions on the issue itself.

PR #1192, opened 2018-12-27
Gradle Plugin Enhancements

roguexz:
* Introduced a new plugin 'thorntail-arquillian' which is useful for library-only type of projects.
* Make use of 'thorntail-arquillian' plugin to get accurate dependency information (falling back to the IdeaModel if the plugin was not applied).
* For Arquillian tests, handle custom developer-defined configurations.
* Automatically add the dependency on 'Gradle Tooling API' when the 'thorntail-arquillian' plugin is added.
* Automatically apply the 'thorntail-arquillian' plugin if the original 'thorntail' plugin is applied.
* Delete class files that were never really used and not required anymore.

-----


sberyozkin:
Hi, can you please confirm the Gradle examples are OK after this change ?

roguexz:
@sberyozkin I am going to be raising a PR to the examples project as well to show the different configuration modes. Let me do that and share the details here.

roguexz:
@sberyozkin I have updated the [Gradle Examples](https://github.com/thorntail/thorntail-examples/pull/199) and ensured that they are running from the IDE (IntelliJ) as well as from the command line (both tests as well as standalone jar).

There was a bug in my current PR. I have fixed it and also fixed the reason for not detecting the bug in the examples project.

Please review and provide your comments.

roguexz:
@sberyozkin Please let me know if you need additional information about the PR. The other conversation ([1](https://groups.google.com/forum/#!topic/thorntail/vC4TQMUhhB0) & [2](https://github.com/thorntail/thorntail/pull/1191)) that I am initiating are around providing a more wholistic integration with Gradle and to get better integration with [composite projects](https://docs.gradle.org/current/userguide/composite_builds.html).

arquillian/gradle-adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/gradle/GradleDependencyAdapter.java
```diff
@@ -54,57 +58,21 @@ public GradleDependencyAdapter(Path projectDir) {
...
-                        upstreamProjects.forEach(moduleName -> testScopedLibs.addAll(
-                                ARTIFACT_DEPS_OF_PRJ.getOrDefault(moduleName, emptyMap())
-                                        .getOrDefault(DEP_SCOPE_TEST, emptySet())));
-                        providedScopeDeps.stream().filter(testScopedLibs::contains).forEach(collectedDependencies::add);
-                    }
-
-                });
-
-                collectedDependencies.forEach(declaredDependencies::add);
+                // 1. Attempt to fetch the dependencies via the Thorntail model.
```
roguexz, line 67: Moved the above logic in to a method of its own `getDependenciesViaIdeaModel`. I first try to load the details from the `ThorntailConfiguration` model (`getDependenciesViaThorntailModel`). This model is available if developers make use of the `thorntail` plugin *or* make use of the new plugin that I am introducing (`thorntail-arquillian`) plugin. By access data via a custom plugin, we have greater control on the way data is constructed instead of grappling with workarounds, which is what we have been doing so far.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/AbstractThorntailPlugin.java
```diff
@@ -0,0 +1,100 @@
...
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.gradle.tooling.provider.model.ToolingModelBuilder;
+import org.gradle.tooling.provider.model.ToolingModelBuilderRegistry;
+
+/**
+ * The base plugin that initializes the tooling model.
+ */
+@SuppressWarnings("UnstableApiUsage")
+public class AbstractThorntailPlugin implements Plugin<Project> {
```
roguexz, line 30: I moved code from the `PackagePlugin` class in to an abstract class that is extended by the `PackagePlugin` as well as the `ThorntailArquillian` plugin. We need a separate vector (in this case a plugin) for targeting test cases. If we overload the existing plugin and insist that developers need to add it to their projects, then we will end up messing up the lifecycle of "library-modules" which don't have to be JEE applications.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/GradleDependencyResolutionHelper.java
```diff
@@ -260,4 +345,45 @@ private static void printDependencyMap(Map<DependencyDescriptor, Set<DependencyD
         }
         return value;
     }
+
+    // Translate the different Gradle dependency scopes in to values that map to ShrinkWrap's scope type.
+    // c.f.: https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_plugin_and_dependency_management
+    private static final Map<String, String> REMAPPED_SCOPES = new HashMap<>();
+
+    static {
```
roguexz, line 353: I had to rework this class to take in to account all possible "known" configurations (a.k.a., dependency scopes) for Gradle 5 based projects. In the above code, there is a *fallback* which translates unknown configurations (a.k.a., user defined configurations) in to the closest "known" configuration. I have tested this behavior as well.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailExtension.java
```diff
@@ -256,11 +261,50 @@ public boolean isHollow() {
...
+            dependencyMap = GradleDependencyResolutionHelper.determineProjectDependencies(project, "runtimeClasspath", false);
         }
         return dependencyMap;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Map<DependencyDescriptor, Set<DependencyDescriptor>> getTestDependencies() {
```
roguexz, line 273: Instead of polluting the existing `getDependencies` method, I have included a separate method for `test` scoped dependencies.

sberyozkin:
Hi @roguexz, since I'm not a Gradle expert I can't really offer some good feedback but the PR itself is obviously a quality one, you know what you are doing, so I'm fine with merging it, given that the examples are OK now

sberyozkin:
@Ladicek, I'll merge it shortly, unless you have some more comments

roguexz:
Thank you @sberyozkin. I have a very strong feeling that I am probably the only one using the Gradle based integration 🤣 ... But, I have faith that once it is all done, I can convince more folks to start using it.

sberyozkin:
Hi, +1

sberyozkin (approved):

Ladicek:
I have no objections nor comments. It slightly worries me that the Gradle plugin evolves so fast that it will soon be better than the Maven plugin :-)

PR #1193, opened 2019-01-02
THORN-2090: Override AnnotationDeploymentScenarioGenerator.

mkouba:
- there was an incompatible change in Arquillian SPI in 1.3.0, see also
ARQ-2173


Ladicek (dismissed):

Ladicek:
Looks like rebasing would be nice, but the change itself LGTM.

Ladicek:
Just thinking, will this affect older Arquillian versions? I guess hopefully maybe probably not, but I don't really know.

mkouba:
@Ladicek which versions do you have in mind? TT2 is currently using 1.1.x and I've tested this manually with 1.3.0 and 1.4.1.

Ladicek:
I mean if it's good with today's 1.1.x.

mkouba:
@Ladicek CI builds passed so I suppose it is ;-)

Ladicek:
Ha, yea, this is about `@DefaultDeployment`, which we use a lot in the tests. Thanks!

Ladicek (approved):

PR #1197, opened 2019-01-05
THORN-2292: Environment Variable usage missing

kenfinnigan:

-----

Motivation
----------
Environment variable usage doc was created and added to codebase, but never included into document structure.

Modifications
-------------
Fix some headings in module doc and add 'Environment Variable' configuration section to include it under.

Result
------
No impact to product. Adds Environment Variable usage to docs.

### DIFF
--- a/docs/assemblies/assembly_configuring-a-thorntail-application.adoc
+++ b/docs/assemblies/assembly_configuring-a-thorntail-application.adoc
@@ -26,7 +26,10 @@ include::../modules/proc_specifying-jdbc-drivers-for-hollow-jars.adoc[leveloffse
 [id='configuring-a-thorntail-application-using-environment-variables_{context}']
 == Environment Variables
 
-Using environment variables for configuring your application, or overriding values provided in YAML files, is advantageous for environments like Docker, Kubernetes, and OpenShift.
+Use environment variables to configure your application or override values stored in YAML files.
+ifndef::product[]
+This approach is useful when deploying to environments like Docker, Kubernetes, or OpenShift.
+endif::[]
 
 include::../modules/con_application-configuration-using-environment-variables.adoc[leveloffset=+2]
 
--- a/docs/assemblies/assembly_configuring-a-thorntail-application.adoc
+++ b/docs/assemblies/assembly_configuring-a-thorntail-application.adoc
@@ -23,6 +23,13 @@ include::../modules/proc_setting-system-properties-using-the-command-line.adoc[l
 
 include::../modules/proc_specifying-jdbc-drivers-for-hollow-jars.adoc[leveloffset=+2]
 
+[id='configuring-a-thorntail-application-using-environment-variables_{context}']
+== Environment Variables
+
+Using environment variables for configuring your application, or overriding values provided in YAML files, is advantageous for environments like Docker, Kubernetes, and OpenShift.
+
+include::../modules/con_application-configuration-using-environment-variables.adoc[leveloffset=+2]
+
 [id='configuring-a-thorntail-application-using-yaml-files_{context}']
 == YAML files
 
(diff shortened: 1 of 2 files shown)


Ladicek:
LGTM, but I'll let @tradej approve :-)

docs/assemblies/assembly_configuring-a-thorntail-application.adoc
```diff
@@ -23,6 +23,13 @@ include::../modules/proc_setting-system-properties-using-the-command-line.adoc[l
 
 include::../modules/proc_specifying-jdbc-drivers-for-hollow-jars.adoc[leveloffset=+2]
 
+[id='configuring-a-thorntail-application-using-environment-variables_{context}']
```
tradej, line 26: Please remove this anchor ID, linking to the middle of a module or assembly is discouraged (except justified cases like API reference) as we don't want the reader to arrive at a place where some prior context may be missing.

docs/assemblies/assembly_configuring-a-thorntail-application.adoc
```diff
@@ -23,6 +23,13 @@ include::../modules/proc_setting-system-properties-using-the-command-line.adoc[l
...
+== Environment Variables
```
tradej, line 27: I'm not sure you need to put this header here, as there is little more than in the `con_application-configuration-using-environment-variables.adoc` file itself.

I suggest you add the below introduction into the introduction in `con_application-configuration-using-environment-variables.adoc`, and include that file with `leveloffset=+1` instead of `+2`, which will make it equal with other sections in this assembly.

_This is a suggestion, and if you believe the current status is superior, feel free to disregard it._

docs/assemblies/assembly_configuring-a-thorntail-application.adoc
```diff
@@ -23,6 +23,13 @@ include::../modules/proc_setting-system-properties-using-the-command-line.adoc[l
...
+
+Using environment variables for configuring your application, or overriding values provided in YAML files, is advantageous for environments like Docker, Kubernetes, and OpenShift.
```
tradej: For simplicity, I suggest you rephrase the sentence as:

```
Use environment variables to configure your application or override values stored in YAML files.
ifndef::product[]
This approach is useful in deployments like Docker, Kubernetes, or OpenShift.
endif::[]
```

I have restricted the information about the deployments to community docs only because in the product, we assume only OpenShift deployment AFAICT. If that's not the case, a similar sentence mentioning only OpenShift should be added for prod builds.

tradej (changes requested):
Thank you for the PR, comments inline.

docs/assemblies/assembly_configuring-a-thorntail-application.adoc
kenfinnigan: But I just followed the pattern as just below this the YAML section creates an anchor ID as well.

docs/assemblies/assembly_configuring-a-thorntail-application.adoc
kenfinnigan: I felt it was better to stick with the pattern to allow for additional content to be added in the future

docs/assemblies/assembly_configuring-a-thorntail-application.adoc
kenfinnigan: I modified it, but with slightly different wording as the above made it seem like Docker, etc were the deployment as opposed to the environment

docs/assemblies/assembly_configuring-a-thorntail-application.adoc
tradej: Those anchor shouldn't be there either, it's an oversight on my part when I was creating those files. Since there are some xrefs pointing to them, I'll address that in a separate issue. Feel free to leave the ID in as I'll deal with it then.

docs/assemblies/assembly_configuring-a-thorntail-application.adoc
tradej: Sure.

tradej (approved):

tradej:
FYI I have filed [THORN-2299](https://issues.jboss.org/browse/THORN-2299) to track the anchor ID issue.

PR #1204, opened 2019-01-07
THORN-2290: Clarify default datasource in docs

kenfinnigan:

-----

Motivation
----------
It's unclear from the documentation that a default datasource is created.

Modifications
-------------
Add content specifying that a default datasource is created, what it's called, and how to override its name.

Result
------
No impact to product, improve clarity of docs.

### DIFF
--- a/docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
+++ b/docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
@@ -28,7 +28,7 @@
 +
 In the `thorntail` -> `datasources` -> `jdbc-drivers` tree, define the driver by
 key, which will be used later for connecting it to the datasource.
-Set the `driver-module-name` to that defined in the `module.xml`.
+Set the `driver-module-name` to that defined in the `module.xml` file.
 +
 [source,yaml]
 ----
--- a/docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
+++ b/docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
@@ -6,7 +6,7 @@
 == Prerequisites
 
 * A Maven-based application.
-* The `module.xml` file created according to relevant {WildFly} instructions.
+* The `module.xml` file created in `src/main/resources/modules` directory according to relevant {WildFly} instructions.
   This file is necessary to expose the driver through JBoss Modules.
 
 [discrete]
@@ -28,6 +28,7 @@
 +
 In the `thorntail` -> `datasources` -> `jdbc-drivers` tree, define the driver by
 key, which will be used later for connecting it to the datasource.
+Set `driver-module-name` to that defined by the `module.xml` from earlier.
 +
 [source,yaml]
 ----
--- a/docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
+++ b/docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
@@ -28,7 +28,7 @@
 +
 In the `thorntail` -> `datasources` -> `jdbc-drivers` tree, define the driver by
 key, which will be used later for connecting it to the datasource.
-Set the `driver-module-name` to that defined by the `module.xml` from above.
+Set the `driver-module-name` to that defined in the `module.xml`.
 +
 [source,yaml]
 ----
--- a/docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
+++ b/docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
@@ -28,7 +28,7 @@
 +
 In the `thorntail` -> `datasources` -> `jdbc-drivers` tree, define the driver by
 key, which will be used later for connecting it to the datasource.
-Set the `driver-module-name` to that defined by the `module.xml` from earlier.
+Set the `driver-module-name` to that defined by the `module.xml` from above.
 +
 [source,yaml]
 ----
--- a/docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
+++ b/docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
@@ -6,7 +6,7 @@
 == Prerequisites
 
 * A Maven-based application.
-* The `module.xml` file created in `src/main/resources/modules` directory according to relevant {WildFly} instructions.
+* The `module.xml` file created in the `src/main/resources/modules` directory according to relevant {WildFly} instructions.
   This file is necessary to expose the driver through JBoss Modules.
 
 [discrete]
@@ -28,7 +28,7 @@
 +
 In the `thorntail` -> `datasources` -> `jdbc-drivers` tree, define the driver by
 key, which will be used later for connecting it to the datasource.
-Set `driver-module-name` to that defined by the `module.xml` from earlier.
+Set the `driver-module-name` to that defined by the `module.xml` from earlier.
 +
 [source,yaml]
 ----
--- a/docs/howto/create-a-datasource/index.adoc
+++ b/docs/howto/create-a-datasource/index.adoc
@@ -19,7 +19,7 @@ a pool of connections to a particular database using the driver.
 
 == Configuring a JDBC driver
 
-There are two options how to configure JDBC drivers:
+There are the following options how to configure JDBC drivers:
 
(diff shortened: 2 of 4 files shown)


docs/howto/create-a-datasource/using-default-datasource.adoc
```diff
@@ -0,0 +1,16 @@
...
+[discrete]
+== Prerequisites
+
+* A Maven-based application.
+* A JDBC Driver configured, either auto detected or manually.
+
+[discrete]
+== Procedure
+
+The presence of a JDBC driver will automatically create a datasource with the name `ExampleDS`.
```
Ladicek: Only applies for certain set of JDBC drivers. We have a list in the docs (https://docs.thorntail.io/2.2.1.Final/#_autodetectable_drivers), maybe we should put a link here?

Ladicek:
LGTM, but I'll let @tradej approve :-)

docs/howto/create-a-datasource/using-default-datasource.adoc
kenfinnigan: From looking at the code it looked like installed JDBC drivers, auto or manual, will trigger the default datasource creation

Ladicek: Aha! Makes sense, thanks.

docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
```diff
@@ -6,7 +6,7 @@
 == Prerequisites
 
 * A Maven-based application.
-* The `module.xml` file created according to relevant {WildFly} instructions.
+* The `module.xml` file created in `src/main/resources/modules` directory according to relevant {WildFly} instructions.
```
tradej: Please add ...in _the_ `src...`

docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
```diff
@@ -28,6 +28,7 @@
 +
 In the `thorntail` -> `datasources` -> `jdbc-drivers` tree, define the driver by
 key, which will be used later for connecting it to the datasource.
+Set `driver-module-name` to that defined by the `module.xml` from earlier.
```
tradej: Please add ...Set _the_ `driver-module-name`...

docs/howto/create-a-datasource/index.adoc
```diff
@@ -25,7 +25,13 @@ include::auto-detecting-jdbc-drivers.adoc[leveloffset=+2]
 
 include::configuring-the-jdbc-driver-manually.adoc[leveloffset=+2]
 
-include::configuring-a-datasource.adoc[leveloffset=+1]
+== Configuring a Datasource
+
+There are two options for how to configure a datasource:
```
tradej: Please replace _two_ with _the following._ We try to avoid referencing numbers of things in text if possible as the numbers may change and thus introduce unnecessary maintenance overhead.

docs/howto/create-a-datasource/using-default-datasource.adoc
```diff
@@ -0,0 +1,16 @@
+
+[id='using-default-datasource_{context}']
+= Using the default datasource
+
+[discrete]
+== Prerequisites
+
+* A Maven-based application.
+* A JDBC Driver configured, either auto detected or manually.
```
tradej: Please rephrase, e. g. _A JDBC Driver configured manually or auto-detected._

docs/howto/create-a-datasource/using-default-datasource.adoc
tradej: In the _Procedure_ sub-section, only a list of steps is allowed. The information you put here would make a good introduction, and instead you can put here an example of how to use the `thorntail.ds.name` property:

1. Customize the name of the datasource In the $CONFIG_FILE by adding the `thorntail.ds.name` property:
    EXAMPLE_CODE
2. Some other steps?

tradej (changes requested):
Thanks for the PR, @kenfinnigan, I've left comments inline.

docs/howto/create-a-datasource/using-default-datasource.adoc
kenfinnigan: By introduction, do you mean the text should be moved before the prerequisite section?

tradej: Exactly. Possibly include there some more information like where using the default datasource is preferable or, on the other hand, where it is discouraged. Thanks.

kenfinnigan: @tradej what's the way to say "put your value here"?

Is it like "{value}"?

tradej: @kenfinnigan I'm not sure I quite understand the question, can you provide a little more context, please? `{value}` is a dereference of the attribute (i. e. AsciiDoc variable) named `value`.

docs/howto/create-a-datasource/using-default-datasource.adoc
```diff
@@ -0,0 +1,24 @@
...
+[discrete]
+== Prerequisites
+
+* A Maven-based application.
+* A JDBC Driver configured manually or auto-detected.
+
+[discrete]
+== Procedure
+
+. Edit the `project-defaults.yml` file to customize the datasource name. The configuration is stored under `thorntail`
```
tradej: Please have only one sentence per line. This helps when maintaining the text and to keep the lines shorter.

docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
```diff
@@ -28,6 +28,7 @@
 +
 In the `thorntail` -> `datasources` -> `jdbc-drivers` tree, define the driver by
 key, which will be used later for connecting it to the datasource.
+Set the `driver-module-name` to that defined by the `module.xml` from earlier.
```
tradej: Please replace _earlier_ with an explicit reference to a file/config/point in time where the variable was defined. We try to avoid ambiguity as much as possible.

tradej (changes requested):
May I ask for two more changes? Otherwise LGTM.

docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
kenfinnigan: How do I refer to the prereq step from this doc?

docs/howto/create-a-datasource/using-default-datasource.adoc
kenfinnigan: Basically how do I represent some yaml config like:

----
thorntail:
  ds:
    name: myDS
----

Where "myDS" is basically whatever name they want. Usually I'd make it something like "{yourName}" but not sure how we're meant to do that

kenfinnigan:
Any other comments @tradej?

docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
```diff
@@ -28,6 +28,7 @@
 +
 In the `thorntail` -> `datasources` -> `jdbc-drivers` tree, define the driver by
 key, which will be used later for connecting it to the datasource.
+Set the `driver-module-name` to that defined by the `module.xml` from above.
```
tradej: After some further thinking, I think the best way is to say _...defined in the `src/main/resources/modules/module.xml` file._ or _...defined in the `module.xml` file._ I find the word _above_ slightly confusing as to where we're actually pointing the reader.

docs/howto/create-a-datasource/configuring-the-jdbc-driver-manually.adoc
```diff
@@ -28,6 +28,7 @@
 +
 In the `thorntail` -> `datasources` -> `jdbc-drivers` tree, define the driver by
 key, which will be used later for connecting it to the datasource.
+Set the `driver-module-name` to that defined in the `module.xml`.
```
tradej: Grammatically, this isn't great (you'd say _defined in `module.xml`_ or, preferably, _defined in the `module.xml` file_), can you please opt for one of those variants? In the case of `driver-module-name`, the situation is a bit different because when you read the instruction out loud, it goes _Set the driver module name,_ which works. Thanks.

tradej (approved):
Thanks, @kenfinnigan, LGTM now.

PR #1216, opened 2019-01-22
THORN-2313: Gradle plugin: provide tasks to run|start|stop the applic…

TomasHofman:
…ation

Motivation
----------
The ability to run Thorntail app directly from gradle would (1.) simplify running
functional tests in gradle projects and (2.) improve user experience.

Modifications
-------------
* Tasks "thorntail-[run|start|stop]" were added to the gradle plugin.
* Configuration properties relevant for running the app were added to
  ThorntailExtension of the gradle plugin.

Result
------
It is now possible to run the Thorntail application via "thorntail-run" gradle
task. The "thorntail-start" gradle task can be used to start the application
before integration tests are executed. User should arrange for the
"thorntail-stop" task to be run after integration tests to terminate the
application process. If the stop task is not invoked, the application process
will be terminated at the end of the build.

- [ ] Have you followed the guidelines in our [Contributing](https://thorntail.io/community/contributing/) document?

-----

### DIFF
--- a/plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
+++ b/plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
@@ -1,0 +1,386 @@
+package org.wildfly.swarm.plugin.gradle;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+import org.gradle.BuildListener;
+import org.gradle.BuildResult;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.GradleException;
+import org.gradle.api.artifacts.repositories.MavenArtifactRepository;
+import org.gradle.api.initialization.Settings;
+import org.gradle.api.invocation.Gradle;
+import org.gradle.api.tasks.TaskAction;
+import org.gradle.api.tasks.bundling.Jar;
+import org.gradle.api.tasks.bundling.War;
+import org.gradle.initialization.BuildCancellationToken;
+import org.slf4j.helpers.MessageFormatter;
+import org.wildfly.swarm.bootstrap.util.BootstrapProperties;
+import org.wildfly.swarm.fractions.FractionDescriptor;
+import org.wildfly.swarm.fractions.FractionList;
+import org.wildfly.swarm.fractions.FractionUsageAnalyzer;
+import org.wildfly.swarm.tools.ArtifactResolvingHelper;
+import org.wildfly.swarm.tools.ArtifactSpec;
+import org.wildfly.swarm.tools.BuildTool;
+import org.wildfly.swarm.tools.DependencyManager;
+import org.wildfly.swarm.tools.exec.SwarmExecutor;
+import org.wildfly.swarm.tools.exec.SwarmProcess;
+
+public class StartTask extends DefaultTask {
+
+    private static final String EXCLUDE_PREFIX = "!";
+    private static final String INDENTATION_DELIMITER = "\n  ";
+    private static final String DEBUG_PORT_PROPERTY = "thorntail.debug.port";
+    private static final String FAILED_TO_START_MESSAGE = "Thorntail process failed to start: ";
+
+    private ThorntailExtension extension;
+    private boolean waitForProcess = false;
+
+    public StartTask() {
+        extension = getProject().getExtensions().getByType(ThorntailExtension.class);
+    }
+
+    void waitForProcess() {
+        waitForProcess = true;
+    }
+
+    @TaskAction
+    public void startApplication() {
+        final SwarmExecutor executor;
+        if (extension.isUseUberJar()) {
+            executor = uberJarExecutor();
+        } else if (!getProject().getTasks().withType(War.class).isEmpty()) {
+            executor = warExecutor();
+        } else if (!getProject().getTasks().withType(Jar.class).isEmpty()) {
+            executor = jarExecutor();
+        } else {
+            throw new GradleException("Unable to determine type of the project, neither war or jar tasks were found.");
+        }
+
+        executor.withJVMArguments(extension.getJvmArguments());
+        executor.withArguments(extension.getArguments());
+
+        final SwarmProcess process;
+        try {
+
+            File processFile;
+            try {
+                processFile = Files.createTempFile("thorntail-process-file", null).toFile();
+                getLogger().info("Thorntail process file: " + processFile.toPath().toString());
+            } catch (IOException e) {
+                throw new GradleException("Error while creating Thorntail process file");
+            }
+
+            process = executor.withDebug(getDebugPort())
+                    .withProcessFile(processFile)
+                    .withProperties(extension.getProperties())
+                    .withProperties(ThorntailUtils.getPropertiesFromFile(extension.getPropertiesFile()))
+                    .withStdoutFile(extension.getStdoutFile() != null ? getProject().file(extension.getStdoutFile()).toPath() : null)
+                    .withStderrFile(extension.getStderrFile() != null ? getProject().file(extension.getStderrFile()).toPath() : null)
+                    .withEnvironment(extension.getEnvironment())
+                    .withEnvironment(ThorntailUtils.getPropertiesFromFile(extension.getEnvironmentFile()))
+                    .withWorkingDirectory(getProject().getProjectDir().toPath())
+                    .withProperty("remote.maven.repo",
+                            getProject().getRepositories().withType(MavenArtifactRepository.class).stream()
+                                    .map(mavenArtifactRepository -> mavenArtifactRepository.getUrl().toASCIIString())
+                                    .collect(Collectors.joining(",")))
+                    .execute();
+            registerCancellationListeners(process);
+
+            process.awaitReadiness(extension.getStartTimeout(), TimeUnit.SECONDS);
+
+            if (!process.isAlive()) {
+                throw new GradleException(
+                        MessageFormatter.format(FAILED_TO_START_MESSAGE + "process not ready in {} seconds",
+                                extension.getStartTimeout()).getMessage());
+            }
+            if (process.getError() != null) {
+                throw new GradleException(FAILED_TO_START_MESSAGE
+                        + (process.getError() != null ? process.getError().getMessage() : "unknown reason"),
+                        process.getError());
+            }
+
+        } catch (IOException e) {
+            throw new GradleException(FAILED_TO_START_MESSAGE + e.getMessage(), e);
+        } catch (InterruptedException e) {
+            throw new GradleException(FAILED_TO_START_MESSAGE + "interrupted", e);
+        }
+
+        if (waitForProcess) {
+            try {
+                process.waitFor();
+            } catch (InterruptedException e) {
+                try {
+                    process.stop(extension.getStopTimeout(), TimeUnit.SECONDS);
+                } catch (InterruptedException ie) {
+                    // Do nothing
+                }
+            } finally {
+                process.destroyForcibly();
+            }
+        } else {
+            // save the process into a property so it can be stopped by another task later
+            getProject().getExtensions().getExtraProperties().set(PackagePlugin.THORNTAIL_PROCESS_PROPERTY, process);
+        }
+    }
+
+    private SwarmExecutor uberJarExecutor() {
+        getLogger().info("Creating uber-jar Thorntail executor.");
+        return new SwarmExecutor().withExecutableJar(getUberJarArchivePath().toPath());
+    }
+
+    private SwarmExecutor warExecutor() {
+        getLogger().info("Creating WAR Thorntail executor.");
+
+        final String finalName = ThorntailUtils.getArchiveTask(getProject()).getArchiveName();
+
+        final SwarmExecutor executor = new SwarmExecutor()
+                .withModules(getModules())
+                .withProperty(BootstrapProperties.APP_NAME, finalName)
+                .withClassPathEntries(getClassPathEntries(Collections.singletonList(getArchivePath().toPath()), false));
+
+        if (extension.getMainClassName() != null) {
+            getLogger().info("With Thorntail app main class " + extension.getMainClassName());
+            executor.withMainClass(extension.getMainClassName());
+        } else {
+            getLogger().info("With default Thorntail app main class.");
+            executor.withDefaultMainClass();
+        }
+
+        executor.withProperty(BootstrapProperties.APP_PATH, getArchivePath().getPath());
+        return executor;
+    }
+
+    private SwarmExecutor jarExecutor() {
+        getLogger().info("Creating WAR Thorntail executor.");
+
+        final String finalName = ThorntailUtils.getArchiveTask(getProject()).getArchiveName();
+
+        // TODO: application path could possibly contain more classesDirs
+        // TODO: getClassesDir() no longer exists in gradle 4+?, gradle-plugins dep should be upgraded, but current
+        // version is not available as a maven artifact
+//        getProject().getConvention().getPlugin(JavaPluginConvention.class).getSourceSets()
+//                .getByName(SourceSet.MAIN_SOURCE_SET_NAME).getOutput().getClassesDir();
+        final List<Path> sourcePaths = Arrays.asList(getProject().getBuildDir().toPath().resolve("classes/java/main"),
+                getProject().getBuildDir().toPath().resolve("resources/main"));
+        getLogger().info("Source paths: {}", sourcePaths);
+
+        final SwarmExecutor executor = new SwarmExecutor()
+                .withModules(getModules())
+                .withProperty(BootstrapProperties.APP_NAME, finalName)
+                .withClassPathEntries(sourcePaths)
+                .withClassPathEntries(getClassPathEntries(sourcePaths, true));
+
+        if (extension.getMainClassName() != null) {
+            getLogger().info("With Thorntail app main class " + extension.getMainClassName());
+            executor.withMainClass(extension.getMainClassName());
+        } else {
+            getLogger().info("With default Thorntail app main class.");
+            executor.withDefaultMainClass();
+        }
+
+        return executor;
+    }
+
+    private File getArchivePath() {
+        final Jar task = ThorntailUtils.getArchiveTask(getProject());
+        final File archivePath = task.getArchivePath();
+        getLogger().info("Archive path: {}", archivePath.getAbsolutePath());
+        return archivePath;
+    }
+
+    private File getUberJarArchivePath() {
+        final File outputFile = getProject().getTasks().withType(PackageTask.class)
+                .getByName(PackagePlugin.THORNTAIL_PACKAGE_TASK_NAME).getOutputFile();
+        getLogger().info("Thorntail Uber-JAR archive path: {}", outputFile.getAbsolutePath());
+        return outputFile;
+    }
+
+    private List<Path> getModules() {
+        List<Path> modules = extension.getModules().stream().map(File::toPath).collect(Collectors.toList());
+        getLogger().info("With Thorntail modules:\n  {}",
+                modules.stream().map(Path::toString).collect(Collectors.joining(INDENTATION_DELIMITER)));
+        return modules;
+    }
+
+    private List<Path> getClassPathEntries(List<Path> sourcePaths, boolean scanDependencies) {
+        Collection<ArtifactSpec> allDependencies = GradleToolingHelper
+                .toDeclaredDependencies(extension.getDependencies()).getRuntimeExplicitAndTransientDependencies();
+        getLogger().info("Original dependencies including transitive:\n  {}",
+                allDependencies.stream().map(ArtifactSpec::toString).collect(Collectors.joining(INDENTATION_DELIMITER)));
+
+        final List<Path> classpathEntries = allDependencies.stream()
+                .map(artifactSpec -> artifactSpec.file.toPath())
+                .collect(Collectors.toList());
+
+        final boolean hasThorntailDeps = allDependencies.stream()
+                .anyMatch(a -> FractionDescriptor.THORNTAIL_GROUP_ID.equals(a.groupId())
+                        && DependencyManager.WILDFLY_SWARM_BOOTSTRAP_ARTIFACT_ID.equals(a.artifactId()));
+        getLogger().info("Thorntail dependencies detected: {}", hasThorntailDeps);
+
+        // automatic fraction detection
+        if (extension.getFractionDetectionMode() != BuildTool.FractionDetectionMode.never
+                && (extension.getFractionDetectionMode() == BuildTool.FractionDetectionMode.force
+                || !hasThorntailDeps)) {
+            getLogger().info("Detecting fractions.");
+
+            final FractionUsageAnalyzer analyzer = new FractionUsageAnalyzer(FractionList.get());
+            // always scan application source files
+            sourcePaths.forEach(analyzer::source);
+            // scan dependencies if indicated
+            if (scanDependencies) {
+                classpathEntries.forEach(analyzer::source);
+            }
+
+            final Set<FractionDescriptor> detectedFractions;
+            try {
+                detectedFractions = analyzer.detectNeededFractions()
+                        .stream()
+                        // filter out fractions already on classpath
+                        .filter(fd -> !allDependencies.contains(ArtifactSpec.fromFractionDescriptor(fd)))
+                        .collect(Collectors.toSet());
+                getLogger().info("Detected fractions:\n  {}",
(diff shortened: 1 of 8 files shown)


Ladicek:
@roguexz Could you please review? :-) Thanks!

TomasHofman:
@roguexz I wanted to add some integration tests here, but the fact that this gradle plugin is implemented as a maven project makes it bit difficult. Problem is that things like gradle-test-kit are not available in the form of maven artifacts (Or at least I didn't find them. For that matter I can't even find basic gradle libs like gradle-core-api as maven artifacts, which would make it impossible to upgrade this plugin to gradle 5).

How do you go about testing this? Just manually on examples?

roguexz:
@TomasHofman 

First up, thank you for this PR. Based on the questions that you are asking, I am assuming that you have managed to figure out the current challenges with developing the Gradle plugin 😃 

The current approach for developing the Gradle plugin is less than ideal. I usually test the code against my personal projects and then port the capabilities to the examples project. This is not an efficient model and will need to change in the future, i.e., building the plugin as an independent Gradle project. In order to get there, we need to ensure that there are enough API contracts in place so that we can disconnect the Gradle extensions from the main project.

I am traveling for work at the moment, so give me a couple of days and I will review the code. I made an initial pass and the proposed changes look good. I shall give you my feedback in a couple of days.

TomasHofman:
Linking simple example of a project running integration tests:
https://github.com/TomasHofman/thorntail-examples/blob/integration-test-example/gradle-examples/jaxrs-integration-tests/build.gradle

and project leveraging fraction auto-detection:
https://github.com/TomasHofman/thorntail-examples/blob/integration-test-example/gradle-examples/autodetect-jaxrs/build.gradle

(Need to create readme files yet.)

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+
+            process = executor.withDebug(extension.getDebugPort())
+                    .withProcessFile(tmp)
+                    .withProperties(extension.getProperties())
+                    .withProperties(ThorntailUtils.getPropertiesFromFile(extension.getPropertiesFile()))
+                    .withStdoutFile(extension.getStdoutFile() != null ? getProject().file(extension.getStdoutFile()).toPath() : null)
+                    .withStderrFile(extension.getStderrFile() != null ? getProject().file(extension.getStderrFile()).toPath() : null)
+                    .withEnvironment(extension.getEnvironment())
+                    .withEnvironment(ThorntailUtils.getPropertiesFromFile(extension.getEnvironmentFile()))
+                    .withWorkingDirectory(getProject().getProjectDir().toPath())
```
roguexz, line 92: I am not entirely sure if the working directory ought to be the build-dir or project-dir. I've observed developers hard code file paths in test cases just because the working directory was set to the project-dir. I am good either ways, though I would prefer that we set the WD to the build dir. Your thoughts?

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+                    .withProperty("remote.maven.repo",
+                            getProject().getRepositories().withType(MavenArtifactRepository.class).stream()
```
roguexz, line 94: Nice!!

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+
+            File tmp;
+            try {
+                tmp = Files.createTempFile("thorntail-process-file", null).toFile();
+                getLogger().info("Process file: " + tmp.toPath().toString());
+            } catch (IOException e) {
+                throw new GradleException("Error while creating process file");
+            }
+
+            process = executor.withDebug(extension.getDebugPort())
```
roguexz: As an enhancement (for a later PR), we could check if the port is already used or not. This would help developers who run the `thorntail-start` task and forget about it, snooze their laptop and try running it again. I learnt it the hard way for something I did earlier :)

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+            process.awaitReadiness(extension.getStartTimeout(), TimeUnit.SECONDS);
+
+            if (!process.isAlive()) {
+                throw new GradleException("Process failed to start");
+            }
+            if (process.getError() != null) {
+                throw new GradleException("Error starting process", process.getError());
+            }
+
+        } catch (IOException e) {
```
roguexz, line 113: It would be nice if the error messages were standardized across the different exceptions. Like, `Error starting Thorntail process.` (followed by any additional information that we wish to post).

Any concerns around standardizing the messages?

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+        } catch (InterruptedException e) {
+            throw new GradleException("Error waiting for deployment", e);
+        }
+
+        if (waitForProcess) {
+            try {
+                process.waitFor();
+            } catch (InterruptedException e) {
+                try {
+                    process.stop(10, TimeUnit.SECONDS);
```
roguexz: You may want to increase this timeout to say 2 minutes, or make use of the same value as available in the extension. In the past, there was a verifier in the build process that would attempt to see if the tmp files were deleted or not. The build used to fail quite consistently on my laptop till I realized that the deletion process was running slow (on a corporate laptop). So, please change this to a higher number (in minutes), or make use of the same timeout that we have defined in the extension.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+                } catch (InterruptedException ie) {
+                    // Do nothing
+                }
+            } finally {
+                process.destroyForcibly();
+            }
+        } else {
+            // save the process into a property so it can be stopped by another task later
+            getProject().getExtensions().getExtraProperties().set(PackagePlugin.THORNTAIL_PROCESS_PROPERTY, process);
```
roguexz, line 133: I am a little lost here. Can you confirm the behavior of the `thorntail-start` task? Will it,

1. Start the process and let it run in the background? Meaning, will the Gradle process exit after the Thorntail process has been started?

or,

2. Wait for an explicit control termination from the user (Ctrl+C)?

I haven't worked with the `SwarmProcess` handler so I am not sure of this behavior.

If the behavior follows `option 1`, then you should consider writing the process ID to a file in the `buildDir`. This would be relevant if the Gradle daemon is unavailable or terminates unexpectedly.

If the behavior follows `option 2`, then the `thorntail-stop` task is just acting as a pretty command for terminating the process at which point, we should ask if we want to have `start` & `run` as separate tasks.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+        }
+    }
+
+    private SwarmExecutor uberJarExecutor() {
+        getLogger().info("Creating uber-jar Thorntail executor.");
```
roguexz, line 138: With so many `info` logs, maybe we should consider prefixing the logs with something like `[THORNTAIL] `. That would make the logs easily searchable. Like in a CI system.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+import org.wildfly.swarm.tools.ArtifactResolvingHelper;
+import org.wildfly.swarm.tools.ArtifactSpec;
+import org.wildfly.swarm.tools.BuildTool;
+import org.wildfly.swarm.tools.DependencyManager;
+import org.wildfly.swarm.tools.exec.SwarmExecutor;
+import org.wildfly.swarm.tools.exec.SwarmProcess;
+
+public class StartTask extends DefaultTask {
+
+    private static final String EXCLUDE_PREFIX = "!";
```
roguexz, line 41: Is this capability (of excluding fractions) available as part of the standard Maven plugin as well, or is this something that you are adding as part of this plugin? I like the idea, but wanted to make sure if it is for parity or our own testing.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+        Collection<ArtifactSpec> allDependencies = GradleToolingHelper
+                .toDeclaredDependencies(extension.getDependencies()).getRuntimeExplicitAndTransientDependencies();
+        getLogger().info("Original dependencies including transitive:\n  {}",
+                allDependencies.stream().map(ArtifactSpec::toString).collect(Collectors.joining(INDENTATION_DELIMITER)));
+
+        final List<Path> classpathEntries = allDependencies.stream()
+                .map(artifactSpec -> artifactSpec.file.toPath())
+                .collect(Collectors.toList());
+
+        final boolean hasThorntailDeps = allDependencies.stream()
```
roguexz, line 226: Do you want to rename this variable from `hasThorntailDeps` to `hasBootstrapDeps` or something similar? That way, when we refactor the plugin code to include only the essential pieces, it would be easier to read this code. What do you think?

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+                List<Path> resolvedPaths = artifactResolvingHelper().resolveAll(fractionsSpecs).stream()
+                        .map(s -> s.file.toPath())
+                        .collect(Collectors.toList());
+                classpathEntries.addAll(resolvedPaths);
+            } catch (Exception e) {
+                throw new GradleException("Failed to resolve fraction dependencies.", e);
+            }
+        }
+
+        getLogger().info("Computed classpath entries:\n  {}",
```
roguexz, line 299: You might also want to consider reducing the number of logs going to `INFO`. I can imagine the logs getting super huge if we keep printing the dependencies / classpath entries for each step. You should consider printing the final computation alone at `INFO` and move the rest to `DEBUG`.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    private ArtifactResolvingHelper artifactResolvingHelper() {
+        return new GradleArtifactResolvingHelper(getProject());
+    }
+
+    private void stopProcess(SwarmProcess process) {
+        if (process != null && process.isAlive()) {
+            try {
+                process.stop(10, TimeUnit.SECONDS);
```
roguexz: Same comment for timeouts as in the above case.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+            } catch (InterruptedException e) {
+                // ignore
+            } finally {
+                process.destroyForcibly();
```
roguexz, line 316: I hope there is no side-effect of calling this method on a process that has already exited gracefully. The [Javadocs](https://docs.oracle.com/javase/8/docs/api/java/lang/Process.html#destroyForcibly--) don't provide much insights.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StopTask.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.gradle.api.DefaultTask;
+import org.gradle.api.tasks.TaskAction;
+import org.wildfly.swarm.tools.exec.SwarmProcess;
+
+public class StopTask extends DefaultTask {
+
+    @TaskAction
+    public void stopThorntailProcess() {
+        SwarmProcess process = (SwarmProcess) getProject().findProperty(PackagePlugin.THORNTAIL_PROCESS_PROPERTY);
```
roguexz, line 15: As discussed in the above comment, we might need to look at the scenario where the process was kicked off and is running in the background. The Gradle daemon might have crashed and now we wish to stop the process. In such a scenario, if the process ID is available in the `buildDir`, then we can attempt to stop the process (if it is still running). If there is no known processID then we just ignore.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailExtension.java
```diff
@@ -60,6 +62,26 @@
 
     private boolean hollow = false;
 
+    private Integer debugPort = null;
+
+    private File stdoutFile;
```
roguexz, line 67: `stdout` or `stdOut`? Either way works. Please use the one that is more prevalent.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    void waitForProcess() {
+        waitForProcess = true;
+    }
+
+    @TaskAction
+    public void startApplication() {
+        final SwarmExecutor executor;
+        if (extension.isUseUberJar()) {
```
roguexz, line 60: Is there a reason why you wouldn't want to use the `UberJar` only? Meaning, `thorntail-start` just blindly depends on `thorntail-package`. That way, we don't need to incur the maintenance of the build tooling. Did you run in to any issues specifically?

roguexz (commented):
@TomasHofman Sorry for the delay. I have given my first round of feedback. I had a few questions. Can you please let me know your views about them?

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: Yeah, basically my intention was to keep it safe by following the maven plugin behavior as closely as possible, and maven plugin appears to be using project dir:
https://github.com/thorntail/thorntail/blob/master/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/StartMojo.java#L133

The build dir might be better in case that the process decides to create some files, so that it doesn't pollute the project. One think I could test whether this affects location of the log files in case when std or err output file is set to relative path. Will do that tomorrow.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: That's a good idea, I didn't thought of that.

What I thought about and then forgot was to enable for the debug port to be overridden by a property from a command line, something like adding "-Pthorntail.debugPort=8888" to the gradle command so that we don't have to modify the build file to enable debugging. Maybe would leave that also for a latter PR, not to complicate this one further.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: That could be good for clarity, so that user understands what's going on. Will do that.

TomasHofman:
@roguexz np, we have face-to-face meeting this week, so I'm little out of time too. I will try to reply to the rest and hopefully prepare new PR in the following days. Thanks for great input!

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
Ladicek: Only speaking about the proposed system property name: there's a property used for setting a debug port already: `thorntail.debug.port`. I think it would be good to stick to existing names.

roguexz: Agreed.

roguexz:
@TomasHofman Happy to help. The timelines work for me as well. I have been looking at ways of spawning the process using either [Java Service Wrapper](https://wrapper.tanukisoftware.com/doc/english/product-overview.html) | [Apache Commons Daemon](http://commons.apache.org/proper/commons-daemon/). The licensing of JSW is a little complicated and I wouldn't want to introduce that in to the mix. With respect to ACD, we may need to do some tests before proposing that change. These are potential enhancements to consider for a future PR and have no bearing on your current PR. We can always work with the in-process model for now with the known limitations.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailExtension.java
TomasHofman: Maven plugin has
```
    @Parameter(alias = "stdoutFile", property = "thorntail.stdout")
    public File stdoutFile;
```
so I would keep this notation.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: So setting stdoutFile in thorntail extension actually always sets an absolute file, so working directory has no effect here. I would keep WD = project dir to stay consistent with the maven plugin.

```
thorntail {
  stdoutFile = file('stdout.log') // or new File('stdout.log')
  println stdoutFile.absolutePath // prints <projectDir>/stdout.log
}
```

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: Good point, it will look for a project property "thorntail.debug.port".

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: This is something I wanted to look into later - I was aware that `process.stop(...)` does not seem to initiate container shutdown at all, the application process keeps running happily until either:
* destroyForcibly() is called,
* or a gradle daemon terminates.
I presumed it was a bug in `process.stop()` and that it behaves the same way in maven plugin, but now that I actually tested it I can see that when running via the maven plugin the app container shutdowns correctly.

What I think makes the difference is that the maven plugin seems to propagate CTRL+C to the process, which triggers the shutdown even before `process.stop()` is called. In gradle the CTRL+C just terminates the gradle client but is not propagated into the process.

I will need to find a way to initiate the shutdown from the outside.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: I think it just exits when the process was already terminated.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StopTask.java
TomasHofman: One thing that discouraged me from this was that there is no cross-platform way of doing it. There's no API to even retrieve process ID until JDK 9, and then there would be the killing part...

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: I think it could shave couple of seconds off the build time, if you don't need to build the UberJar and use the compiled classes instead.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: Keeping WS = project dir to stay consistent with maven plugin.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: * Detecting "thorntail.debug.port" project property is done.
* Leaving check that port is free for later.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: I see, that was mess. Fixed.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: All should be fixed:
* Stop timeout is configurable, default is 120 seconds.
* Fixed the problem with the process stopping too slowly: https://github.com/TomasHofman/thorntail/commit/1c25dd6828540e7423c684aaf290667f1b5dcae6#diff-d252ac069893b056c7cc65cf320827b8L108

  The problem was that `SwarmProcess#stop()` first deleted the process file and than waited up to the `stopTimeout` for the process to terminate by itself, before it called `Process#destroy()`. You mentioned this was how it worked in the past - shutdown triggered by process file deletion. I deleted that waiting, since I believe that `Process#destroy()` should now be called straight away. `destroy()` now triggers the service shutdown.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StopTask.java
TomasHofman: IMO we should postpone this until newer JDK. With JDK 8 it would require us to use reflection to access private field on JDK type (`Process#pid`), which would trigger illegal access warnings on newer JDKs. It doesn't feel right to add these kind of reflections now, when we should rather be actively removing them. WDYT?

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StartTask.java
TomasHofman: This was also consistency-with-maven-plugin thing. I'm not absolutely sure about the practical aspects, only that I would personally appreciate an option of not having to build an uber-jar.

TomasHofman:
OK, I updated the PR. It currently contains several commits, so it's easier to see what's been changed in the original commit. I would squash the commits later when we are OK with the code.

sberyozkin:
@roguexz, so all is good now ? if yes then I can merge... thanks

roguexz:
@TomasHofman @sberyozkin 
I just got back from my trip. I shall review the code changes today. I don't think there are any further comments, but let me review the changes none the less.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/StopTask.java
roguexz: Agreed

roguexz (dismissed):
Changes look good. I shall take it for a spin once it is merged and propose additional enhancements as I encounter them. Thank you @TomasHofman for your contributions. Really appreciate it.

TomasHofman:
Happy to help @roguexz, thanks again for your input.

@sberyozkin I squashed into a single commit.

sberyozkin:
Thanks @TomasHofman @roguexz

sberyozkin (approved):

sberyozkin:
@TomasHofman, if you can find some time, have a look please if the gradle examples in thorntail-examples can be enhanced somehow to show the latest improvements, thanks

TomasHofman:
@sberyozkin yes, that was the next thing I wanted to look at.

PR #1217, opened 2019-01-22
Initial MP JWT and Keycloak integration

sberyozkin:
This is a work in progress, please do not merge


testsuite/testsuite-keycloak-mpjwt/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>io.thorntail</groupId>
+      <artifactId>microprofile-jwt</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.keycloak</groupId>
+      <artifactId>keycloak-core</artifactId>
+      <version>4.8.1.Final</version>
```
Ladicek: There should be a constant for this somewhere. We'll be updating to the latest 4.8.x soon (currently, x == 3, hopefully that won't change).

Ladicek:
Is this still WIP? I see the builds failed, so I assume yes.

sberyozkin:
Yes, still WIP, hope to get more time on working on it now that the istio community boosters testing is done

testsuite/testsuite-keycloak-mpjwt/pom.xml
sberyozkin: Thanks. testsuites don't inherit from the build-parent, but I will push it to the testsuite-parent. However I'm still not sure where this KC factory should actually reside in the main source code. So it is also likely these direct keycloak dependencies will go away from this test :-)

sberyozkin:
retest this please

sberyozkin:
@Ladicek can you please have another look, it is all green now. FYI, I will follow up with at least 1 PR, but for now it will be simpler to keep the KC code inside the test

pom.xml
```diff
@@ -1231,6 +1231,7 @@
         <module>testsuite/testsuite-jolokia</module>
         <module>testsuite/testsuite-jolokia-keycloak</module>
         <module>testsuite/testsuite-keycloak</module>
+        <module>testsuite/testsuite-keycloak-mpjwt</module>
```
Ladicek, line 1234: I wonder whether `testsuite-keycloak` and `testsuite-keycloak-mpjwt` should be in the `core-testsuite` profile instead of `unsupported-testsuite`?

Ladicek: I'll answer myself: they shouldn't, because they depend on `keycloak-server` :-)

testsuite/testsuite-keycloak-mpjwt/src/test/java/org/wildfly/swarm/keycloak/KeycloakMicroprofileJwtTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.wildfly.swarm.jaxrs.JAXRSArchive;
+import org.wildfly.swarm.microprofile.jwtauth.keycloak.SecuredApplication;
+import org.wildfly.swarm.microprofile.jwtauth.keycloak.SecuredResource;
+import org.wildfly.swarm.microprofile.jwtauth.keycloak.deployment.principal.KeycloakJWTCallerPrincipal;
+import org.wildfly.swarm.microprofile.jwtauth.keycloak.deployment.principal.KeycloakJWTCallerPrincipalFactory;
+
+import io.smallrye.jwt.auth.principal.JWTCallerPrincipalFactory;
+
+/**
+ * @author Bob McWhirter
```
Ladicek: :-)

testsuite/testsuite-keycloak-mpjwt/src/test/java/org/wildfly/swarm/keycloak/KeycloakMicroprofileJwtTest.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+@RunWith(Arquillian.class)
+public class KeycloakMicroprofileJwtTest {
+
+    @Deployment
+    public static Archive<?> createDeployment() throws Exception {
+        JAXRSArchive deployment = ShrinkWrap.create(JAXRSArchive.class, "test.war");
+        deployment.addClass(SecuredApplication.class);
+        deployment.addClass(SecuredResource.class);
+        deployment.addResource(JWTCallerPrincipalFactory.class);
```
Ladicek: Why are some classes added via `addClass` and other via `addResource`?

Ladicek:
I took another look, added a few comments, but mainly want to ask: what does this actually do? If I understand correctly, it merely adds a test which demonstrates how to integrate Keycloak with MP JWT. I guess you want to provide this integration out of the box in some fraction in that "at least 1 PR"?

sberyozkin:
@Ladicek thanks, yes, it is about validating the tokens using the KC adapter (as opposed to a default MP-JWT code), as part of the MP-JWT flow, and making these tokens accessible via MP JWT API. Indeed, I'd like to make it available out of the box, my current thinking is to push it to the keycloak fraction itself - so that when both mpjwt and kc fractions are loaded and MP-JWT login config active it would work.
The main reason I did not do it with this PR is that MP-JWT still requires to add some configuration which is really not needed by the KC factory, for ex the test adds a PEM key just to get an MP-JWT check passed but which is not needed otherwise for the test to pass. Making sure this check is not needed requires some major changes, so I thought I'd not mix it all into a single PR 
I've had various thoughts on where else this can be put, not sure the introduction of the dedicated fraction is needed yet - until a login mechanism optimized for KC (which can handle OIDC sessions, etc. as opposed to the current MP-JWT one) is available....

sberyozkin:
This test also relies on some simplifications made in smallrye-jwt 1.1.1, so it can serve on its own as an example of how to customize, in case someone would like to override a default factory for ex

Ladicek (approved):

PR #1219, opened 2019-01-31
THORN-2324: add toggle for filtering WEB-INF/lib

emattheis:
Motivation
----------
Provides the ability to deactivate the default filtering of JARs from
WEB-INF/lib when repacking WARs. This avoids situations where the
developer knows that no filtering is required but the plugin removes
required dependencies in error.

Modifications
-------------
A boolean property is added to the build tool and package mojo
controlling whether or not to filter WEB-INF/lib. The property defaults
to true in order to preserve existing behavior by default.

Result
------
Developers wishing to do so can set filterWebinfLib to false in their
poms to disabled fitlering in order to mitigated any issues caused by
the plugin. It will be up to the developer to ensure that any JARs in
WEB-INF/lib belong there.

### DIFF
--- a/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
+++ b/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
@@ -58,8 +58,8 @@ public class PackageMojo extends AbstractSwarmMojo {
     @Parameter(alias = "bundleDependencies", defaultValue = "true", property = "thorntail.bundleDependencies")
     protected boolean bundleDependencies;
 
-    @Parameter(alias = "filterWebinfLib", defaultValue = "true", property = "thorntail.filterWebinfLib")
-    protected boolean filterWebinfLib;
+    @Parameter(alias = "filterWebInfLib", defaultValue = "true", property = "thorntail.filterWebInfLib")
+    protected boolean filterWebInfLib;
 
     /**
      * Make a fully executable jar for *nix machines by prepending a launch script to the jar.
@@ -141,7 +141,7 @@ public class PackageMojo extends AbstractSwarmMojo {
                 .properties(this.properties)
                 .mainClass(this.mainClass)
                 .bundleDependencies(this.bundleDependencies)
-                .filterWebinfLib(this.filterWebinfLib)
+                .filterWebInfLib(this.filterWebInfLib)
                 .executable(executable)
                 .executableScript(executableScript)
                 .fractionDetectionMode(fractionDetectMode)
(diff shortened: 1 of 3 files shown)


emattheis:
`mvn clean install` succeeded locally on macOS 10.13.6 using Oracle JDK 1.8.0_172 and Maven 3.6.0 - not sure what went wrong on Semaphore...

emattheis:
Example project where this PR provides a workaround: https://github.com/emattheis/thorntail-issue/tree/filterWebinfLib

plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
```diff
@@ -58,6 +58,9 @@
     @Parameter(alias = "bundleDependencies", defaultValue = "true", property = "thorntail.bundleDependencies")
     protected boolean bundleDependencies;
 
+    @Parameter(alias = "filterWebinfLib", defaultValue = "true", property = "thorntail.filterWebinfLib")
+    protected boolean filterWebinfLib;
```
Ladicek: I'd call it `filterWebInfLib`, but perhaps that's just me :-)

Ladicek:
ok to test

Ladicek:
Please ignore SemaphoreCI, that's just for the `v4` branch.

Ladicek:
Maybe one more thing -- there's a documentation for all the Maven plugin options, this one should probably be added there. It's in `docs/modules/ref_thorntail-maven-plugin-configuration-options.adoc`. Thanks!

plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
emattheis: Totally agree. I THOUGHT I was following the convention set by `WebInfLibFilteringArchive`, but now that I look again it's definitely camel-case. My eyes must have been getting tired. I'll refactor it to be consistent.

emattheis:
I was thinking about the documentation this morning, also. I'll add the new parameter there as well.

plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
Ladicek: Thank you!

emattheis: fixed in commit https://github.com/thorntail/thorntail/pull/1219/commits/875c320d6f1ee4ab6d5484ba4374b350f1637309

emattheis:
@Ladicek I've updated the documentation [here](https://github.com/thorntail/thorntail/blob/f123bb6ae30761241cdddcb1e590270ce4b8d6b3/docs/modules/ref_thorntail-maven-plugin-configuration-options.adoc)

Ladicek (approved):

Ladicek:
Very nice, thanks!

PR #1235, opened 2019-02-25
[THORN-1852] Remove the unused test key config

sberyozkin:
Motivation
----------
Remove the unused test key configuration; keycloak.json provides all the configuration

Modifications
-------------
MP JWT fraction tries to create JWTAuthContextInfo if no JWTAuthContextInfoProvider can provide JWTAuthContextInfo and throws the exception if the verification key (and issuer) is not preconfigured. Apart from the fact that it is inconsistent with the way JWTAuthContextInfoProviders themselves treat it by not throwing the exception and only logging the relevant message (both in Smallrye JWT and the MP JWT fraction extension), it also makes an unnecessary demand on the users of the custom factories to register such a key even when it is not needed

The MP JWT fraction has been modified to log the message if no key is available and a test has been added confirming that it does not introduces a security issue - the default factory enforces that the key is available anyway.

Result
------
Simpler to work with the custom factories like the one depending on KC and keycloak.json

-----


sberyozkin:
@Ladicek This is a simple follow up, to do with removing the unused test key, as implied earlier.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/JWTAuthMechanismFactory.java
```diff
@@ -120,22 +120,23 @@ public AuthenticationMechanism create(String mechanismName, FormParserFactory fo
...
+                        if (jwksRefreshIntervalURL != null) {
+                            jwksRefreshInterval = readURLContent(jwksRefreshIntervalURL);
+                        }
                     }
+                    if (jwksRefreshInterval == null) {
+                        throw new IllegalStateException("JWKS Refresh Interval should be set when JWKS URI is used.");
+                    }
+                    contextInfo.setJwksRefreshInterval(Integer.valueOf(jwksRefreshInterval.trim()));
+                } else {
+                    log.debug("Neither a static key nor a JWKS URI was set.");
```
Ladicek, line 138: So this situation is OK? I mean -- the situation is probably OK when you integrate with Keycloak, but what if you don't and still omit both a static key and JWKS URI?

sberyozkin: This is why I added [this test](https://github.com/thorntail/thorntail/pull/1235/files#diff-baf5ffd0545003af243b14fd97938f31) to confirm that removing this check is safe even with the default code. Note it is part of the default microprofile jwt test suite. Also as I mentioned, this check is not actually done in the [portable configuration case](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java), as well as with [this provider](https://github.com/thorntail/thorntail/blob/master/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/config/JWTAuthContextInfoProvider.java) due to [this code](https://github.com/thorntail/thorntail/blob/master/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/JWTAuthMechanismFactory.java#L70). 
So it is safe to drop this check at this level, it is indirectly enforced anyway [in here](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTCallerPrincipalFactory.java):
```
2019-02-25 15:38:55,931 INFO  [io.undertow.request.security] (default task-1) Failed to authenticate JWT bearer token
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.989 s - in org.wildfly.swarm.microprofile.jwtauth.NoVerificationKeyTest
```

Ladicek: I'm not thinking correctness here, but user-friendliness. But I'm no expert on this code, so I'm gonna trust your judgement :-)

Ladicek (approved):

PR #1237, opened 2019-02-28
THORN-2322: Thorntail Maven Plugin Strips Required Dependencies from …

TomasHofman:
…WEB-INF/lib

Motivation
----------
DeclaredDependencies#getDirectDeps() doesn't maintain the order of direct
dependencies as they were specified in the POM. Those direct dependencies are
then going to be passed for resolution in a random order, which may lead to
different transitive dependencies versions being resolved (compared to what
Maven originally resolved). Later, those wrong dependencies versions break
dependency analysis during repackaging.

Modifications
-------------
DeclaredDependencies#getDirectDeps() returns dependencies ordered so that
compile and provided deps come before the rest, however inside those two groups
the original insertion order is maintained.

Result
------
Valid dependencies are no longer removed.

- [ ] Have you followed the guidelines in our [Contributing](https://thorntail.io/community/contributing/) document?

-----

### DIFF
--- a/tools/src/main/java/org/wildfly/swarm/tools/DeclaredDependencies.java
+++ b/tools/src/main/java/org/wildfly/swarm/tools/DeclaredDependencies.java
@@ -23,9 +23,12 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.wildfly.swarm.bootstrap.env.DependencyTree;
 import org.wildfly.swarm.bootstrap.util.MavenArtifactDescriptor;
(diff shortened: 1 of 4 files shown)


kenfinnigan (dismissed):
If this solves bad removal of transitives I will be extremely happy!

Thanks for taking a look at this

tools/src/main/java/org/wildfly/swarm/tools/DeclaredDependencies.java
```diff
@@ -246,13 +247,23 @@ public void writeTo(File file) {
...
+        // different resolved transitive dependencies.
+
+        Collection<ArtifactSpec> originalDeps = super.getDirectDeps();
+        List<ArtifactSpec> reordered = new ArrayList<>(originalDeps.size());
+        originalDeps.stream().filter(s -> PRIORITIZED_SCOPES.contains(s.scope)).forEach(reordered::add);
+        originalDeps.stream().filter(s -> !PRIORITIZED_SCOPES.contains(s.scope)).forEach(reordered::add);
+        return reordered;
+    }
+
+    private static final List<String> PRIORITIZED_SCOPES = new ArrayList<>();
```
Ladicek: Should be a `Set`?

Ladicek: (I know it's tiny, but conceptually, it should be a `Set` :-) )

Ladicek:
Impressive analysis!

tools/src/main/java/org/wildfly/swarm/tools/DeclaredDependencies.java
Ladicek: Or really just an inline `Predicate`.

TomasHofman: Yeah, good point, I will make it a predicate :).

TomasHofman: I made it a Set.

Ladicek (approved):

PR #1239, opened 2019-03-05
[THORN-1852] Moving KC factory from the test into its own fraction

sberyozkin:
Motivation
----------
At the moment the custom MP JWT Keycloak factory is located inside the testsuite. It needs to become available as part of the fraction for users wishing to use the KC validation be able to add this fraction to their deployments

Modifications
-------------
Keycloak factory has been moved into a new keycloak-microprofile-jwt fraction. It requires a dependency on the microprofile-jwt fraction so the earlier idea to have it moved to the 'keycloak' fraction has been rejected. Having it located in its own dedicated fraction is better longer term - some of this code may get moved to smallrye-jwt-keycloak, and the fraction can also override the JAAS authentication mechanism which is provided by microprofile-jwt to support KC flows better  

Result
------
The users using Keycloak to validate the JWT tokens and now wishing to access these tokens via MP-JWT API will only have to add a new 'keycloak-microprofile-jwt' dependency and reuse the existing keycloak.json configurations


fractions/keycloak-microprofile-jwt/module.conf
```diff
@@ -0,0 +1,3 @@
+org.wildfly.swarm.keycloak.mpjwt export=true
```
Ladicek: This is the module of this fraction, no? Isn't that exported automatically?

fractions/keycloak-microprofile-jwt/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+        <filtering>true</filtering>
+      </resource>
+    </resources>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.thorntail</groupId>
+      <artifactId>microprofile-jwt</artifactId>
+    </dependency>
```
Ladicek, line 43: Shouldn't it also depend on `io.thorntail:keycloak`?

sberyozkin: No, keycloak fraction is using a different mechanism, it needs a KEYCLOAK login type, and uses the undertow adapter code to validate. The new fraction works with the Mp JWT mandated MP-JWT login type and the KC factory shipped inside it uses a simpler mechanism for validating because it is constrained by the microprofile-jwt fraction auth mechanism implementation, i.e it only returns 401 if the token validation fails. So in the future I'd also consider updating the new fraction to make this authentication mechanism to be based on the same underlying undertow KC adapter code as the one the 'keycloak' fraction is based upon, but it is a next step :-)

Ladicek: So the user must not depend on both of these fractions (`keycloak` and `keycloak-microprofile-jwt`) simultaneously?

fractions/keycloak-microprofile-jwt/src/main/resources/modules/keycloak/io/smallrye/jwt/main/module.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module xmlns="urn:jboss:module:1.3" name="keycloak.io.smallrye.jwt">
+    <resources>
+      <artifact name="org.keycloak:keycloak-core:${version.keycloak}" />
+      <artifact name="org.keycloak:keycloak-common:${version.keycloak}" />
+      <artifact name="org.keycloak:keycloak-adapter-core:${version.keycloak}" />
+      <artifact name="org.keycloak:keycloak-admin-cli:${version.keycloak}" />
```
Ladicek: This looks wrong. There should never be more than 1 module that uses a JAR as its resource root IMHO (otherwise you'll get those weird classloading errors). There are already `org.keycloak` modules that include these artifacts, so this one (which BTW has a very strange name) should have a dependency on them.

fractions/keycloak-microprofile-jwt/src/main/resources/modules/keycloak/io/smallrye/jwt/main/module.xml
```diff
@@ -0,0 +1,17 @@
...
+      <artifact name="com.fasterxml.jackson.core:jackson-core:${version.com.fasterxml.jackson}"/>
+      <artifact name="com.fasterxml.jackson.core:jackson-databind:${version.com.fasterxml.jackson}"/>
+      <artifact name="com.fasterxml.jackson.core:jackson-annotations:${version.com.fasterxml.jackson}"/>
```
Ladicek: I think we've got modules for these as well, but not sure.

fractions/keycloak-microprofile-jwt/pom.xml
sberyozkin: They won't conflict if they do, or if for some reasons they need to support both MP-JWT and KEYCLOAK login types. It is just a 'keycloak' fraction has no effect on the microprofile-jwt flow...

fractions/keycloak-microprofile-jwt/src/main/resources/modules/keycloak/io/smallrye/jwt/main/module.xml
sberyozkin: sure let me check...

fractions/keycloak-microprofile-jwt/src/main/resources/modules/keycloak/io/smallrye/jwt/main/module.xml
sberyozkin: I'll check...

fractions/keycloak-microprofile-jwt/src/main/resources/modules/keycloak/io/smallrye/jwt/main/module.xml
sberyozkin: it worked for the jackson artifacts but not for the keycloak ones, I've tried to use the modules as dependencies instead by bringing the [keycloak adapter feature pack](https://github.com/keycloak/keycloak/tree/master/distribution/feature-packs/adapter-feature-pack) in the pom and using the relevant modules.
My only explanation is that it works for Jackson because it is used indirectly by the KC factory while the classes referenced from the KC factory require an artifact-level support, possibly influenced by the fact ServiceLoader is involved... I'm not too concerned about using the artifacts instead of the modules at the moment.

sberyozkin:
I need to to one one minor optimization to the factory itself though, to avoid creating KeycloakDeployment on every request

sberyozkin:
retest this please

sberyozkin:
retest this please

sberyozkin:
@Ladicek, FYI, I've confirmed locally that adding a 'keycloak' fraction to the list of dependencies in the keycloak-microprofile-jwt test has no effect on the test, it is safe, would not be a problem if someone adds it by accident... Have another look at this PR please

fractions/keycloak-microprofile-jwt/pom.xml
```diff
@@ -0,0 +1,106 @@
...
+    <!-- Provided APIs -->
+    <dependency>
+      <groupId>org.keycloak</groupId>
+      <artifactId>keycloak-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.keycloak</groupId>
+      <artifactId>keycloak-adapter-core</artifactId>
+      <version>${version.keycloak}</version>
+    </dependency>
```
Ladicek, line 54: Are these really meant to be exposed to the user? I removed the `keycloak-admin-cli`, as that looked really weird, but could add that back if that was intentional.

sberyozkin: I only added this one to deal with some of the missing HTTP client API classes :-), so sure it can go away, thanks

Ladicek: But the other two are correct? I mean, does depending on this fraction really mean that some Keycloak APIs are now available to the user? Or is that only for the fraction itself (in which case these deps should be `<scope>provided</scope>`)?

sberyozkin: I've added the provided scope, as indeed Keycloak API should not be exposed (users can use 'keycloak' fraction if they need Keycloak API) thanks

fractions/keycloak-microprofile-jwt/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+    </resources>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.thorntail</groupId>
+      <artifactId>microprofile-jwt</artifactId>
+    </dependency>
+    
+    <!-- Provided APIs -->
```
Ladicek: Please remove this one line.

fractions/keycloak-microprofile-jwt/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+    <dependency>
+      <groupId>org.keycloak</groupId>
+      <artifactId>keycloak-core</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.keycloak</groupId>
+      <artifactId>keycloak-adapter-core</artifactId>
```
Ladicek, line 52: Perhaps should add this to `build-parent/pom.xml` into `<dependencyManagement>`?

fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/deployment/KeycloakJWTCallerPrincipalFactory.java
```diff
@@ -12,21 +13,37 @@
...
         } catch (Throwable ex) {
             throw new ParseException("Failure to parse the token", ex);
         }
     }
+
+    private static KeycloakDeployment createDeployment() {
+        ClassLoader cl = Thread.currentThread().getContextClassLoader();
+        InputStream keycloakJson = cl.getResourceAsStream("keycloak.json");
+        if (keycloakJson == null) {
+            keycloakJson = cl.getResourceAsStream("WEB-INF/keycloak.json");
```
Ladicek: Will this work? Not sure if `WEB-INF` is on "classpath"... Did you try both?

fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/deployment/KeycloakJWTCallerPrincipalFactory.java
```diff
@@ -12,21 +13,37 @@
...
+        }
+        if (keycloakJson == null) {
+            log.warn("keycloak.json resource is not available");
+            return null;
```
Ladicek: Also the `null` return value here will just cause a NPE later, no?

fractions/keycloak-microprofile-jwt/README.adoc
```diff
@@ -0,0 +1,9 @@
+= Keycloak Microprofile JWT
```
Ladicek: In this entire file, `Microprofile` -> `MicroProfile`.

fractions/keycloak-microprofile-jwt/README.adoc
```diff
@@ -0,0 +1,9 @@
...
+
+Uses Keycloak Adapter API to validate the tokens issued by Keycloak and makes these tokens accessible with Microprofile JWT API.
+ 
+It depends on the `keycloak.json` configuration file and ignores the Microprofile JWT configuration properties given that the mapping from the former to the latter is not always possible. It does not use the WildFly Keycloak Adapter subsystem.
+
+Standard Microprofile JWT `MP-JWT` authentication type is required by this fraction.
+
+
```
Ladicek, line 9: Please mention somewhere that, unlike MP JWT, this can redirect to the login page, when you mentioned that, I found it pretty useful for me to understand what this fraction offers.

sberyozkin: For the redirect to work more work needs to be done to overcome the current microprofile-jwt login module constraints, I will open a new issue after this PR goes in

fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/deployment/KeycloakJWTCallerPrincipalFactory.java
sberyozkin: Right, there will be a warning, but no NPE visible (will be caught by the Throwable catch block) - the user will see JWT authentication failed or something like that, this is what microprofile-jwt will log indirectly. Not sure throwing something at the static initialization time is worth it, may be some more work can be done later on, checking the YAML properties, etc. Do you prefer IllegalStateException (or similar) at this stage ?

fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/deployment/KeycloakJWTCallerPrincipalFactory.java
sberyozkin: I don't know, you suggested to add this check on IRC ? Or did you mean something else ?

sberyozkin: Hmm, looks like my response got lost...So, I don't know, you suggested add this check on IRC or did you mean something else ?

fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/deployment/KeycloakJWTCallerPrincipalFactory.java
sberyozkin: My response here got lost too... NPE won't be leaking, the users will see microprofile-jwt logging 'JWT authentication failed' or something like that.  Perhaps, the code creating the KC principal can be enhanced later on to check the system properties (originally set as YAML properties) if the deployment is null...

Ladicek: Ah true, I didn't notice the `catch` block. This is probably fine for now :-)

fractions/keycloak-microprofile-jwt/README.adoc
Ladicek: Ah OK. I think you can open a JIRA already, but sure, doesn't need to be done in this PR.

fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/deployment/KeycloakJWTCallerPrincipalFactory.java
Ladicek: Well I did suggest to look for `WEB-INF/keycloak.json`, as I know this is where `keycloak.json` is supposed to be put. I have no idea if this code will do that correctly :-)

sberyozkin: You are right the code I've added is ineffective.
OK, I've looked at [this code](https://github.com/thorntail/thorntail/blob/master/fractions/keycloak/src/main/java/org/wildfly/swarm/keycloak/runtime/SecuredArchivePreparer.java).
So, the WF keycloak subsystem (=> undertow KC login code) does need keycloak.json in WEB-INF, but this fraction does not because it uses the adapter API directly.
That is not to say this fraction should not support it, but I would not mind looking at as an improvement request as opposed to a blocking issue for this PR. Or would you rather have it sorted out sooner rather than later ?

sberyozkin: I guess it makes sense to sort it out sooner rather than later for users to avoid copying keycloak.json if moving from 'keycloak' to this fraction

Ladicek: Exactly. Also things outside of `WEB-INF` are publicly accessible, which we probably don't want here.

sberyozkin: I don't think it applies, [these keycloak.jsons](https://github.com/thorntail/thorntail-examples/tree/master/security/keycloak/src/main/resources), named differently in the demo, end up in WEB-INF/classes.

Ladicek (approved):

sberyozkin:
retest this please

sberyozkin:
Those failures were not really related but I'll wait for few hours before the next results are available just to see if those test failures are reproducible

PR #1240, opened 2019-03-12
THORN-2287: better support for multi-module projects

michalszynkiewicz:
- [X] Have you followed the guidelines in our [Contributing](https://thorntail.io/community/contributing/) document?
nope but I checked thorntail runner with launcher and one other project and they worked :)

-----

### DIFF
--- a/thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
+++ b/thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
@@ -24,6 +24,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
+import java.util.stream.Collectors;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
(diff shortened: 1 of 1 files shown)


thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
```diff
@@ -115,29 +147,29 @@ private void addWebappResourceToWar(Path path) {
...
                     .filter(File::isFile)
-                    .forEach(file -> addClassToWar(file, classesDirectory));
+                    .forEach(file -> addClassToXar(output, prefix, file, classesDirectory));
         } catch (IOException e) {
             throw new RuntimeException("Failed to add classes to war", e);
         }
     }
 
-    private void addClassToWar(File file, File classesDirectory) {
+    private void addClassToXar(ZipOutputStream output, String prefix, File file, File classesDirectory) {
```
Ladicek: `Xar` -> `Zip` or `Archive`

thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
```diff
@@ -54,9 +55,40 @@ private WarBuilder(ZipOutputStream output, List<String> classesDirs, List<File>
...
+
+    private File buildJar(String classesDir) {
+        try {
+            File jar = File.createTempFile("thorntail-user-jar", ".jar");
+
+            try (FileOutputStream fos = new FileOutputStream(jar);
+                 ZipOutputStream out = new ZipOutputStream(fos)) {
+                addClassesToXar(out, "", new File(classesDir));
+            }
+            return jar;
```
Ladicek, line 82: Where are these files created? I'm thinking if we should somehow take care of deleting them or not.

thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
```diff
@@ -54,9 +55,40 @@ private WarBuilder(ZipOutputStream output, List<String> classesDirs, List<File>
...
+        } catch (Exception any) {
+            throw new RuntimeException("Unable to build jar from directory: " + classesDir);
+        }
+    }
+
+    // the assumption is that the Runner is invoked in the WAR module's directory
```
Ladicek, line 88: I wonder if that assumption makes sense. WAR could possibly be detected by presence of `src/main/webapp` or some other means. And we should perhaps detect if the found more than 1 WAR and bail out in that case. WDYT?

thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
michalszynkiewicz: Zip it is, thanks

thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
michalszynkiewicz: To have all the dependencies visible for `Runner`, one should run the project from the war module. That was my reasoning.

thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
michalszynkiewicz: Good catch, I forgot about `deleteOnExit` for it.
They are created in `/tmp`

thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
Ladicek: Ah OK. Is it possible to detect if the `Runner` was executed from another module and offer an error message? Just thinking out loud.

michalszynkiewicz: I think it would be best to figure it out from Maven or Gradle files.
Otherwise we may check if webapp dir is there but I think it's perfectly okay not to have it.

I added another bullet to https://issues.jboss.org/browse/THORN-2244 regarding figuring out what is a `WAR` and what's not.

Ladicek: Cool, thanks.

Ladicek (approved):

PR #1241, opened 2019-03-12
THORN-2354: wrong license headers in some files in the main codebase

Ladicek:
Motivation
----------
There's a bunch of files in the main Thorntail codebase with
a wrong license header; specifically, the old JBoss LGPL
license header. Thorntail has been LGPL licensed for a short
while in 2015 (back then, it was called WildFly Swarm), but
has switched to Apache 2.0 license on Jun 3rd 2015, see [1].
The old license headers are therefore nothing but a bug and
need to be fixed.

[1] https://github.com/thorntail/thorntail/commit/57e71f2fb1c38a299412ad08e3cdd9ddaacb8412

Modifications
-------------
Replaced LGPL license headers with Apache license headers
in all files where it clearly has been a bug.

Some test files have been rewritten from scratch.

Result
------
No behavioral change.


michalszynkiewicz (dismissed):

Ladicek:
Marking as _hold_ for now, as there's a discussion whether this is needed.

richardfontana:
@Ladicek Looks OK to me.

PR #1244, opened 2019-03-15
THORN-2352: Making it much simpler to configure microprofile-jwt

sberyozkin:
Motivation
----------
Microprofile-jwt login module configuration is complex, would be good to let users type just a couple of YAML lines instead  

Modifications
-------------
MPJWTLoginModuleCustomizer is introduced. It supports two new properties, `thorntail.microprofile.jwtauth.realm` to let users customize the name of the security domain and `thorntail.microprofile.jwtauth.roles.file` to let users customize the location of the roles properties file. `thorntail.microprofile.jwtauth.roles.map` to inline the roles mapping in the YAML can be supported next

Result
------
Significantly simpler microprofile-jwt configuration is possible, with the current configuration format still being supported

### DIFF
--- a/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTLoginModuleCustomizer.java
+++ b/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTLoginModuleCustomizer.java
@@ -67,7 +67,8 @@ public class MPJWTLoginModuleCustomizer implements Customizer {
      * Realm name
      */
     @Configurable("thorntail.microprofile.jwtauth.realm")
-    @AttributeDocumentation("Realm name")
+    @AttributeDocumentation("If set, a security domain with this name that supports MicroProfile JWT is automatically created"
+                             + "in the security subsystem. The realmName parameter of the @LoginConfig annotation must be set to the same value.")
     private Defaultable<String> jwtRealm = string("");
 
     /**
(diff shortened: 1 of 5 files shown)


fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTLoginModuleCustomizer.java
```diff
@@ -0,0 +1,79 @@
...
+                });
+            });
+        }
+    }
+
+    /**
+     * Realm name
+     */
+    @Configurable("thorntail.microprofile.jwtauth.realm")
+    @AttributeDocumentation("Realm name")
```
Ladicek: Is this a reference to an existing security domain, or a name of a new security domain? I mean, as a user, why do I have to configure this?

sberyozkin: This is a [realm name from the MP JWT annotation](https://github.com/thorntail/thorntail/blob/master/testsuite/testsuite-microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/TestApplication.java#L11). This is what the users have to [type in](https://github.com/thorntail/thorntail/blob/master/testsuite/testsuite-microprofile-jwt/src/test/resources/project-empty-roles.yml#L8) here at the moment. So at the MP JWT level it is a 'realm' but on the security fraction level it is 'securityDomain', I thought I'd use the former term for the new property

Ladicek: Ah I see! Could the attribute documentation be improved? Something like

> If set, a security realm with this name is automatically created that supports MicroProfile JWT. The `realmName` parameter of the `@LoginConfig` annotation must be set to the same value. (Under the hood, the created security realm is actually a security domain in the `security` subsystem.)

WDYT?

Ladicek:
retest this please

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTLoginModuleCustomizer.java
sberyozkin: You are right, makes sense to clarify

Ladicek (approved):

Ladicek:
retest this please

Ladicek:
I just found that `thorntail.microprofile.jwtauth` is an old prefix and shouldn't be used :-( It should have been `thorntail.microprofile.jwt`.

PR #1247, opened 2019-03-19
THORN-2360 Update XA datasource driver for MySQL

juangon:
- [X] Have you followed the guidelines in our [Contributing](https://thorntail.io/community/contributing/) document?

-----


Ladicek:
Is that a difference between various MySQL JDBC driver versions?

juangon:
@Ladicek . As what I see it changed since mysql connector Java 6.x onwards (until 5.x used other package)

Ladicek:
Hmm. I wonder if there's a way to support both...

juangon:
Yes, indeed, that would be nice having. Maybe we shouldn't hardcode the XA datasource class at all and just force users to set it through project-defaults.yml.

Ladicek:
I have a solution that is backwards compatible, closing this PR. Thanks for the report and partial solution!

PR #1249, opened 2019-03-20
THORN-2360: fix MySQL JDBC driver autodetection

Ladicek:
Motivation
----------
The XA datasource class in the MySQL JDBC driver has changed
between versions 5.1 and 8.0. We currently always set
the XA datasource class for version 5.1, which doesn't work
when version 8.0 is present.

Modifications
-------------
Set the correct XA datasource class of the MySQL JDBC driver
based on presence of the 8.0 class. If the 8.0 class is present,
it will be used, otherwise the 5.1 class will be set.

Result
------
MySQL JDBC driver autodetection works fine with MySQL JDBC driver
versions 5.1 and 8.0.

-----


Ladicek:
@juangon This is a proper fix, thanks again for the bug report!

juangon (approved):

PR #1258, opened 2019-04-01
[THORN-2368] Support for configuring the roles in the project configuration file

sberyozkin:


fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
```diff
@@ -60,6 +62,31 @@
     @Configurable("thorntail.microprofile.jwt.default-missing-method-permissions-deny-access")
     private Defaultable<Boolean> defaultMissingMethodPermissionsDenyAccess = bool(true);
 
+    /**
+     * Realm name
+     */
+    @Configurable("thorntail.microprofile.jwt.realm")
+    @Configurable("thorntail.microprofile.jwtauth.realm")
```
Ladicek, line 69: Can remove this one. We only have `jwtauth` for backwards compatibility, `jwt` is the correct thing to do.

Ladicek: Ah please ignore, I see it was copied from somewhere else.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
```diff
@@ -131,6 +134,18 @@ public void process() throws Exception {
...
+            createRolePropertiesFileFromMap();
+        }
+    }
+
+    private void createRolePropertiesFileFromMap() {
+        StringBuilder sb = new StringBuilder();
+        for (Map.Entry<String, String> entry : fraction.getRolesPropertiesMap().entrySet()) {
+            sb.append(entry.getKey()).append('=').append(entry.getValue()).append('\n');
+        }
+        archive.add(new ByteArrayAsset(sb.toString().getBytes()), "WEB-INF/classes/roles.properties");
```
Ladicek: Should the filename perhaps be a bit more random, so that we don't risk overwriting an existing one? `roles.properties` is quite predictable. How about `autogenerated-roles.properties`?

Ladicek:
Failure because of Checkstyle:

```
src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java:[83] (regexp) RegexpSingleline: Line has trailing spaces
```

sberyozkin:
Oops I thought I fixed those, Eclipse might've added the one after I updated the docs...

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
sberyozkin: Sure, I'll have a look

Ladicek (approved):

PR #1265, opened 2019-04-05
[THORN-2377] update fraction plugin and make sure we don't have multiple versions of Jackson

Ladicek:

-----

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -22,7 +22,7 @@
   <description>Build Parent to bring in required dependencies</description>
 
   <properties>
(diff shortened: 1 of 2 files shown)


michalszynkiewicz (approved):

roguexz (commented):
That's interesting!! I didn't know we could override the artifact version for ALL modules. This could really help with the Grade plugin as well where let's say some one decided to use a newer library version. It will be great if you could share some documents that talk a little more in detail about overrides, it I could hear start a thread on the Google group.

Ladicek:
We can since today :-) But only during fraction build. That is, this doesn't apply when building an application, only when building Thorntail itself.

PR #1269, opened 2019-04-11
THORN-2385: MP OpenTracing: nested spans aren't reported

Ladicek:
Motivation
----------
Say I have a simple JAX-RS resource like this:

```java
@Path("/")
public class RestSimpleResource {
    @Inject
    private MyService myService;

    @GET
    public Response tracedOperation() {
        return Response.ok().entity(myService.hello()).build();
    }
}
```

It uses a simple service looking like this:

```java
@ApplicationScoped
public class MyService {
    @Inject
    private Tracer tracer;

    @Traced
    public String hello() {
        tracer.activeSpan().log("Hello tracer");
        return "Hello from traced service";
    }
}
```

When MicroProfile OpenTracing is enabled, a single invocation
of the JAX-RS endpoint should produce a single trace with two spans:
one span for the JAX-RS endpoint itself (because JAX-RS endpoints
are traced automatically per MP OpenTracing), and a nested span
for the service invocation (because it's `@Traced`).

This used to work up to Thorntail 2.2.1.Final, but doesn't work
since 2.3.0.Final. The culprit seems to be this change to SmallRye
OpenTracing: https://github.com/smallrye/smallrye-opentracing/pull/25
After this change, the interceptor for handling the `@Traced` calls
no longer comes from SmallRye OpenTracing itself, but from OpenTracing
Java Interceptors project. The necessary integration work in Thorntail
wasn't done when updating to the new SmallRye OpenTracing version.

Modifications
-------------
Properly integrate the new SmallRye OpenTracing version. Specifically,
the OpenTracing Java Interceptors artifact needs to be added to the
SmallRye OpenTracing `module.xml`. Also the interceptor itself
needs to be registered in the CDI extension.

Result
------
The `@Traced` interceptor works correctly.

-----

### DIFF
--- a/fractions/microprofile/microprofile-opentracing/src/main/java/org/wildfly/swarm/mpopentracing/deployment/OpenTracingCDIExtension.java
+++ b/fractions/microprofile/microprofile-opentracing/src/main/java/org/wildfly/swarm/mpopentracing/deployment/OpenTracingCDIExtension.java
@@ -16,11 +16,13 @@
 
 package org.wildfly.swarm.mpopentracing.deployment;
 
+import io.opentracing.contrib.interceptors.OpenTracingInterceptor;
+import org.jboss.logging.Logger;
+
 import javax.enterprise.event.Observes;
 import javax.enterprise.inject.spi.BeanManager;
 import javax.enterprise.inject.spi.BeforeBeanDiscovery;
 import javax.enterprise.inject.spi.Extension;
-import org.jboss.logging.Logger;
 
 /**
  * @author Pavol Loffay
@@ -34,6 +36,7 @@ public class OpenTracingCDIExtension implements Extension {
     String extensionName = OpenTracingCDIExtension.class.getName();
     for (Class<?> clazz : new Class<?>[] {
             TracerProducer.class,
+            OpenTracingInterceptor.class,
     }) {
       bbd.addAnnotatedType(manager.createAnnotatedType(clazz), extensionName + "_" + clazz.getName());
     }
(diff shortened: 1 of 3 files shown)


Ladicek:
@pavolloffay could you please take a look? Thanks!

pavolloffay:
I think you are right the CDI interceptor is not being invoked - so the span for CDI is missing.

I think @jpkrohling moved the interceptor to the separate library. I am not sure what is required to register it in thorntail. In smallrye TCK the interceptor is automatically discovered https://github.com/smallrye/smallrye-opentracing/pull/25/files#diff-1f073f0e30a725752136b28b30d44a59L19

pavolloffay:
@jpkrohling I am wondering if it could be also broken in WF.

Ladicek:
Yea, I've seen that. Just adding the `opentracing-interceptors` artifact didn't help, I had to register the interceptor manually.  I added a test here: https://github.com/rhoar-qe/thorntail-test-suite/commit/4fa31cbb4b0e49141848fe5f8f48d52ffc5cf902

jpkrohling:
> I am wondering if it could be also broken in WF.

It might: WF seems to be using 1.1.1.

Ladicek:
Wow, the MP OpenTracing TCK fails, this probably isn't the best fix.

Ladicek:
Had to fix the TCK runner, as it added all of `io.opentracing` to the deployment, which 1. meant that the interceptor is registered twice, causing all the nice TCK failures; 2. isn't how it works in real applications. The TCK, however, doesn't need all of `io.opentracing`, just `io.opentracing.tracerresolver` and `io.opentracing.mock`, so the fix was pretty simple.

manovotn (approved):
Looks good now, let's see what CI thinks about it!

manovotn:
@pavolloffay @jpkrohling FYI this very likely will affect WFLY as they currently register the old interceptor [here](https://github.com/wildfly/wildfly/blob/master/microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java#L39).

PR #1278, opened 2019-04-29
THORN-2390: Fix the principal leak and prepare for moving the producers code to smallrye-jwt

sberyozkin:
Motivation
----------
ApplicationScoped JAX-RS resources get JsonWebToken injected from the first request only which is a critical security bug in the Thorntail MP JWT integration.

Modifications
-------------
The CDI producer code was synchronized with the one from the Quarkus Smallrye JWT extension. The extra positive side-effect is that it will also make it easier to move the code from both projects into smallrye-jwt.
Additionally PrincipalProducer was enhanced with @Dependent producer method annotation as it appears it is not needed with Quarkus Arc. Nill JsonWebToken is introduced to support the CDI proxy creation when no JsonWebToken is available in the HTTP header.
Finally the redundant JWTAccount has been removed - it is not effective (not used by the deprecated PicketBox as expected) and the workaround has been in place in Thorntail for a while.   

Result
------
JsonWebToken principal does not leak any more, and the CDI producers code is ready to be consolidated in smallrye-jwt


Ladicek (approved):

Ladicek:
I didn't look into it in detail, but looks good on the surface :-)

sberyozkin:
It sounds good :-)

michalszynkiewicz (approved):

michalszynkiewicz:
@sberyozkin it looks good to me and I like that we no longer have a request scoped bean instead of an application scoped bean with a thread local :)
Why exactly was it failing before?

PR #1280, opened 2019-04-30
THORN-2389: Move CDI Producers code into smallrye-jwt

sberyozkin:
Motivation
----------
There are duplicate copies of MP JWT CDI producers in Thorntail and Quarkus, so they need to be consolidated in smallrye-jwt

Modifications
-------------
All the local CDI producers have been removed and the ones from smallrye-jwt are now being used.
Additionally, an unused ClaimLiteral class has also been removed

Result
------

CDI producers code can not be controlled in a single place, smallrye-jwt, and easily reused in Thorntail, Quarkus


sberyozkin:
At the moment it depends on smallrye-jwt 1.1.4-SNAPSHOT

Ladicek (changes requested):
I like removing code! LGTM, but need to wait for SmallRye JWT release, that's the only change I'm requesting :-)

sberyozkin:
Sure, we will need to release smallrye-jwt 1.1.4 first

Ladicek (approved):

PR #1285, opened 2019-05-07
THORN-2396: investigate why some fractions don't have module.conf

Ladicek:
Motivation
----------
Some fractions provide a `module.xml` directly
in `src/main/resources/modules`, which can be wrong and might need
replacing by `module.conf`.

Modifications
-------------
Unified the approach taken for `jaxrs-*` fractions: they shouldn't
use `module.conf`, because they are not really full fractions,
they just extend the `jaxrs` fraction.

All directories under `fractions` that don't have `module.conf`
now have `no.module.conf` file with a short explanation.

Also few unrelated typos are fixed in this commit.

Result
------
No behavioral change.

-----

### DIFF
--- a/fractions/javaee/full/no.module.conf
+++ b/fractions/javaee/full/no.module.conf
@@ -1,0 +1,2 @@
+# This fraction doesn't need module.conf, because it is not a real fraction.
+# It just aggregates fractions on Maven level.
--- a/fractions/javaee/jaxrs-jsonb/src/main/resources/modules/org/wildfly/swarm/jaxrs/jsonb/main/module.xml
+++ b/fractions/javaee/jaxrs-jsonb/src/main/resources/modules/org/wildfly/swarm/jaxrs/jsonb/main/module.xml
@@ -1,0 +1,5 @@
+<module xmlns="urn:jboss:module:1.3" name="org.wildfly.swarm.jaxrs.jsonb">
+  <dependencies>
+    <module name="org.jboss.resteasy.resteasy-json-binding-provider"/>
+  </dependencies>
+</module>
--- a/arquillian/resolver/pom.xml
+++ b/arquillian/resolver/pom.xml
@@ -17,8 +17,8 @@
   <groupId>io.thorntail</groupId>
   <artifactId>arquillian-resolver</artifactId>
 
-  <name> Arquillian Resolver</name>
-  <description> Arquillian Resolver</description>
+  <name>Arquillian Resolver</name>
+  <description>Arquillian Resolver</description>
 
   <packaging>jar</packaging>
 
--- a/arquillian/test/pom.xml
+++ b/arquillian/test/pom.xml
@@ -16,8 +16,8 @@
 
   <artifactId>arquillian-test</artifactId>
 
-  <name> Arquillian Test</name>
-  <description> Arquillian Test</description>
+  <name>Arquillian Test</name>
+  <description>Arquillian Test</description>
 
   <packaging>jar</packaging>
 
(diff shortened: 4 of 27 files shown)


Ladicek:
retest this please

Ladicek:
retest this please

fractions/javaee/full/no.module.conf
```diff
@@ -0,0 +1 @@
+# This doesn't need module.conf, because it is not a real fraction. It just aggregates fractions on Maven level.
```
sberyozkin: May be change `This` to `This component` or `This module` ? if you agree then it would apply to to all the similar comments. Np if you prefer to keep it in the current format

fractions/javaee/jaxrs-jsonb/src/main/resources/modules/org/wildfly/swarm/jaxrs/jsonb/main/module.xml
```diff
@@ -0,0 +1,5 @@
+<module xmlns="urn:jboss:module:1.3" name="org.wildfly.swarm.jaxrs.jsonb">
```
sberyozkin, line 1: I just wonder, would it make sense to document in the Thorntail docs when one fraction needs module.xml or module.conf or both ? I honestly don't understand why the `jsonb` fraction needs a module.xml now, the tests are passing. I guess it is mainly fr consistency purposes but it is only a guess :-)

fractions/javaee/full/no.module.conf
Ladicek: Yea, good point.

fractions/javaee/jaxrs-jsonb/src/main/resources/modules/org/wildfly/swarm/jaxrs/jsonb/main/module.xml
Ladicek: I remember I forced you to add `module.conf` to the `jaxrs-jsonb` fraction, and now I'm moving it back to `module.xml`, oh the irony :-) So, indeed it's for consistency, but if `module.xml` is generated from `module.conf` by the fraction plugin, it contains more module dependencies. Those are not necessary in this case, as I tried to explain in the accompanying `no.module.conf`.

sberyozkin: I forgot about it, np :-). OK, so far as I understand now we'll have a more optimal module.xml for the `jaxrs-jsonb`

Ladicek: I guess what I wanted to say is that we're always learning :-) I inspected all the fractions that don't have a `module.conf`, changed my mind about the `jaxrs-jsonb`, and added a `no.module.conf` file to all directories under `fractions/` that don't have a `module.conf`.

fractions/javaee/full/no.module.conf
Ladicek: Changed to "This fraction".

fractions/javaee/jaxrs-jsonb/src/main/resources/modules/org/wildfly/swarm/jaxrs/jsonb/main/module.xml
sberyozkin: Sounds good

sberyozkin (approved):

fractions/javaee/jaxrs-jsonb/src/main/resources/modules/org/wildfly/swarm/jaxrs/jsonb/main/module.xml
Ladicek: Thanks!

PR #1287, opened 2019-05-07
THORN-2403: Support for smallrye-jwt default group claim and custom token headers

sberyozkin:
Motivation
----------
Make the microprofile-jwt fraction, with the help of smallrye-jwt 1.1.4, support the tokens issued by non-MP JWT aware providers and accept the tokens which may not be coming in HTTP Authorization (httpOnly Cookie is a popular alternative). Both requests originated from the Quarkus users but it can be useful for Thorntail as well   

Modifications
-------------
Support for the tokens without the groups claim is enabled by `smallrye.jwt.claims.groups` property. 
Support for the token in the Cookie header is enabled by the setting `smallrye.jwt.token.header` to `Cookie` and `smallrye.jwt.token.cookie` to the name of the cookie which contains the token.
`JWTAuthMechanism` has been updated to check `Authorization`, `Cookie` and the arbitrary custom headers

Result
------
Better coverage of the practical cases involving the JWT tokens


fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/JWTAuthMechanism.java
```diff
@@ -64,54 +65,74 @@ public JWTAuthMechanism(JWTAuthContextInfo authContextInfo) {
...
+        } else if (COOKIE.toString().equals(authContextInfo.getTokenHeader())
+            && authContextInfo.getTokenCookie() != null) {
+            List<String> cookieValues = exchange.getRequestHeaders().get(authContextInfo.getTokenHeader());
+            if (cookieValues != null) {
+                for (String cookieValue : cookieValues) {
+                    for (String cookie : cookieValue.split(";")) {
+                        if (cookie.startsWith(authContextInfo.getTokenCookie())) {
+                            String[] parts = cookie.split("=");
+                            if (parts.length == 2) {
+                                bearerToken = parts[1].trim();
```
Ladicek: I'd very much prefer to not parse cookies here. There's a proper cookie parser in Undertow, let's use that.

testsuite/testsuite-microprofile-jwt/src/test/resources/project-default-groups.yml
```diff
@@ -0,0 +1,13 @@
+mp:
+  jwt:
+    verify:
+      publickey: MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAlivFI8qB4D0y2jy0CfEqFyy46R0o7S8TKpsx5xbHKoU1VWg6QkQm+ntyIv1p4kE1sPEQO73+HY8+Bzs75XwRTYL1BmR1w8J5hmjVWjc6R2BTBGAYRPFRhor3kpM6ni2SPmNNhurEAHw7TaqszP5eUF/F9+KEBWkwVta+PZ37bwqSE4sCb1soZFrVz/UT/LF4tYpuVYt3YbqToZ3pZOZ9AX2o1GCG3xwOjkc4x0W7ezbQZdC9iftPxVHR8irOijJRRjcPDtA6vPKpzLl6CyYnsIYPd99ltwxTHjr3npfv/3Lw50bAkbT4HeLFxTx4flEoZLKO/g0bAoV2uqBhkA9xnQIDAQAB
+      issuer: "http://testsuite-jwt-issuer.io"
+smallrye:
+  jwt:
+    claims:
+      groups: User
```
Ladicek, line 9: What's the purpose of the `mp.*` and `smallrye.*` settings and can't they not be set under `thorntail.*`?

sberyozkin: I wanted to use and also check that the standard MP JWT properties are effective, they are in smallrye-jwt code so I thought it would be worth to check them out. The new 'smallrye-jwt' properties have no yet MP JWT analog but we have discussed with @starksm64 that the equivalent MP JWT props should also be available, so before starting adding even more properties in Thorntail it is worth waiting a bit

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/JWTAuthMechanism.java
sberyozkin: This code is straightforward meaning that, as far as the goal here is only capture a 'somecookie=token' pair which is not expected to contain anything else. We do not need a mechanism to capture all the cookies and make sure all is good even if no value is available or is some cookie has some extra properties. So is it worth bringing in the Undertow parser ? And if you feel yes, can you point to some example please ?

testsuite/testsuite-microprofile-jwt/src/test/resources/project-default-groups.yml
sberyozkin: Re the new `smallrye` properties, we can have thorntail equivalents but smallrye-jwt 1.1.4 is out now and 1.1.5-SNAPSHOT will need to be tweaked a bit to support the case where a (Thorntail) extension submits its own properties into the smallrye-jwt code setting up the JWAuthContextInfo. But again, may be it is worth waiting a bit

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/JWTAuthMechanism.java
Ladicek: Well, Undertow is already there, if I understand the code correctly, so it wouldn't bring any more dependencies.

Actually now that I'm thinking about it, Undertow has probably already parsed the cookies, so it should be a matter of looking up the correct one from `exchange.getRequestCookies()`, no?

sberyozkin: Let me check :-)

sberyozkin: Worked very well, much simpler now

sberyozkin:
Need #1280 going in first as there will be a conflict in JWTAuthMechanism, so adding a DO NOR MERGE prefix back for now

Ladicek:
I'll merge #1280 tomorrow, or feel free to merge it after tests pass.

Ladicek:
Merged #1280.

Ladicek:
retest this please

Ladicek:
Ah retest was probably a bad idea, there's a merge conflict now :-)

sberyozkin:
Just resolved it, it is only in the pom, probably some extra space added somewhere, I was expecting a conflict in JWTAuthMechanism, but I was confused, the JWTAuthMechanism conflict was only in Quarkus

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/JWTAuthMechanism.java
```diff
@@ -64,54 +65,65 @@ public JWTAuthMechanism(JWTAuthContextInfo authContextInfo) {
...
+            && authContextInfo.getTokenCookie() != null) {
+            Cookie cookie = exchange.getRequestCookies().get(authContextInfo.getTokenCookie());
+            if (cookie != null) {
+                bearerToken = cookie.getValue();
+            }
+        } else {
+            bearerToken = exchange.getRequestHeaders().getFirst(authContextInfo.getTokenHeader());
+        }
+        if (bearerToken != null && UndertowLogger.SECURITY_LOGGER.isTraceEnabled()) {
+            UndertowLogger.SECURITY_LOGGER.tracef("Bearer token: %s", bearerToken);
```
Ladicek, line 122: This looks suspicious. I think we should never ever log secrets, not even on trace level.

sberyozkin: Right, this is optional as far as I can see, i.e, one needs to enable the security logger. I'm assuming this is why UndetowLogger.SECURITY_LOGGER exists. One possible reason is that a log can be used to confirm that a given token has actually been used for some proof

Ladicek: Ah, I see.

sberyozkin:
@Ladicek FYI, I've opened https://issues.jboss.org/browse/THORN-2408 to support the Thorntail specific properties as well, will take care of it once smallrye-jwt 1.1.5 is out. Otherwise this PR looks to be ready to me

Ladicek (approved):

PR #1289, opened 2019-05-09
THORN-2400: Update keycloak-microprofile-jwt to support the adapter YAML config

sberyozkin:
Motivation
----------
A user would like to access KC tokens via MP-JWT API while configuring the adapter in YAML

Modifications
-------------
The archive preparer for `keycloak-micropfile-jwt` has been updated to introspect the equivalent system properties and converting them to keycloak.json with the help of the basic JSON builder. IMHO there is no need (yet) to have some generic JSON builder support for this issue though if really needed it can be provided in due time 

Result
------
`keycloak-micropfile-jwt` fraction can now have the KC adapter configured with `keycloak.json` and the equivalent YAML properties

### DIFF
--- a/fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/runtime/KeycloakMicroprofileJwtArchivePreparer.java
+++ b/fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/runtime/KeycloakMicroprofileJwtArchivePreparer.java
@@ -16,6 +16,7 @@
 package org.wildfly.swarm.keycloak.mpjwt.runtime;
 
 import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Method;
@@ -31,6 +32,10 @@ import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.Node;
 import org.wildfly.swarm.spi.api.DeploymentProcessor;
 import org.wildfly.swarm.spi.runtime.annotations.DeploymentScoped;
+
+import com.fasterxml.jackson.core.JsonEncoding;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
 
 @DeploymentScoped
 public class KeycloakMicroprofileJwtArchivePreparer implements DeploymentProcessor {
@@ -65,7 +70,6 @@ public class KeycloakMicroprofileJwtArchivePreparer implements DeploymentProcess
     private InputStream getKeycloakJsonFromSystemProperties() {
         final String thorntailPrefix = "thorntail.keycloak.secure-deployments.[" + archive.getName() + "].";
         final String swarmPrefix = "swarm.keycloak.secure-deployments.[" + archive.getName() + "].";
-        final String doubleQuote = "\"";
         final String credentialsSecretProperty = "credentials.secret.value";
 
         Map<String, String> kcAdapterSystemProps = new LinkedHashMap<>();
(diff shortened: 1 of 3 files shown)


fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/runtime/KeycloakMicroprofileJwtArchivePreparer.java
```diff
@@ -54,6 +63,63 @@ public void process() throws IOException {
...
 
+    private InputStream getKeycloakJsonFromSystemProperties() {
+        final String thorntailPrefix = "thorntail.keycloak.secure-deployments.[" + archive.getName() + "].";
+        final String swarmPrefix = "swarm.keycloak.secure-deployments.[" + archive.getName() + "].";
+        final String doubleQuote = "\"";
+        final String credentialsSecretProperty = "credentials.secret.value";
+
+        Map<String, String> kcAdapterSystemProps = new LinkedHashMap<>();
+        Properties systemProps = System.getProperties();
+        for (Enumeration<Object> en = systemProps.keys(); en.hasMoreElements();) {
```
Ladicek: `Properties` is also a `Map`, so you can iterate its `entrySet()`.

fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/runtime/KeycloakMicroprofileJwtArchivePreparer.java
```diff
@@ -54,6 +63,63 @@ public void process() throws IOException {
...
+                boolean credentials = entry.getKey().equals(credentialsSecretProperty);
+                if (credentials) {
+                    key = "secret";
+                    sb.append("\"credentials\" : {");
+                }
+                sb.append(doubleQuote).append(key).append(doubleQuote);
+                sb.append(":");
+                String value = !credentials && (isBoolean(entry.getValue()) || isSimpleNumber(entry.getValue()))
+                    ? entry.getValue() : doubleQuote + entry.getValue() + doubleQuote;
+                sb.append(value);
```
Ladicek: Can the value possibly contain something that would require escaping?

(In other words, I don't understand why people still try to assemble JSON by manually concatenating strings, it's _not_ easier than using a library.)

Ladicek:
Otherwise LGTM.

fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/runtime/KeycloakMicroprofileJwtArchivePreparer.java
sberyozkin: Not in [keycloak.json](https://github.com/keycloak/keycloak/blob/master/adapters/oidc/adapter-core/src/test/resources/keycloak.json) as far as I'm aware (it is only simple numbers), this is why I'm not sure it is worth bringing some JSONP or whatever dependency just to build a basic JSON object.

sberyozkin: I.e, since it is a fairly specific task (for example, mapping between a `credentials.secret.value` and a JSON map property, etc), bringing in the library will not make it much simpler, I thought for a moment, OK, I will can write some generic code which accumulates all 'value' property and then generically convert them into JSON maps, etc. but then thought, was it really worth it ? :-), it is only a keycloak.json format which has to be supported

Ladicek: Well, this fraction depends on Keycloak, so it already brings in Jackson Core :-)

sberyozkin: Sure, let me check this one as well, thanks

sberyozkin: @Ladicek, I've replaced the StringBuilder with Jackson JsonGenerator, it did not eliminate the code checking the simple number or boolean format since the input is not typed and is coming from the system properties, but it did help with removing that manual checks for commas, and the object start/end writes, so the code has become cleaner, thanks for suggesting it

sberyozkin: I will only replace Integer with Long

Ladicek (dismissed):

fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/runtime/KeycloakMicroprofileJwtArchivePreparer.java
Ladicek: Great, thanks!

fractions/keycloak-microprofile-jwt/src/main/java/org/wildfly/swarm/keycloak/mpjwt/runtime/KeycloakMicroprofileJwtArchivePreparer.java
sberyozkin: Done

Ladicek (approved):

PR #1297, opened 2019-05-30
THORN-1616: add the "ejb-mdb" fraction

Ladicek:
Motivation
----------
There is no need to include the entire `messaging` fraction
if the application only produces or consumes messages
to/from an external messaging server.

Modifications
-------------
Added a new fraction called `ejb-mdb`, which contains a minimum
subset of `messaging` enough to use MDBs. It expects a resource
adapter to be deployed by the user.

Result
------
Users can now use MDBs connected to external messaging server
without using the entire `messaging` fraction.

-----

### DIFF
--- a/fractions/javaee/ejb-mdb/README.adoc
+++ b/fractions/javaee/ejb-mdb/README.adoc
@@ -1,0 +1,50 @@
+= EJB MDB
+
+Provides support for Message Driven Beans.
+
+For this to work, you need to deploy a resource adapter for an external messaging server.
+The name of this resource adapter must be configured in the `ejb3` subsystem.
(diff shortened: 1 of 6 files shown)


Ladicek:
This is now good for review. Tests are added here: https://github.com/rhoar-qe/thorntail-test-suite/pull/57 (because they require Docker).

Ladicek:
Not totally sure who to ask for review :-)

michalszynkiewicz:
Nothing stands out as wrong here :)

I think e.g. embedded Artemis could be used for tests w/o Docker

michalszynkiewicz (approved):

Ladicek:
Hmm, I could create a "standalone server" with Artemis, similarly to what we have for Keycloak. But those tests couldn't be executed in the product profile, so we'd have to have them in the standalone TS as well :-)

PR #1305, opened 2019-06-24
THORN-2411: implement MicroProfile RestClient 1.3

Ladicek:
Motivation
----------
MP RestClient 1.3 is part of MP 3.0.

Modifications
-------------
SmallRye RestClient was moved to RESTEasy. So this commit
removes all dependencies on SmallRye RestClient and instead
updates RESTEasy to version 3.8.0 which implements
MP RestClient 1.3. This RESTEasy version is newer than
what WildFly 15.0.1 has, so we need to override some
module definitions.

Result
------
RESTEasy 3.8.0 is used and MP RestClient 1.3 is supported.

-----


sberyozkin:
@Ladicek I think it looks very precise. But what happens if someone has a `jaxrs` fraction and `microprofile-rest-client` ones, would there be a risk of different RestEasy version clashing ?

Ladicek:
So that's exactly why the `module-rewrite.conf` replaces a version string of all RESTEasy artifacts in all (`ALL:ALL`) `module.xml` files with 3.8.0.Final. Unless I missed some artifact, there should be no way how multiple RESTEasy versions could come in.

sberyozkin:
OK, so via this mechanism we move to 3.8.0.Final across the board and it would be dropped once we move to WF 17...

Ladicek:
Yes, exactly.

sberyozkin (dismissed):

Ladicek:
> Unless I missed some artifact

I actually found one :-), fixed now.

Ladicek:
I'll investigate the failures, they are all in `testsuite-microprofile-restclient`.

Ladicek:
Test failures have been caused by MP RestClient 1.3 prescribing a default value of the `Accept` header to be `application/json` unless the client interface specifies otherwise. Which is clearly a breaking change.

sberyozkin:
Yes, Produces fixes it. (Might be interesting to check if ClientHeaderParam works instead when no Produces available but it is definitely not related to this upgrade :-) )

sberyozkin (approved):

PR #1306, opened 2019-06-26
THORN-2435: OpenAPI: query parameter "format" value

Ladicek:
Motivation
----------
The MicroProfile OpenAPI specification clearly states that the optional
query string parameter `format` can have 2 values: `YAML` and `JSON`.
Our current implementation defaults to YAML and checks the `format`
parameter (and the `Accept` header) if it should switch to JSON.

However, it doesn't check the `format` parameter correctly. Instead
of checking for `JSON`, it checks for `application/json`.

Modifications
-------------
Check the `format` parameter for `JSON` value as well. The check
is case insensitive.

Result
------
More conformant to the specification. No breaking change,
as the original behavior is retained.

-----


michalszynkiewicz (approved):

michalszynkiewicz:
I must admit I'm surprised this code is not in SmallRye

Ladicek:
Probably because it's Undertow-specific.

PR #1310, opened 2019-06-27
[THORN-2433] Making configuring @LoginConfig realmName optional

sberyozkin:
Motivation
----------
Let users to avoid having to specify the MP JWT realm in the `LoginConfig` annotation if a `thorntail.microprofile.jwt.realm` property is set

Modifications
-------------
`MPJWTAuthExtensionArchivePreparer` has been updated to use the `thorntail.microprofile.jwt.realm` value if `LoginConfig``s `realmName` is empty/null. Also the error is logged if both values are not empty and non equal 

Result
------
No need for users to type the realm name twice in the code and the configuartion

### DIFF
--- a/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
+++ b/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
@@ -68,7 +68,7 @@ public class MicroProfileJWTAuthFraction implements Fraction<MicroProfileJWTAuth
     @Configurable("thorntail.microprofile.jwt.realm")
     @Configurable("thorntail.microprofile.jwtauth.realm")
     @AttributeDocumentation("If set, a security domain with this name that supports MicroProfile JWT is automatically created in the security subsystem."
-                            + " The realmName parameter of the @LoginConfig annotation must be set to the same value.")
+                            + " If the @LoginConfig realName property is configured then it must have the same value.")
     private Defaultable<String> jwtRealm = string("");
 
     /**
--- a/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
+++ b/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
@@ -67,8 +67,8 @@ public class MicroProfileJWTAuthFraction implements Fraction<MicroProfileJWTAuth
      */
     @Configurable("thorntail.microprofile.jwt.realm")
     @Configurable("thorntail.microprofile.jwtauth.realm")
-    @AttributeDocumentation("Defines the security domain which should be used for MicroProfile JWT. This property will be ignored if the security domain with this name already exists. "
-                            + "If this option is set then the realmName property of the @LoginConfig annotation does not have to be configured but if it is then it must have the same value as this option")
+    @AttributeDocumentation("Defines the security domain which should be used for MicroProfile JWT. If no security domain with this name exists, one will be created using sensible defaults. "
+                            + " If this option is set, then the realmName property of the @LoginConfig annotation does not have to be configured; but if it is, then it must have the same value as this option.")
     private Defaultable<String> jwtRealm = string("");
 
     /**
--- a/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
+++ b/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
@@ -67,7 +67,7 @@ public class MicroProfileJWTAuthFraction implements Fraction<MicroProfileJWTAuth
      */
     @Configurable("thorntail.microprofile.jwt.realm")
     @Configurable("thorntail.microprofile.jwtauth.realm")
-    @AttributeDocumentation("Defines the security domain which should be used for MicroProfile JWT but be ignored if the security domain with this name already exists. "
+    @AttributeDocumentation("Defines the security domain which should be used for MicroProfile JWT. This property will be ignored if the security domain with this name already exists. "
                             + "If this option is set then the realmName property of the @LoginConfig annotation does not have to be configured but if it is then it must have the same value as this option")
     private Defaultable<String> jwtRealm = string("");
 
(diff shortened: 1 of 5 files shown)


fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
```diff
@@ -68,7 +68,7 @@
     @Configurable("thorntail.microprofile.jwt.realm")
     @Configurable("thorntail.microprofile.jwtauth.realm")
     @AttributeDocumentation("If set, a security domain with this name that supports MicroProfile JWT is automatically created in the security subsystem."
-                            + " The realmName parameter of the @LoginConfig annotation must be set to the same value.")
+                            + " If the @LoginConfig realName property is configured then it must have the same value.")
```
Ladicek: Typo, `realName` -> `realmName`.

Ladicek:
Otherwise LGTM.

I'm thinking perhaps there might be more than one security domain and the user might want to use a different one than what was created by `thorntail.microprofile.jwt.realm`, but that's probably an unlikely corner case.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
Ladicek: Oh and the text should probably also mention that this realm will automatically be used for JWT auth.

sberyozkin: @Ladicek the 1st part of the doc already says so or would you like to expand/change it ? Something like `MicroProfile JWT realm name. If set, a security domain with this name is automatically created in the security subsystem.` ?

Ladicek: The text says that a security domain is automatically created. But with this PR, not only is it automatically created, but is also automatically used (and in fact must be used). Do I make sense? :-)

sberyozkin:
I'd agree it is a corner case as I'm not sure more than 1 sec domain would ever be configured. This property is not really about facilitating such a configuration, but only about collapsing a given security domain config into a single property for the MP JWT flow. So if someone has LoginConfig with realmName now and 2 security domains then the question which sec domain to use would remain with the verbose and the compact sec domain config options :-)

Ladicek:
Yea. It's not that big of a deal, we should just come up with some text that explains this. Current state of this PR has:

> If set, a security domain with this name that supports MicroProfile JWT is automatically created in the security subsystem. If the @LoginConfig realName property is configured then it must have the same value.

Perhaps we should rephrase that to:

> Defines which security domain should be used for MicroProfile JWT. If no security domain with this name exists, one will be created with sensible defaults. If this option is set, then you don't have to set the realmName property of the @LoginConfig annotation; if you do, however, it must have the same value as this option.

Or something like that. WDYT?

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
sberyozkin: But this is the case even now on the master ? This PR only allows to have LoginConfg realmName be empty

Ladicek: On current master, if you set this option, a security domain will be automatically created, but you still need to configure it in `@LoginConfig(realmName)`. With this PR, you no longer need to set `@LoginConfig(realmName)`.

sberyozkin:
@Ladicek Sure, let me try to adapt it

sberyozkin:
@Ladicek I've found it very hard to try to update the comment :-) Have a look please, if you'd like to change it a bit then please type it in here and I will copy and paste it as is :-)

Ladicek:
If the security domain with given name already exists, the property is ignored _now_, on current master, With this PR, it will no longer be ignored, but will be used for selecting the security domain in case `@LoginConfig(realmName)` is not set.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
```diff
@@ -67,8 +67,8 @@
      */
     @Configurable("thorntail.microprofile.jwt.realm")
     @Configurable("thorntail.microprofile.jwtauth.realm")
-    @AttributeDocumentation("If set, a security domain with this name that supports MicroProfile JWT is automatically created in the security subsystem."
-                            + " The realmName parameter of the @LoginConfig annotation must be set to the same value.")
+    @AttributeDocumentation("Defines the security domain which should be used for MicroProfile JWT. This property will be ignored if the security domain with this name already exists. "
+                            + "If this option is set then the realmName property of the @LoginConfig annotation does not have to be configured but if it is then it must have the same value as this option")
```
Ladicek: I'd use something like this:

> Defines the security domain which should be used for MicroProfile JWT. If no security domain with this name exists, one will be created using sensible defaults.
> If this option is set, then the realmName property of the @LoginConfig annotation does not have to be configured; but if it is, then it must have the same value as this option.

sberyozkin:
It is ignored in this PR for the purpose of **creating** a security domain if the other one already exists the same way is it is done on the current master. 
And there was already a constraint (which is not enforced on the master) that LoginConfig realmName which is used to select the domain must have the same value if this option is set. So if LoginConfig realmName is not set then all this PR does is effectively makes that constraint enforced.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
sberyozkin: Agreed, done

Ladicek:
> there was already a constraint (which is not enforced on the master) that LoginConfig realmName which is used to select the domain must have the same value if this option is set

If that's true, then I didn't notice it. I thought it's perfectly possible to define one security domain using `thorntail.microprofile.jwt.realm`, another security domain using the explicit verbose configuration, and select either one of them using `@LoginConfig`. Better to have that constraint enforced, I think :-)

Ladicek (approved):

sberyozkin:
Hi @Ladicek Oh, the last comment made me think that may be my PR is too restrictive, I mean, there is really no major reason why LoginConfig realmName and thorntail realm properties must be the same, if we do want allow the users this kind of flexibility. I'll be perfectly fine with dropping this comparison in the archive preparer either now if you prefer or later if needed.

Ladicek:
I don't really mind, it's quite a corner case. IMHO this can be merged as is (once Jenkins approves :-) ).

PR #1311, opened 2019-07-08
[THORN-2436] Let users disable MP JWT and make LoginConfig optional if the jwt realm is configured

sberyozkin:
Motivation
----------
Do not require users to add @LoginConfig if the  `thorntail.microprofile.jwt.realm` property is set.
Also add the ability to disable MP JWT similarly to Quarkus.

Modifications
-------------
`thorntail.microprofile.jwt.enabled` property. The security domain selection is skipped if the feature is disabled or if no @LoginConfig with the authentication method `MP-JWT` is found and no `thorntail.microprofile.jwt.realm` is set

Result
------
JAX-RS `Application` will not have to be tied to to the MP JWT API

### DIFF
--- a/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
+++ b/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
@@ -84,7 +84,7 @@ public class MicroProfileJWTAuthFraction implements Fraction<MicroProfileJWTAuth
     @Configurable("thorntail.microprofile.jwt.realm")
     @Configurable("thorntail.microprofile.jwtauth.realm")
     @AttributeDocumentation("Defines the security domain which should be used for MicroProfile JWT. If no security domain with this name exists, one will be created using sensible defaults. "
-                            + " If this option is set, then the realmName property of the @LoginConfig annotation does not have to be configured; but if it is, then it must have the same value as this option.")
+                            + " If this option is set, then the @LoginConfig annotation is not needed but if it is available then its realmName property, if set, must have the same value as this option.")
     private Defaultable<String> jwtRealm = string("");
 
     /**
@@ -103,8 +103,7 @@ public class MicroProfileJWTAuthFraction implements Fraction<MicroProfileJWTAuth
     @AttributeDocumentation("Roles properties map")
     private Map<String, String> rolesPropertiesMap;
 
-    @AttributeDocumentation("If both this and 'thorntail.microprofile.jwt.realm' properties are enabled then the @LoginConfig annotation does not have to be present."
-                            + "If this option is disabled then the MP JWT authentication mechanism will not be activated")
+    @AttributeDocumentation("Set this to false to disable the MP JWT authentication mechanism. Defaults to true.")
     @Configurable("thorntail.microprofile.jwt.enabled")
     private Defaultable<Boolean> jwtEnabled = bool(true);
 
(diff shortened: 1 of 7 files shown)


fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
```diff
@@ -92,16 +97,19 @@ public void process() throws Exception {
                 realm = realmNameProp.asString();
             }
             // Set the web.xml login-config auth-method and jboss-web.xml security domain
-            if (authMethod != null) {
-                WebXmlAsset webXml = war.findWebXmlAsset();
-                webXml.setLoginConfig(authMethod.asString(), realm);
```
Ladicek, line 97: Removing this means it will no longer be possible to set a different auth method than `MP-JWT`. Which was probably never intended, but it seems it was possible. Not sure about consequences.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
```diff
@@ -103,6 +103,19 @@
     @AttributeDocumentation("Roles properties map")
     private Map<String, String> rolesPropertiesMap;
 
+    @AttributeDocumentation("If both this and 'thorntail.microprofile.jwt.realm' properties are enabled then the @LoginConfig annotation does not have to be present."
+                            + "If this option is disabled then the MP JWT authentication mechanism will not be activated")
+    @Configurable("thorntail.microprofile.jwt.enabled")
+    private Defaultable<Boolean> jwtEnabled = bool(true);
```
Ladicek, line 108: So, do we really need this? It's enabled by default, so users don't have to enable it explicitly. And if they include the MP JWT fraction, they probably want MP JWT :-)

If we want this, I'd rephrase the documentation. Something like:

> Set this to false to disable the MP JWT authentication mechanism. Defaults to true.

And if you want to mention that `@LoginConfig` is not necessary, I'd put it to the documentation of `jwtRealm`, which already mentions that if set, the `realmName` property of the `@LoginConfig` annotation is optional.

Ladicek:
Otherwise LGTM.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
sberyozkin: @Ladicek yes, the documentation for this annotation says that more than `MP-JWT` is possible but within the scope of `microprofile-jwt` only `MP-JWT` matters, the other standard methods like `Basic` etc are not really relevant

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
sberyozkin: @Ladicek I'm not 100% sure now myself if it is needed. It can be handy I guess if a user wants to ensure it is not activated against random `Application` without `LoginConfig`, as an extra safety switch, or, just disable it anyway as Quarkus can do, so on balance I guess it should be OK to keep it.

sberyozkin: @Ladicek have a look please at the updated docs, the hardest part as usual for me :-)

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
```diff
@@ -84,7 +84,7 @@
     @Configurable("thorntail.microprofile.jwt.realm")
     @Configurable("thorntail.microprofile.jwtauth.realm")
     @AttributeDocumentation("Defines the security domain which should be used for MicroProfile JWT. If no security domain with this name exists, one will be created using sensible defaults. "
-                            + " If this option is set, then the realmName property of the @LoginConfig annotation does not have to be configured; but if it is, then it must have the same value as this option.")
+                            + " If this option is set, then the @LoginConfig annotation is not needed but if it is available then its realmName property, if set, must have the same value as this option.")
```
Ladicek: available -> present

sberyozkin: Sorry, missed it, just updated

Ladicek (approved):

PR #1315, opened 2019-07-10
[THORN-2434] Add signer-pub-key-location, deprecate jwks.uri and the use of signer-pub-key for key assets

sberyozkin:
Motivation
----------
Inconsistent use of properties between Thorntail and smalrye-jwt to refer to the inlined and external key content. 
Please see the [THORN-2434](https://issues.jboss.org/browse/THORN-2434) for more details. 

Modifications
-------------
Using `thorntail.microprofile.jwt.token.sub-pub-key` to refer to the external key assets, as well as `thorntail.microprofile.jwt.token.jwks-uri` to set HTTPS JWK set URI has been deprecated.
 
`thorntail.microprofile.jwt.token.sub-pub-key-location` has been introduced.

The user of all legacy `mpjwt.*` has been deprecated but only at the log message level for now. 

Two tests which were only checking the the classpath and file based resources could be loaded were removed, the existing tests have been updated to check that such key assets are actually effective, and a new test testing `thorntail.microprofile.jwt.token.sub-pub-key-location` has been added

Result
------
Hopefully it will become less confusing going forward to deal with configuring the key assets


fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
```diff
@@ -40,21 +40,26 @@
     @Configurable("thorntail.microprofile.jwtauth.token.issuedBy")
     private Defaultable<String> tokenIssuer = string("http://localhost");
 
-    @AttributeDocumentation("The public key of the JWT token signer. Can be prefixed 'file:' or 'classpath:' for key assets, otherwise the key contents are expected")
+    @AttributeDocumentation("The public key of the JWT token signer. Using it for the 'file:' or 'classpath:' key assets is deprecated, use the 'thorntail.microprofile.jwt.token.signer-pub-key-location' property instead")
     @Configurable("thorntail.microprofile.jwt.token.signer-pub-key")
     @Configurable("thorntail.microprofile.jwtauth.token.signerPubKey")
     private String publicKey;
 
+    @AttributeDocumentation("Location of the public key of the JWT token signer.")
```
Ladicek: In what format?

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
```diff
@@ -40,21 +40,26 @@
     @Configurable("thorntail.microprofile.jwtauth.token.issuedBy")
     private Defaultable<String> tokenIssuer = string("http://localhost");
 
-    @AttributeDocumentation("The public key of the JWT token signer. Can be prefixed 'file:' or 'classpath:' for key assets, otherwise the key contents are expected")
+    @AttributeDocumentation("The public key of the JWT token signer. Using it for the 'file:' or 'classpath:' key assets is deprecated, use the 'thorntail.microprofile.jwt.token.signer-pub-key-location' property instead")
```
Ladicek: I'd prefer this:

> The public key of the JWT token signer. Can be prefixed 'file:' or 'classpath:' to refer to external assets, but this is deprecated; use 'thorntail.microprofile.jwt.token.signer-pub-key-location' instead.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/JWTAuthMechanismFactory.java
```diff
@@ -81,18 +81,22 @@ public AuthenticationMechanism create(String mechanismName, FormParserFactory fo
...
+                String publicKeyLocation = getResource(properties, "signerPubKeyLocation", "MP-JWT-SIGNER-KEY-LOCATION");
+                if (publicKeyLocation == null) {
+                    publicKeyLocation = getResource(properties, "jwksUri", "MP-JWT-SIGNER-KEY-LOCATION");
+                }
+                if (publicKeyLocation != null) {
+                    contextInfo.setPublicKeyLocation(publicKeyLocation);
+                    if (publicKeyLocation.startsWith("https:")) {
+                        String jwksRefreshInterval = getResource(properties, "jwksRefreshInterval", "MP-JWT-JWKS-REFRESH");
+                        if (jwksRefreshInterval == null) {
+                            throw new IllegalStateException("JWKS Refresh Interval should be set when JWKS URI is used.");
```
Ladicek: Should we say "must be set" instead of "should"?

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/config/JWTAuthContextInfoProvider.java
```diff
@@ -80,6 +80,8 @@
                 NONE.equals(super.getMpJwtPublicKey().get()) && NONE.equals(super.getMpJwtLocation().get())) {
             return Optional.empty();
         }
+        log.warn("The use of all mpjwt.* properties is deprecated, use thorntail.microprofile.jwt.* properties instead");
```
Ladicek, line 83: I think the MP JWT spec defines some of these properties, no? Hmm, actually no, the MP JWT spec defines a couple of `mp.jwt.verify.*` properties. Where do we handle those?

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
```diff
@@ -119,35 +119,44 @@ public void process() throws Exception {
...
             war.addAsManifestResource(EmptyAsset.INSTANCE, "MP-JWT-DENY-NONANNOTATED-METHODS");
         }
 
         if (fraction.getJwksUri() != null) {
+            log.warn("Using 'thorntail.microprofile.jwt.token.jwks-uri' for the HTTPS based JWK sets is deprecated, "
+                    + "use the 'thorntail.microprofile.jwt.token.signer-pub-key-location' "
+                    + "property instead");
             log.debugf("JwksUri: %s", fraction.getJwksUri());
-            war.addAsManifestResource(new StringAsset(fraction.getJwksUri()), "MP-JWT-JWKS");
+            war.addAsManifestResource(new StringAsset(fraction.getJwksUri()), "MP-JWT-SIGNER-KEY-LOCATION");
```
Ladicek: Should we only add `MP-JWT-SIGNER-KEY-LOCATION` if one doesn't exist? This code would mean that `thorntail.microprofile.jwt.token.jwks-uri` would override `thorntail.microprofile.jwt.token.signer-pub-key-location`, if both are specified.

(This becomes quite complex already...)

Ladicek:
Also, I don't like deleting tests that should still be working.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/config/JWTAuthContextInfoProvider.java
sberyozkin: In smallrye-jwt, in the base provider.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/JWTAuthMechanismFactory.java
sberyozkin: I'm not even sure this null check is needed, the code was there before so I just kept it, there is a default value of 60 mins.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/config/JWTAuthContextInfoProvider.java
Ladicek: Ah sure, that should have been obvious :-) Thanks!

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
sberyozkin: I think may be I should ignore the jwksUri value if the location is set, but the same could be said if the signer-pub-key is set. Agree it is not becoming simpler, given that it is only a deprecation so there is need to support overlapping properties. I don't really mind much even abandoning this PR (though the current state is not close to the perfection either :-) ), let me tweak few things and then see how sound it is.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/JWTAuthMechanismFactory.java
Ladicek: I'm not sure if we should throw an exception if there's a default, but well...

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java
Ladicek: Yea, agree, overlapping properties are more of a corner case.

sberyozkin:
@Ladicek the test which was loading the classpath resource was depending on [this code](https://github.com/thorntail/thorntail/blob/master/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/runtime/MPJWTAuthExtensionArchivePreparer.java#L126). But this code has gone in this PR and the handling of such a resource is now taken care by smallrye-jwt `KeyLocationResolver`. Furthermore this test was not testing the key was actually used but only that is was available to load but it does not test anything but it can be loaded, not that it is used by MP-JWT, so it would be needed to refactor the test to do what others tests already do, sign the token, etc, so I just updated one of the existing tests to use a classpath resource.
The same story is about the file based resource test. But at least it would still work if I kept it. I can restore it if you'd like ? I've updated one of the test to actually test that the file based key is effective.
Let me know please what you prefer

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/JWTAuthMechanismFactory.java
sberyozkin: I'll just update it to set the context only if it is not null, Jose4j code has also some default value

sberyozkin:
@Ladicek have a look please; can tweak the docs further, and happy enough to restore at least the file based key loading test (as I mentioned, the file based resource is being tested now with some other test, but I can restore it - but I'll have to remove after the upgrade to `smallrye-jwt-1.1-2.0.1` anyway :-) because the file based resource will be taken care of there)

Ladicek:
@sberyozkin You're right, I took a look and the deleted tests were quite useless. Agree to remove them.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
```diff
@@ -40,21 +40,26 @@
     @Configurable("thorntail.microprofile.jwtauth.token.issuedBy")
     private Defaultable<String> tokenIssuer = string("http://localhost");
 
-    @AttributeDocumentation("The public key of the JWT token signer. Can be prefixed 'file:' or 'classpath:' for key assets, otherwise the key contents are expected")
+    @AttributeDocumentation("The public key of the JWT token signer. Can be prefixed 'file:' or 'classpath:' to refer to external assets, but this is deprecated; use 'thorntail.microprofile.jwt.token.signer-pub-key-location' instead")
     @Configurable("thorntail.microprofile.jwt.token.signer-pub-key")
     @Configurable("thorntail.microprofile.jwtauth.token.signerPubKey")
     private String publicKey;
 
+    @AttributeDocumentation("Location of the public key of the JWT token signer. The key can be in PEM or JWK formats, in the latter case it can be part of the JWK Set")
```
Ladicek: I meant in what format can the location be :-) It seems `file:` and `classpath:` should still work, also `https://` should work for the JWKS, is there something else?

sberyozkin: @Ladicek Oh I see, will update. If no scheme is used then it will be indirecty defaulted to the `classpath:` in smalrye-jwt. **Update** it is not a default, because smallrye-jwt is expected to try both classpath and file assets if no scheme is used. But as I mentioned earlier, `file:` is not yet supported at the smallrye-jwt level, so I need to do few more tweaks to ensure this property can be used for file based assets

Ladicek: OK, thanks!

sberyozkin:
@Ladicek this should be better tested now, and as always, open for more doc/etc tweaks :-)

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
```diff
@@ -40,21 +40,26 @@
     @Configurable("thorntail.microprofile.jwtauth.token.issuedBy")
     private Defaultable<String> tokenIssuer = string("http://localhost");
 
-    @AttributeDocumentation("The public key of the JWT token signer. Can be prefixed 'file:' or 'classpath:' for key assets, otherwise the key contents are expected")
+    @AttributeDocumentation("The public key of the JWT token signer. Can be prefixed 'file:' or 'classpath:' to refer to external assets, but this is deprecated; use 'thorntail.microprofile.jwt.token.signer-pub-key-location' instead")
     @Configurable("thorntail.microprofile.jwt.token.signer-pub-key")
     @Configurable("thorntail.microprofile.jwtauth.token.signerPubKey")
     private String publicKey;
 
+    @AttributeDocumentation("Location of the public key of the JWT token signer. Can be prefixed 'file:', 'classpath:' (optional for the classpath resources) or 'https:'. Only JWK Sets can be HTTPS location based.")
```
Ladicek: OK, so if this is classpath resource by default (no prefix), the `classpath:` prefix is basically ignored, or can be `file:`, or `https://` URL in case of JWKS, I'd probably say:

> Location of the public key of the JWT token signer. By default, or when the 'classpath:' prefix is present, this is a classpath resource. Can be prefixed with 'file:' to refer to an external file. Can also be a HTTPS URL of a JWK Set.

WDYT?

Ladicek:
Otherwise LGTM.

fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/MicroProfileJWTAuthFraction.java
sberyozkin: More precise, looks good to me. 1 sec... Done

Ladicek (approved):

Ladicek:
Windows failure is a known flaky test in MP FT TCK (`BulkheadAsynchRetryTest.testBulkheadMethodAsynchronousRetry55`). Gonna merge.

PR #1316, opened 2019-07-12
THORN-2399: move all version-related Maven properties to the root POM

Ladicek:
Motivation
----------
Currently, versions of components are scattered through various POMs,
most of them are in `/pom.xml`, `/build-parent/pom.xml` and
`/testsuite/pom.xml`. This is hard to navigate, and can be radically
simplified by just moving all of them to the root POM.

Modifications
-------------
Moved all version properties to the root POM. Nothing but Maven properties
moves; specifically `<dependencyManagement>` doesn't not change.

Result
------
All component versions are specified on a single place. No behavioral change.

-----

### DIFF
--- a/arquillian/adapter/pom.xml
+++ b/arquillian/adapter/pom.xml
@@ -104,18 +104,6 @@
       <artifactId>junit</artifactId>
       <scope>compile</scope>
     </dependency>
-    <dependency>
-      <groupId>org.mockito</groupId>
-      <artifactId>mockito-all</artifactId>
-      <version>1.10.19</version>
-      <scope>test</scope>
-      <exclusions>
-        <exclusion>
-          <groupId>org.hamcrest</groupId>
-          <artifactId>hamcrest-core</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
 
   </dependencies>
 
--- a/arquillian/daemon/pom.xml
+++ b/arquillian/daemon/pom.xml
@@ -22,13 +22,6 @@
 
   <packaging>jar</packaging>
 
-<!--
-  <properties>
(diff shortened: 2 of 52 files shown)


sberyozkin:
@Ladicek but why the root pom as opposed to the build-parent/pom.xml ?

Ladicek:
Because some modules (notably test suite) don't (on purpose) inherit from `build-parent`. I want to have the version numbers on a single place, and this is the only possible one I can think of. (It still leaves out `docs/howto`, which doesn't inherit from anything, but that's manageable I think.)

sberyozkin:
@Ladicek sure, getting them all set in a single place is good, but then `build-parent/pom.xml` won't really be a real build-parent, it will only have a dependency management section which testsuites still won't be able to depend upon. Can you expand please on why `testsuites` do not depend on `build-parent/pom.xml`. Just would like to understand...

Ladicek:
`build-parent` is a parent of the main build, but not of the test suite. This is, if I remember correctly, to isolate the test suite from the dependency management of the main build. (My personal opinion is that the test suite should be in a completely different Maven reactor, but that would complicate the CI job quite a bit.)

sberyozkin:
OK, it is clearer now.

sberyozkin (approved):

Ladicek:
Thanks! And sorry, I'm not able to exactly articulate the purpose of `build-parent`. You can take a look at THORN-1440 and commit `c90723af0ed1f2c3ac64f65d5de2f10a5a1f0410` for more info, though it's not really comprehensive.

Ladicek:
I'm not sure what I just did, I just wanted to remove the request for Michal's review :-) Anyway, will merge after CI finishes.

sberyozkin:
@Ladicek no problems, thanks for the info

PR #1327, opened 2019-08-15
THORN-2457: update Guava and remove Netflix Guava module

Ladicek:
Motivation
----------
We use an ancient version of Guava. We should update it
to something more modern, ideally aligned to our base WildFly
version. Since Thorntail is still based on WildFly 15.0.1,
we should use Guava 25.0, per [1].

And since WildFly already ships a module for Guava (see [2]),
we can as well get rid of the special Netflix Guava module.

[1] https://github.com/wildfly/wildfly/blob/15.0.1.Final/pom.xml#L221
[2] https://github.com/wildfly/wildfly/blob/15.0.1.Final/feature-pack/src/main/resources/modules/system/layers/base/com/google/guava/main/module.xml

Modifications
-------------
Update Guava version and remove the Netflix Guava module.

Result
------
No behavioral change. Slightly easier maintenance.

-----


Ladicek:
Apparently this version of Guava is too new for Selenium (which is brought in by Arquillian Drone). Coincidentally, there was a bug filed recently (THORN-2454) that I'd ultimately want to fix by removing Arquillian Drone completely. It's of no use to us, some plain HTTP Client (we tend to use `fluent-hc`) is a much better fit for that kind of tests we use Arquillian Drone for.

Ladicek:
Ah, the failures are all in HOWTOs, and those manage Guava version on their own. So it's a simple revert :-)

michalszynkiewicz (approved):

Ladicek:
I also ran the standalone test suite locally, and it passes. Merging.

PR #1329, opened 2019-08-21
THORN-2459: multiple SmallRye Config versions in uberjar, old one used at runtime

Ladicek:
Motivation
----------
With a Thorntail application that uses MicroProfile Config,
the resulting uberjar contains two versions of SmallRye Config:
`smallrye-config-1.3.4.jar` and `smallrye-config-1.3-1.0.0.jar`.
At runtime, the former (which is older) is used.

This is because the `microprofile-config-wildfly-feature-pack`
in `fractions/microprofile/microprofile-config/pom.xml` comes
_before_ `wildfly-feature-pack`, and so `wildfly-feature-pack`
wins when it comes to who provides the SmallRye Config `module.xml`.
The order of declarations of feature pack dependencies needs to change.

Modifications
-------------
The `microprofile-config-wildfly-feature-pack` dependency now
comes _after_ `wildfly-feature-pack`, and so it wins when
searching for `module.xml` of SmallRye Config.

Result
------
Correct implementation of MP Config is now used.
Only one SmallRye Config JAR in the uberjar.

-----

### DIFF
--- a/fractions/microprofile/microprofile-config/pom.xml
+++ b/fractions/microprofile/microprofile-config/pom.xml
@@ -97,19 +97,6 @@
             </exclusions>
         </dependency>
         <dependency>
-            <groupId>io.thorntail</groupId>
-            <artifactId>microprofile-config-wildfly-feature-pack</artifactId>
-            <version>${version.microprofile-config-wildfly}</version>
-            <type>zip</type>
-            <scope>provided</scope>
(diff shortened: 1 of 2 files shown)


Ladicek:
I kinda expected the MP Config TCK to fail, as we see it internally as well. Will dig into this.

PR #1334, opened 2019-09-11
THORN-2466: Demonstrate unused ObjectMapper

samolsen:
Motivation
----------
Add unit tests demonstrating a customized
Jackson ObjectMapper is not used by JAX-RS

Modifications
-------------
Added unit tests demonstrating this bug

Result
------
Causes build to fail due to incorrect behavior

-----


Ladicek:
ok to test

Ladicek:
I just pushed 2 commits to your branch, the 1st makes tests in this module considerably simpler, and the 2nd fixes your problem. Its commit message includes details. WDYT?

samolsen:
That looks great to me, and the commit explanation really clarifies how the framework behaves. Thank you!

Ladicek:
Great, thanks! I'll let Jenkins finish the tests and, assuming they pass, merge this PR.

roguexz:
Thank you for the clarification @Ladicek. I was running in to this issue a couple of days ago and was wondering why my tests were failing (when I distinctly remember) them passing earlier :)

Ladicek (approved):

Ladicek:
Thanks!

roguexz:
Also a small observation. It is better to instantiate the ObjectMapper once and then use that instance, instead of creating it every time the provider is instantiated.

I ran into a peculiar big where I was trying to make multiple REST calls using ForkJoin and I would see random failures around de-serialization. I had custom Jackson extensions and modules, which wouldn't load due to some race condition. So I moved the object mapper instantiation to a static block and reused that instance.

Ladicek:
Good point!

PR #1337, opened 2019-09-14
Support for BOM style dependencies in the Gradle plugins

roguexz:

-----
@TomasHofman @Ladicek 

I have been playing around with consuming the Gradle's version of [BOM definitions](https://docs.gradle.org/current/userguide/java_platform_plugin.html) (a.k.a. `java-platform`) and I noticed that the plugin always returns the bom parent as having the artifact type as `jar` and then that fails during dependency resolution.

Since in the Gradle plugin, all the dependencies have been resolved, we can skip any dependency that does not have a resolved artifact.

This assumes that developers can not create a Gradle project that has no outputs but has dependencies, in which case the work around would be to just include those dependencies directly. If we know of additional use cases, we can always factor them in at that point.


Ladicek:
test this please

TomasHofman:
@roguexz I'm don't know much about this part of the plugin, but as you describe it it sounds reasonable.

Do you have some test project where this is demonstrated?

Ladicek (approved):
I don't know anything about Gradle, but there's nothing obviously wrong in this PR :-)

roguexz:
I'll raise a separate PR to the examples project the demonstrate this behavior :)

Ladicek:
Example would be great, thanks!

PR #1340, opened 2019-10-01
THORN-2224: allow thorntail runner to work behind proxy

ghost:
# Issue
It appears CachingArtifactResolvingHelper is ignoring the ${user.home}/.m2/settings.xml mirror values and going to hosts direct via it's own list which is then failing as the hosts are not available.

- Typically this issue occurs when running the runner behind a proxy (enterprise). As you cannot disable the hardcoded repositories building the cache takes ages to load. Behind the scenes it just times out for all dependencies.
- To ensure ensure we can also enforce security policies (eg. force use of Nexus rather then online repositories) we should be able to disable these hardcoded repositories

# Adaption
- **Change 1: added extra property**
  - Description:
    Added a property to disable default repositories 'thorntail.runner.ignore.default.repositories'
  - How to use:
    - Set system properties as usual (similar to other configurable properties)
    - No value needs to be set for this property, if present this means the default hardcoded repositories are ignored
  - Motivation:
    - Allows you to enforce security policies
    - Avoids blocked calls from proxy to even occur
    - In combination with user repository property you can nicely force dependency resolving to use your own artifact registry (ex. Nexus)
- **Change 2: allow property file for runner cache configuration**
  - Description:
    Allow users to create a file called 'thorntail-runner-cache.properties' that will set the properties for the cache
  - How to use:
    - Create a file in the root of your application called 'thorntail-runner-cache.properties'
    - You can choose now to configure your cache properties in this file rather than using VM arguments
  - Motivation:
    - No more need to add it to VM configuration (no volatile IDE project configuration)
    - Can be checked in into project repository for ease of configuration en setup without bothering developer

# Feedback
No breaking changes made here to make it work. An idea I was still playing with was to allow all 'thorntail runner' properties to be set through the use of a property file. Something I should extend to all properties for the thorntail runner?

# Checklist


Ladicek:
ok to test

ghost:
> Do we have to check for known names? Could we just set everything that's in the file?

My thought exactly, but this does mean that:
- the file would be renamed to 'thorntail-runner.properties' 
- we extend functionality to configure all runner properties

I'm fine with all of the above :).. I'll make the adaption

ghost:
> This looks like we're overriding system properties with the values from the properties file, it should be the other way around.

You're absolutely right, the way that it's done now lowers the readability

thorntail-runner/src/main/java/org/wildfly/swarm/runner/RunnerConstants.java
```diff
@@ -0,0 +1,57 @@
...
+    public static final String IGNORE_DEFAULT_REPOSITORIES = loadProperty("thorntail.runner.ignore.default.repositories", null);
+    public static final String USER_REPOSITORIES = loadProperty("thorntail.runner.repositories", null);
+    public static final String PRESERVE_JAR = loadProperty("thorntail.runner.preserve-jar", null);
+
+    private RunnerConstants() {
+    }
+
+    private static String loadProperty(String propertyName, String defaultValue) {
+        String propertyValue;
+        try (InputStream st = new FileInputStream("thorntail-runner.properties")) {
```
Ladicek, line 49: This will unnecessarily load the file for each property, again and again. How about creating a `static` block that will initialize all the constants at once?

ghost: Adapted code, I was a bit twisted in just creating a static block.. I rather chose to rework it to to something cleaner (where constants and configurations logic are nicely seperated). Let me know what you think

Ladicek:
This gets unnecessarily complex. I'll take a look at it later.

TPX01:
> This gets unnecessarily complex. I'll take a look at it later.

You'd rather have something like this?

```
public class RunnerConstants {
    public static final String CACHE_STORAGE_DIR;
    public static final String IGNORE_DEFAULT_REPOSITORIES;
    public static final String USER_REPOSITORIES;
    public static final String PRESERVE_JAR;

    static {
        Properties fileProperties = new Properties();
        try (InputStream st = new FileInputStream("thorntail-runner.properties")) {
            fileProperties.load(st);
        } catch (FileNotFoundException e) {
            System.out.println("No configuration (thorntail-runner.properties) found");
        } catch (IOException e) {
            System.out.println("Could not read configuration (thorntail-runner.properties), ignoring file");
        }

        CACHE_STORAGE_DIR = getProperty(fileProperties , "thorntail.runner.cache-location", ".thorntail-runner-cache");
        IGNORE_DEFAULT_REPOSITORIES = getProperty(fileProperties, "thorntail.runner.ignore.default.repositories", null);
        USER_REPOSITORIES = getProperty(fileProperties,"thorntail.runner.repositories", null);
        PRESERVE_JAR = getProperty(fileProperties, "thorntail.runner.preserve-jar", null);
    }

    private RunnerConstants() {
    }

    private static String getProperty(Properties fileProperties, String key, String defaultValue) {
        String value = null;

        if (fileProperties.getProperty(key) !=null) {
            value = fileProperties.getProperty(key);
        }
        if (System.getProperty(key) != null) {
            value = System.getProperty(key);
        }
        return value != null ? value : defaultValue;
    }
}
```

Ladicek:
That looks just about perfect to me :-)

Ladicek (dismissed):

Ladicek:
Let's wait for CI to finish, but looking good now. Thanks!

ghost:
> That looks just about perfect to me :-)

I admit, I rather over-engineered the whole configuration part.. ;)

Ladicek:
I took the liberty to take your branch, squash all the commits, provide a more detailed commit message, and most importantly, make the property name consistent with other Runner properties. So it shall be `thorntail.runner.ignore-default-repositories`, not `thorntail.runner.ignore.default.repositories`. When the CI finishes again, I'll merge the PR. Thanks again!

Ladicek (approved):

ghost:
Thanks for cleaning up these commits Ladicek👌

PR #1341, opened 2019-10-04
Gradle: Align the Gradle version during Arquillian dependency resolution

roguexz:
Previously, the Gradle Arquillian integration would always spin up a Gradle 3.1 daemon. This was becuase we were pulling in the Gradle jar (version 3.1) as a dependency. This has a performance implication on the CI equation as each build would download Gradle 3.1 for testing.

With this commit, I am making the following changes,

1. Mark all Gradle dependencies as provided.
2. Include the required dependencies (from the Gradle platform) dynamically as part of the plugin configurations.
3. Introduce a path forward for adding Gradle specific changes without having to rely on bumping the Gradle library versions.

-----

-----

### DIFF
--- a/plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailArquillianPlugin.java
+++ b/plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailArquillianPlugin.java
@@ -44,6 +44,8 @@ public class ThorntailArquillianPlugin extends AbstractThorntailPlugin {
 
     private static final String BUILD_SCRIPT_FRAGMENT = "META-INF/thorntail-scripts/configure-tasks.groovy";
 
+    private boolean projectConfigured = false;
+
     /**
      * Constructs a new instance of {@code PackagePlugin}, which is initialized with the Gradle tooling model builder registry.
      *
@@ -57,6 +59,12 @@ public class ThorntailArquillianPlugin extends AbstractThorntailPlugin {
 
     @Override
     public void apply(Project project) {
+        if (projectConfigured) {
+            // Nothing to do.
+            return;
+        }
+        projectConfigured = true;
+
         super.apply(project);
 
         //noinspection Convert2Lambda
--- a/plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailArquillianPlugin.java
+++ b/plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailArquillianPlugin.java
@@ -16,11 +16,23 @@
 
 package org.wildfly.swarm.plugin.gradle;
 
-import javax.inject.Inject;
-
+import org.gradle.api.Action;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.dsl.DependencyHandler;
 import org.gradle.tooling.provider.model.ToolingModelBuilderRegistry;
+
+import javax.inject.Inject;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Map;
 
 /**
  * Gradle plugin for enabling Arquillian tests based on Thorntail. This is useful in cases where a Gradle project doesn't
@@ -29,6 +41,8 @@ import org.gradle.tooling.provider.model.ToolingModelBuilderRegistry;
 public class ThorntailArquillianPlugin extends AbstractThorntailPlugin {
 
     public static final String PLUGIN_ID = "thorntail-arquillian";
+
+    private static final String BUILD_SCRIPT_FRAGMENT = "META-INF/thorntail-scripts/configure-tasks.groovy";
 
     /**
      * Constructs a new instance of {@code PackagePlugin}, which is initialized with the Gradle tooling model builder registry.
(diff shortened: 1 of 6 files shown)


roguexz:
@Ladicek @TomasHofman 
Another one for you to review.
I have started looking at performance improvements for the Gradle plugin. As I find out more, I will keep contributing them back.

FYI, I have the changes done for the examples project as well. Just want to make one more change that has a huge impact on the performance of testing the plugins. If I unable to make much progress on that front, I will raise the PR for examples project anyways.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailArquillianPlugin.java
```diff
@@ -44,11 +58,65 @@ public ThorntailArquillianPlugin(ToolingModelBuilderRegistry registry) {
...
+                }
+
+                URL resource = ThorntailArquillianPlugin.class.getClassLoader().getResource(BUILD_SCRIPT_FRAGMENT);
+                if (resource != null) {
+                    try {
+                        File scriptFile = new File(project.getBuildDir(), "thorntail-arquillian-script.gradle");
+                        if (!scriptFile.exists()) {
+                            scriptFile.createNewFile();
+                            String[] lines = readString(resource.openStream()).split("\n");
+                            Files.write(scriptFile.toPath(), Arrays.asList(lines));
```
Ladicek, line 96: Why split the string into lines if you immediately merge them back?

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailArquillianPlugin.java
```diff
@@ -44,11 +58,65 @@ public ThorntailArquillianPlugin(ToolingModelBuilderRegistry registry) {
...
+        });
+    }
+
+    /**
+     * Read the given stream in to a String. This method will close the stream as well.
+     *
+     * @param stream the input stream.
+     * @return the string built out of the data available in the given stream.
+     */
+    private static String readString(InputStream stream) throws IOException {
```
Ladicek, line 116: Isn't it sad that there's no equivalent of `Files.readAllBytes` for classpath resources? :-/

plugins/gradle/gradle-plugin/src/main/resources/META-INF/thorntail-scripts/configure-tasks.groovy
```diff
@@ -0,0 +1,29 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+// Given our current integration model, we end up in a scenario where we are unable to perform all integrations from
+// Java layer as that would mean that we now need to have a dependency on a newer version of the Gradle plugin.
```
Ladicek, line 20: Hmm... I briefly worked with Gradle around their 1.0 time, and I think that even back then, their Java API already exposed enough to find all tasks of type `Test` in the project. In fact, I think the Groovy scripts use the same API that you can use from Java. So I'm not quite sure that we need to add an extra Groovy script; but of course you have much more experience with Gradle than me :-)

plugins/gradle/tooling-api/src/main/java/org/wildfly/swarm/plugin/gradle/GradleToolingHelper.java
```diff
@@ -27,6 +27,17 @@
...
 public final class GradleToolingHelper {
 
+    /**
+     * The System environment that will passed from the plugin to testing tasks which will contain the Gradle version
+     * being used for the current test execution. This environment variable is setup in order to facilitate the testing
+     * of the project against a newer version of Gradle instead of having to upgrade the Gradle wrapper.properties file.
+     * <p>
+     * Also, useful when the project root folder does not contain the Gradle wrapper properties, as is the case with the
+     * Thorntail Gradle examples. Since we cannot anticipate all possible scenarios, it is advisable to use the same
+     * version of Gradle that started this build.
```
Ladicek, line 37: +1

Ladicek (dismissed):
Looks good. Just added a few minor comments, but nothing blocking.

Ladicek:
I'll wait a bit to give you a chance to reply to my comments, but I'd be fine with merging as is.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailArquillianPlugin.java
roguexz: So, when I passed in a string directly (which had `\n`), it messed up the content in the file. I tried seeing if there was a convenience method in the codebase, but was unable to locate one. If you know of something, then I can switch to that.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailArquillianPlugin.java
roguexz: I know! Additionally, the `Files` API, works only with files and not resources inside a Jar or other locations.

plugins/gradle/gradle-plugin/src/main/resources/META-INF/thorntail-scripts/configure-tasks.groovy
roguexz: So the problem that I faced was the following,

```
Test (task) --(references)-- DefaultJavaForkOptions
DefaultJavaForkOptions --(extends)--> DefaultProcessForkOptions --(implements)--> ProcessForkOptions
```

Now, `ProcessForkOptions` isn't available on the classpath, which causes the compilation to fail (as it is an internal API). Since Gradle doesn't ship an API compatible jar that we can use, we end up having to resort to a runtime resolution instead of relying at Reflection at compile time. Between the two choices, the runtime option keeps things simple. I was planning on removing all the reflection code in favor of clear concise Groovy code. Hence the decision to go with this approach. But I am definitely open to suggestions.

plugins/gradle/gradle-plugin/src/main/java/org/wildfly/swarm/plugin/gradle/ThorntailArquillianPlugin.java
```diff
@@ -44,11 +58,68 @@ public ThorntailArquillianPlugin(ToolingModelBuilderRegistry registry) {
...
+                    handler.add(config.getName(), handler.create(gav));
+                } else {
+                    System.err.println("Unable to add Gradle Tooling APIs to the " + CONFIGURATION +
+                            ". Thorntail Arquillian integration might not work.");
+                }
+
+                URL resource = ThorntailArquillianPlugin.class.getClassLoader().getResource(BUILD_SCRIPT_FRAGMENT);
+                if (resource != null) {
+                    try {
+                        if (!project.getBuildDir().exists()) {
```
roguexz, line 89: Resolved a small bug that would have become a show stopper :) .. All thanks to the Thorntail Examples. I should probably add a note to the README reminding myself and others that we should verify the plugin changes against the example projects before raising a PR.

roguexz:
@Ladicek I am done with the changes for this PR. I have verified the changes in the example projects. Additional PRs will only reduce the amount of code that we need to write in the build scripts.

plugins/gradle/gradle-plugin/src/main/resources/META-INF/thorntail-scripts/configure-tasks.groovy
Ladicek: Understood, thanks!

Ladicek (approved):

PR #1342, opened 2019-10-08
Avoid replacing the original war file if no filtering is required

roguexz:
The current implementation will always replace the built war file irrespective of whether the `filterWebInfLib` flag has been set or not. By doing an early check we can speed up the execution time for projects that build large war files.

Additionally, this would also help the Gradle plugin leverage task avoidance which is currently not happening. By replacing the war file (with the same content), Gradle believes that the file has changed and thus re-executes the war task, which will re-execute the thorntail-package task. With this change, the Gradle plugin will be able to avoid repackaging and speed up development.

-----

-----

### DIFF
--- a/tools/src/main/java/org/wildfly/swarm/tools/BuildTool.java
+++ b/tools/src/main/java/org/wildfly/swarm/tools/BuildTool.java
@@ -218,7 +218,7 @@ public class BuildTool {
             original.as(ZipImporter.class).importFrom(inputStream);
         }
 
-        Archive repackaged = this.filterWebInfLib ? new WebInfLibFilteringArchive(original, this.dependencyManager) : original;
+        Archive repackaged = new WebInfLibFilteringArchive(original, this.dependencyManager);
         repackaged.as(ZipExporter.class).exportTo(file, true);
         this.log.info("Repackaged .war: " + file);
     }
--- a/tools/src/main/java/org/wildfly/swarm/tools/BuildTool.java
+++ b/tools/src/main/java/org/wildfly/swarm/tools/BuildTool.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2015-2017 Red Hat, Inc, and individual contributors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -204,6 +204,10 @@ public class BuildTool {
 
 
     public void repackageWar(File file) throws IOException {
+        if (!filterWebInfLib) {
+            return;
+        }
+
         this.log.info("Repackaging .war: " + file);
 
         Path backupPath = get(file);


tools/src/main/java/org/wildfly/swarm/tools/BuildTool.java
```diff
@@ -204,6 +204,10 @@ public File build(String baseName, Path dir) throws Exception {
 
 
     public void repackageWar(File file) throws IOException {
+        if (!filterWebInfLib) {
+            return;
+        }
+
         this.log.info("Repackaging .war: " + file);
```
roguexz, line 211: In this method, the file is first moved, then overwritten by the same content if `filterWebInfLib` is marked as false. Hence the suggestion for improvement. Please do share your thoughts about it.

Ladicek: I think originally, there was no `filterWebInfLib` setting, we always filtered. `filterWebInfLib` defaults to `true`, if I remember correctly.

This change makes sense, it's better than `Archive repackaged = this.filterWebInfLib ? new WebInfLibFilteringArchive(original, this.dependencyManager) : original;`. Perhaps we should revert that back to `Archive repackaged = new WebInfLibFilteringArchive(original, this.dependencyManager)`

roguexz: You are correct. The default value for `filterWebInfLib` is set to true. I looked at the [WebInfLibFilteringArchive](https://github.com/thorntail/thorntail/blob/master/tools/src/main/java/org/wildfly/swarm/tools/WebInfLibFilteringArchive.java#L48) and the only thing that it seems to do is remove the `thorntail-runner` jar from the library assets.

I can make the change to the conditional as well. Let me know if you have other review comments, if not I will update this PR in the evening which will incorporate the above comment.

Ladicek: I don't have any other comment.

There's a Windows test failure, but it seems unrelated, so feel free to go ahead.

roguexz: Perfect timing!

roguexz:
Windows build is broken due to space constraints. I wonder if there are any broken builds that failed to clean up the exploded archives from the tmp directory.

```Caused by: java.io.IOException: There is not enough space on the disk```

Ladicek (approved):

Ladicek:
We've been always struggling with cleanup on Windows :-/ I've cleaned up the Windows agents manually.

retest this please

Ladicek:
Thanks!

PR #1347, opened 2019-10-30
THORN-2492: Maven plugin ignores exclusions when creating wildfly-swarm-manifest.yaml

Ladicek:
Motivation
----------
When the `pom.xml` declares an exclusion for a library, this exclusion
is not taken into account when the `wildfly-swarm-manifest.yaml` file
is created. This can cause library conflicts at runtime.

Example:
The SLF4J logging framework supports different underlying logging systems,
e.g. Log4j or Logback; however, only one of the corresponding binding JAR
files is allowed to be present at runtime.

When a Thorntail project uses the SLF4J logging framework, declares
a direct dependency on one of the bindings, and declares a dependency
on a 3rd party library (e.g. ZooKeeper) which itself declares a dependency
on another binding, this will lead to multiple bindings being present
at runtime. The 3rd party library should normally not declare that dependency,
but if it does, the only solution is to add an exclusion; however, this is
ignored when the `wildfly-swarm-manifest.yaml` file is created, so the SLF4J
system complains at runtime about duplicate bindings.

Modifications
-------------
When the uberjar is being built by the Maven plugin, we can access
the entire dependency tree exactly how Maven sees it. Therefore, we don't
have to figure out the transitive dependencies ourselves, which has a nice
benefit: dependency exclusions are taken into account.

Result
------
Only affects how the Maven plugin builds the uberjar. Other code paths
that build an uberjar are not affected, which unfortunately means
there might be behavioral differences between e.g. Maven plugin
and Arquillian.

-----

### DIFF
--- a/plugins/gradle/tooling-api/src/main/java/org/wildfly/swarm/plugin/gradle/GradleToolingHelper.java
+++ b/plugins/gradle/tooling-api/src/main/java/org/wildfly/swarm/plugin/gradle/GradleToolingHelper.java
@@ -48,10 +48,11 @@ public final class GradleToolingHelper {
         DeclaredDependencies dependencies = new DeclaredDependencies();
         depMap.forEach((directDep, resolvedDeps) -> {
             ArtifactSpec parent = toArtifactSpec(directDep);
-            dependencies.addResolved(parent);
+            dependencies.add(parent);
             if (resolvedDeps != null) {
-                resolvedDeps.forEach(d -> dependencies.addResolved(parent, toArtifactSpec(d)));
+                resolvedDeps.forEach(d -> dependencies.add(parent, toArtifactSpec(d)));
             }
+            dependencies.markComplete(parent);
         });
         return dependencies;
     }
--- a/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
+++ b/plugins/maven/src/main/java/org/wildfly/swarm/plugin/maven/PackageMojo.java
@@ -192,6 +192,7 @@ public class PackageMojo extends AbstractSwarmMojo {
                     declaredDependencies.add(directDep, transientDep);
                 }
             }
+            declaredDependencies.markComplete(directDep);
         }
 
         tool.declaredDependencies(declaredDependencies);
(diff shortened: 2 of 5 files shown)


Ladicek:
Marking as _hold_, because this apparently breaks fraction autodetection.

Ladicek:
This variant seems to work fine with autodetection. I'll run some more tests on Monday or so.

Ladicek:
Hi @roguexz, I'm touching a sensitive area here that I know you modified for the Gradle plugin purposes. The first commit seems OK from the Maven perspective; the second commit is an experiment that I'm not 100% sure about (but the Maven plugin tests also pass). If you could find some time to take a look at this, that would be great. Thanks!

Ladicek:
Hi @roguexz, I appreciate you're busy, just wanted to let you know that it would be really nice to have the Gradle examples working, to serve as a smoke test for the Gradle plugin. Because I'm slowly leaning towards just merging this :-)

roguexz:
Hi Ladislav. Sorry for the hiatus. I was traveling the past few weeks. I just got back a few days ago. Let me pull this up locally and test it.

Last time I was trying to understand how the BuildTool packaged the dependencies. I got to the point of mimicking dependencies as the maven plug-in would provide. Let me rebase and test it tomorrow.

Ladicek:
Thanks @roguexz! I will need to decide about the fate of this PR in the next few days, so you're coming just in time :-)

roguexz (commented):
Changes look fine. I don't think we will have any issues with the Gradle plugin as such. The problem that I am trying to solve is around building `war` files with `package-all-dependencies`. In this scenario, libraries that are part of the `WEB-INF/lib` should be excluded from the build tool bundling, which is challenging. With the exception of test cases building `war` files, the other test cases can be verified. I can make the required changes to the examples project to ensure that all `uber jar` based test cases are executed. Please do let me know if you would want me to make that change.

Ladicek:
@roguexz I'll leave the Gradle examples decision up to you, I'll just note that having them automatically tested in some way would be really good, as we don't have any other test for the Gradle plugin.

Ladicek:
retest this please

PR #1350, opened 2019-11-05
[THORN-2504] Clean up MP JWT CDI code

sberyozkin:
Motivation
----------
`microprofile-jwt` fraction has some unused CDI code and it is difficult to see what parts of this code are still needed.

Modifications
-------------
The unused code has been removed and  Provider-specific injection support code moved to its own class to be replaced shortly by the `smallrye-jwt` equivalent code. 
More tests have been added as well

Result
------
Simpler CDI extension code in Thorntail MP-JWT, and other containers such as WildFly may also benefit from a better CDI support in `smallrye-jwt`


testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/ApplicationScopedPrincipalLeakTest.java
```diff
@@ -52,26 +52,48 @@ protected static WARArchive initDeployment() {
     public void subjectShouldNotLeakToNonSecuredRequest() throws Exception {
         // project-no-roles-props.yml restricts the number of worker threads to 1,
         // that is, all requests are processed by the same single thread
-
-        String response = Request.Get("http://localhost:8080/mpjwt/subject/secured")
-                .setHeader("Authorization", "Bearer " + createToken("MappedRole"))
-                .execute().returnContent().asString();
-        assertThat(response).isEqualTo(TokenUtils.SUBJECT);
-
```
Ladicek, line 60: Why do you remove this? The test will no longer verify what it used to verify, no?

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/PrincipalLeakTest.java
```diff
@@ -57,14 +57,26 @@ protected static WARArchive initDeployment() {
     public void subjectShouldNotLeakToNonSecuredRequest() throws Exception {
         // project-no-roles-props.yml restricts the number of worker threads to 1,
         // that is, all requests are processed by the same single thread
-
```
Ladicek, line 60: Same here, the `subjectShouldNotLeakToNonSecuredRequest` test will no longer verify what it used to verify.

Ladicek:
The main code LGTM, but I'd like to understand the changes to the tests.

testsuite/testsuite-microprofile-jwt/src/test/java/org/wildfly/swarm/microprofile/jwtauth/ApplicationScopedPrincipalLeakTest.java
sberyozkin: This is fine, now this test just tests that a null subject is returned for non secure requests, this removed code is now tested  with `subjectFromJsonWebToken()`, let me add 'secureRequest' prefix to some tests.

Ladicek: Hmm, those tests don't look like they would check that subject won't leak into a subsequent request...?

sberyozkin: @Ladicek OK, I've updated all the tests now to check no subsequent request leak occurs

Ladicek (approved):

PR #1359, opened 2019-12-11
Gradle Plugin: Optimize lookup for module dependencies in DependencyManager

roguexz:
The ModuleAnalyzer class looks for dependencies in a predefined location and assumes that the folder conforms
to the Maven repository layout. If it doesn't find the dependency in that location, then the ModuleAnalyzer will
return an ArtifactSpec object that does not contain the file location. But when using the Gradle plugin, it
is highly possible that the dependency was already retrieved and stored in the Gradle dependency cache.

In this commit, I am performing an additional check to handle such scenarios.

-----

### DIFF
--- a/tools/src/main/java/org/wildfly/swarm/tools/DependencyManager.java
+++ b/tools/src/main/java/org/wildfly/swarm/tools/DependencyManager.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2015-2017 Red Hat, Inc, and individual contributors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
@@ -210,6 +211,9 @@ public class DependencyManager implements ResolvedDependencies {
         this.dependencies.addAll(
                 resolver.resolveAllArtifactsNonTransitively(remainder)
         );
+
+        // populate the dependency map for faster lookups
+        this.dependencies.forEach(s -> dependencyMap.put(s.mavenGav(), s));
     }
 
     private void filterOutRunnerDependencies(ArtifactSpec runnerJar, Collection<ArtifactSpec> explicitDependencies) {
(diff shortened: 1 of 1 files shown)


tools/src/main/java/org/wildfly/swarm/tools/DependencyManager.java
```diff
@@ -233,7 +237,17 @@ private void analyzeModuleDependencies() {
 
     private void analyzeModuleDependencies(ModuleAnalyzer analyzer) {
         List<ArtifactSpec> thorntailDependencies = analyzer.getDependencies();
-        this.moduleDependencies.addAll(thorntailDependencies);
+
+        // ModuleAnalyzer looks for dependencies in a predefined location and assumes that the folder conforms to the
+        // Maven repository layout. If it doesn't find the dependency in that location, then the ModuleAnalyzer will
+        // return an ArtifactSpec object that does not contain the file location. But when using the Gradle plugin, it
+        // is highly possible that the dependency was already retrieved and stored in the Gradle dependency cache.
+
```
roguexz, line 245: Please review this change after going through the implementation in [ModuleAnalyzer](https://github.com/thorntail/thorntail/blob/master/tools/src/main/java/org/wildfly/swarm/tools/ModuleAnalyzer.java#L86).

roguexz:
@Ladicek 
After debugging the code, I found the reason for the underlying failures of the Gradle examples.

Generally speaking, I would like to propose a few changes to the ArtifactResolvingHelper and streamline the usage across both build-tooling as well as the Arquillian integration. Since that would be a change to the core libraries, I shall follow the JIRA route for it.

Ladicek:
Hi @roguexz, thanks for the PR! I'm on vacation starting today, and even though this PR is small, it touches a sensitive area of the code, so I don't feel like merging it without going through the code in an IDE. I hope you don't mind, but I'll leave this after new year. Thanks for understanding.

roguexz:
I completely understand. We can pick this up once you're back. Happy holidays and hope you have wonderful year ahead. Cheers!

roguexz:
Hi @Ladicek. Happy New Year. Do let me know if you have questions about this PR. I have rebased it with the latest set of changes.

-Cheers

Ladicek:
Hi @roguexz, I know I need to look into this. I have some high-priority items on my plate now, but I should have some time for this next week. Thanks for understanding, and happy new year to you too! :-)

Ladicek (approved):

Ladicek:
Sorry it took so long!

roguexz:
Not a problem at all Ladicek. Given that supporting Gradle within Thorntail isn't exactly our primary jobs, I completely understand 😃

PR #1363, opened 2020-01-28
THORN-2529: can't express "null" value in environment variables

Ladicek:
Motivation
----------
When overriding configuration using environment variables (and system
properties), it is not possible to express the `null` value. This is
sometimes required, e.g. to set the various `*-relative-to`
management attributes (where `null` means that the corresponding path
attribute is absolute path).

We should invent a placeholder value, such as `<<null>>`, which
would be translated to `null` when present.

Modifications
-------------
The code that reads environment variables and properties (system
properties included) now recognizes the `<<null>>` placeholder
string and translates that to actual `null` value.

Result
------
Able to express `null` in environment variables and system properties.

-----


Ladicek:
The "Linux - JDK 11" build failed because "Too many open files". Hmm...

Ladicek:
retest this please

PR #1364, opened 2020-01-29
THORN-2530: udpate ShrinkWrap Resolver

Ladicek:
Motivation
----------
Maven Central no longer supports plain HTTP access,
which the old version of ShrinkWrap Resolver uses.
We need to use access Maven Central over HTTPS,
for which we need a new version of ShrinkWrap Resolver.

Modifications
-------------
Updated ShrinkWrap Resolver.

Result
------
No behavioral difference.

-----

### DIFF
--- a/boms/src/main/resources/bom-template.xml
+++ b/boms/src/main/resources/bom-template.xml
@@ -49,6 +49,7 @@
   <properties>
     <version.shrinkwrap>${version.org.jboss.shrinkwrap}</version.shrinkwrap>
     <version.shrinkwrap-descriptors>${version.org.jboss.shrinkwrap.descriptors}</version.shrinkwrap-descriptors>
+    <version.shrinkwrap-resolver>${version.org.jboss.shrinkwrap.resolver}</version.shrinkwrap-resolver>
     <version.thorntail>${project.version}</version.thorntail>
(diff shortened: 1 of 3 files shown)


Ladicek:
This is not enough. Need to look more into it.

Ladicek:
Updated, this should be better.

Ladicek:
Hmm, a lot of tests is failing with `ClassNotFoundException`...

Ladicek:
I've updated Maven on the Jenkins agents. Perhaps latest ShrinkWrap Resolver, which uses Maven 3.6.3, isn't compatible with Maven 3.3.9 on our Jenkins, not sure.

retest this please

Ladicek:
Turns out either Jenkins didn't pick up the Maven 3.6.3 installation, or it isn't enough. I'll probably proceed with an attempt to get the HTTPS fix backported to ShrinkWrap 2.2.x and get a release.

mspasiano:
You can use a workaround like this:
https://github.com/consiglionazionaledellericerche/sigla-main/blob/f2bbad20b0d92f97b19d1d04438f5cb2a7d85341/sigla-web/src/test/java/it/cnr/test/util/Deployments.java#L72

Ladicek:
That would be an option, but would require figuring out all the code paths that use ShrinkWrap Resolver. I think getting ShrinkWrap Resolver 2.2.x fixed and released is the proper fix in the short term.

PR #1365, opened 2020-01-31
THORN-2531: Null value for thorntail.public.url.base

Ladicek:
Motivation
----------
In the previous version of Swarm (e.g. 2018.1.0), we could use
the key `swarm.public.url.base` in the file `usage.txt` in order
to display the host of our application when it has been deployed.

Since THORN-2185, we can use `thorntail.public.url.base`, but
it's not working, it always returns `null`.

The bug has been introduced in `NetworkVariableSupplier#valueOf`:

```java
if (name.equals("thorntail.public.url.base")) {
    return "http://" + valueOf("thorntail.public.host") + ":" + valueOf("thorntail.http.port") + "/";
}

String[] parts = name.split("\\.");
if (parts.length > 0 && parts[0].equals("swarm")) {
    ...
}
```

Modifications
-------------
Updated some code paths that still expected system properties
prefixed with `swarm.*` to correctly read `thorntail.*`.

Result
------
Using `thorntail.public.url.base` should work now.

-----

### DIFF
--- a/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/AnnotationBasedMain.java
+++ b/arquillian/adapter/src/main/java/org/wildfly/swarm/arquillian/adapter/AnnotationBasedMain.java
@@ -27,7 +27,7 @@ import org.wildfly.swarm.arquillian.CreateSwarm;
  * @uathor Ken Finnigan
  */
 public class AnnotationBasedMain {
-    public static final String ANNOTATED_CLASS_NAME = "swarm.arquillian.createswarm.class";
+    public static final String ANNOTATED_CLASS_NAME = "thorntail.arquillian.createswarm.class";
 
     protected AnnotationBasedMain() {
     }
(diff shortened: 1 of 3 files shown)


HubbleDK (commented):
Sounds good but can't we add a test that makes sure it fixes the bug and won't break again ?

Ladicek:
Certainly, but I'm pretty sure Thorntail will not be renamed again :-)

HubbleDK:
I'll take time to test it ASAP

Ladicek:
I killed all the Jenkins jobs to do a Maven update in Jenkins.

retest this please

HubbleDK (approved):
It's working as expected.

proof

[usage.txt](https://github.com/thorntail/thorntail/files/4140937/usage.txt)

Ladicek:
Thanks @HubbleDK for confirmation!

I've reverted Maven in Jenkins to 3.3.9, as the upgrade didn't achieve what I hoped for, and we should probably still support 3.3.9 minimum.

Jenkins, retest this please

PR #1368, opened 2020-03-04
THORN-2535 Default health check response readiness should not be 200

xstefank:

-----

https://issues.redhat.com/browse/THORN-2535

### DIFF
--- a/testsuite/testsuite-microprofile-health/src/test/java/org/wildfly/swarm/microprofile/health/StartupHealthTests.java
+++ b/testsuite/testsuite-microprofile-health/src/test/java/org/wildfly/swarm/microprofile/health/StartupHealthTests.java
@@ -1,0 +1,153 @@
+package org.wildfly.swarm.microprofile.health;
+
+import org.jboss.arquillian.container.test.api.Deployer;
+import org.jboss.arquillian.container.test.api.Deployment;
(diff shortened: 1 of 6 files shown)


Ladicek:
add to whitelist

testsuite/testsuite-microprofile-health/src/test/java/org/wildfly/swarm/microprofile/health/StartupHealthTests.java
```diff
@@ -0,0 +1,153 @@
...
+        deployer.undeploy(APP_NAME);
+    }
+
+    private <T> Callable<T> waitForNonNull(Callable<T> waitFunction) {
+        return () -> {
+            T response = null;
+
+            while ((response = waitFunction.call()) == null) {
+                try {
+                    Thread.sleep(1000);
```
Ladicek: 1 second is a bit too much. It's unlikely, but it can happen that the time between HTTP port becomes available and application deployment finishes is less than 1 second. I'd use 100 millis here.

(Usually, I'd suggest using Awaitility, but for some reason, we don't have that in Thorntail, and it probably doesn't make sense to add it just for this purpose.)

Ladicek (changes requested):
Just one detail, otherwise LGTM.

Ladicek (approved):

PR #1371, opened 2020-04-09
Use of thorntail.runner.webapp-location generates a invalid WAR

fronald:
Jira issue: THORN-2538

Our company is using thorntail for all new microservice projects and we found that using thorntail.runner.webapp-location property an invalid WAR is generated.

For productivity reasons, we run the thorntail runner pointing to the target directory. This is very useful when the application has customizations performed at post-packaging time. For example:

* Maven overlays;
* Custom Anttasks or scripts.

With this fix, we were able to run the thorntail-runner by taking advantage of the hot-swap of classes and JSPs at run time in the IDE with all the complete post-packaging steps that a Maven project in any IDE already supports.

We hope you enjoy the patch with the fix.

Long live Thorntail!

### DIFF
--- a/thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
+++ b/thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
@@ -148,7 +148,9 @@ public class WarBuilder {
                 String projectDir = System.getProperty("thorntail.runner.webapp-location");
                 if (projectDir != null) {
                     projectDir = Paths.get(projectDir).toFile().getAbsolutePath();
-                    if (file.getAbsolutePath().contains("WEB-INF" + File.separator + "classes")) { // Ignore classes.
+                    if (file.getAbsolutePath().contains("WEB-INF" + File.separator + "classes") || // Ignore classes.
+                            file.getAbsolutePath().contains("WEB-INF" + File.separator + "lib") // Ignore JARs.
+                    ) {
                         return;
                     }
                 } else {
--- a/thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
+++ b/thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
@@ -145,8 +145,15 @@ public class WarBuilder {
         File file = path.toFile();
         if (file.isFile()) {
             try {
-                String projectDir = Paths.get("src", "main", "webapp").toFile().getAbsolutePath();
-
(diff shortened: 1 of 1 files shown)


Ladicek:
ok to test

thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
```diff
@@ -145,8 +145,15 @@ private void addWebappResourceToWar(Path path) {
...
-
+                String projectDir = System.getProperty("thorntail.runner.webapp-location");
+                if (projectDir != null) {
+                    projectDir = Paths.get(projectDir).toFile().getAbsolutePath();
+                    if (file.getAbsolutePath().contains("WEB-INF" + File.separator + "classes")) { // Ignore classes.
+                        return;
+                    }
+                } else {
+                    projectDir = Paths.get("src", "main", "webapp").toFile().getAbsolutePath();
+                }
```
Ladicek, line 158: I took a look at the code and found this is really 2 different things:

1. The `addWebappResourceToWar` method doesn't respect the `thorntail.runner.webapp-location` setting. The `addWebAppResourcesToWar` method should really pass the `webappPath` to `addWebappResourceToWar` when invoking it. The `addWebappResourceToWar` method shouldn't reread the system property all the time. This should be a simple change.

2. Filtering out `WEB-INF/classes`. I think the `thorntail.runner.webapp-location` was meant to point to a _source_ directory. Pointing it to a _target_ directory requires such filtering, but I'm not sure if just ignoring `WEB-INF/classes` is enough. How about `WEB-INF/lib`? That should be enough, probably.

I think we probably can filter out these 2 paths unconditionally, even if the system property points to a source directory -- the source directory shouldn't contain these directories at all.

Ladicek:
I have submitted an alternative implementation of the same idea here: https://github.com/thorntail/thorntail/pull/1372 If you could check it out, that would be great :-) Thanks!

thorntail-runner/src/main/java/org/wildfly/swarm/runner/WarBuilder.java
fronald: I can make this change by ignoring the lib directory as well. Soon I will send.

Thank you very much for feedback.

fronald:
> I have submitted an alternative implementation of the same idea here: #1372 If you could check it out, that would be great :-) Thanks!

Ow! Perfect!

Ladicek:
Merged #1372 instead.

PR #1376, opened 2020-05-29
update to MicroProfile 3.3, Keycloak 9.0.3 and Config API 2.7.0

Ladicek:
[THORN-2553] update to MicroProfile 3.3
[THORN-2560] update to Keycloak 9.0.3
[THORN-2562] update to WildFly Config API 2.7.0

Updates to MicroProfile specifications and SmallRye implementations
are relatively straightforward (just dependency updates), with
a couple of exceptions.

MP Config requires new MicroProfile Config WildFly release, where
we update to SmallRye Config 1.6.2 and bring in WildFly subsystem
changes. We also need to exclude some TCK tests (see also [1]),
because SmallRye Config changed behavior when it comes to handling
empty values. This is being debated in MP Config.

MP Fault Tolerance requires updating to SmallRye Fault Tolerance 4,
which no longer uses Hystrix. This requires reimplementing OpenTracing
integration. SmallRye Fault Tolerance 4 has OpenTracing integration,
but it's based on MicroProfile Context Propagation, which we don't
support (and don't plan to support) in Thorntail. Therefore, we now
have a minimal context propagation API directly in Thorntail, for                                                                                                
this single purpose.                                                                                                                                             

MP RestClient previously required implementing an extra SPI                                                                                                      
(`IncomingHeadersProvider`), but that's no longer necessary.                                                                                                     
What is necessary is updating to RESTEasy 3.11.2, which also
requires updating to Jackson 2.10.3.

MP Metrics integration depends on a MSC service that signalizes
that MP Config is ready. The MP Config subsystem in WildFly
doesn't install such service, but we retain that in our own
MicroProfile Config WildFly project.

When updating SmallRye JWT, we need to explicitly add a dependency
on Jakarta Annotations. SmallRye JWT used to have that dependency
itself, but it no longer does. This dependency is also transitively
propagated to consumers, as it was before. Also we need to update
Jose4j, as SmallRye JWT now requires a new version.

With the update to Keycloak 9.0.3, we need to remove the Drools
module from the `keycloak-server` fraction, because Drools are
no longer supported with Keycloak (see also [2]).

The update to WildFly Config API 2.7.0 is only a reversion,
no other changes.

[1] https://github.com/smallrye/smallrye-config/blob/1.6.2/testsuite/tck/tck-suite.xml
[2] https://www.keycloak.org/docs/latest/release_notes/index.html#keycloak-9-0-0

-----


Ladicek:
retest this please

